PR #392, opened 2011-09-05
Output cache now uses CodeIgniter caching library.

weejames:
Updated the output library caching mechanism.  Now uses the Cache
library and associated drivers to do full page caching.  Given that
$_GET is no longer a leper in CI have added that to the cache key
generation.  Any headers set are also cached and resent on subsequent
requests for the cached page.

Have tested with APC, Memcached and file system caches.

### DIFF
--- a/application/config/output.php
+++ b/application/config/output.php
@@ -1,0 +1,3 @@
+<?php
+
+$config['adapter'] = 'file';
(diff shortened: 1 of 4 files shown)


Alphapixels:
+1

PR #868, opened 2012-01-05
Improve the Oracle (oci8) database driver

narfbg:
Changes include:
- `oci_execute()` related:
  - Moved the call from `num_rows()` to `_data_seek()`, as it's purpose is to reset the internal pointer to zero. It's now only called by `CI_DB_oci8_driver::_execute()` (as it should). `_data_seek()` is left there for consistency and isn't called anywhere. This was the main reason for every performance problem with this driver.
  - If we're using PHP version >= 5.3.2 - we'll now utilize the new `OCI_NO_AUTO_COMMIT` mode flag instead of `OCI_DEFAULT`, which is left as an alias.
  - Now only pass the mode flag if we are in the middle of a transaction (see the notes about `OCI_NO_AUTO_COMMIT` in [php.net](http://www.php.net/manual/en/function.oci-execute.php) for details). `OCI_COMMIT_ON_SUCCESS` is the default, so it doesn't need to be explicitly set.
- `num_rows()` related:
  - Now must be called explicitly (used to be automatically called by `CI_DB_driver`) to avoid `result_array` be set every time while it might not be needed (another possible performance glitch).
  - The `num_rows` property is now set to `NULL` by default, so we can have a way to know if 0 is the actual count or we have to get it. Also put multiple hooks for this to be set by other methods. Of course - this still doesn't allow `num_rows()` to be used to check before actual data is fetched (as in other database drivers), because we still need to actually do a `count()` on the result set. I hope that Oracle will some day release a solution for this.
  - Added an oracle-specific note to the `num_rows()` documentation.
- `free_result()` now really frees `stmt_id` and/or `curs_id`, if they are valid resources. Until now it did a check if `result_id` is a resource and try to clear it, but that's actually a boolean value.
- Added `custom_result_object()` and all of the row-fetching methods instead of using the parents from `CI_DB_result`. The new ones only fetch rows when needed, follow `current_row` on their own (we don't want to use `_data_seek()`) and cast from/to arrays/objects when needed. This should also significantly improve performance for those who like to deal with single rows at a time.
- Added `DROP TABLE` support in `CI_DB_oci8_forge`.
- Added `_list_databases()` support in `CI_DB_oci8_utility`.
- Fixed some typos and cleaned up some comments, spaces, unneeded expressions, etc.

I know it's a rather large patch to review/merge, but in order to avoid calling `oci_execute()` multiple times - pretty much all of it should be done at once, so there's no way around it. Plus it should solve every oci8-related issue so far reported.

I've tested most of the stuff (actually, all but DB forge and DB utility) - it works, so ... I hope it doesn't take too much time to review. Any suggestions are welcome. :)


amidia:
for session with oracle database usage, the table and columns names are uppercase, how to make session to comply to this oracle rule? 
it is work when we make strupper to keys in /system/libraries/session.php, but it's feel not right.
any solution sir?

PR #1092, opened 2012-02-29
Add an optional set_mime parameter to force_download()

narfbg:
The function name itself implies that a download would be _forced_, but most browsers would not initiate a download if there's a browser plugin (such as a PDF reader or many video players) or if the file is an image. They would instead handle the content internally and not even ask the user wether to download the file or not. Setting an `application/octet-stream` content type will force the browser to show the Open with/Save dialog.
This is why even though this is new functionality, it should be the default behavior and it's kept as an option to allow sending the correct MIME type, in case that somebody needs that.

Also did some minor clean-up and fixed a possible bug with IE introduced in #957, where the Pragma header would be sent twice.

Forums reference: http://codeigniter.com/forums/viewthread/212055/


Johnwilliams1:
public function sponsorship_opportunities(){
		$data['insert_opportunities']=$this->frontend_model->insert_opportunities();
		$data['files']=$this->frontend_model->brochure_file(2);
		foreach($data['files']->result() as $file){}
		if($data['insert_opportunities']==1){
				$file_name = $file->uploaded_document;
				$file_url = base_url().'uploads/documents/' . $file_name;
				header('Content-Type: application/octet-stream');
				header("Content-Transfer-Encoding: Binary"); 
				header("Content-disposition: attachment; filename=\"".$file_name."\""); 
				header("Content-Length: " . filesize($file_url));

				
				readfile($file_url);
			 $this->session->set_flashdata('message', "<div class='alert alert-success'>Sumbitted.</div>");
					redirect(base_url('home/exhibit_sponsor'));
			 }
			 else{
				 $this->session->set_flashdata('message', "<div class='alert alert-danger'>Something wrong.</div>");
					redirect(base_url('home/exhibit_sponsor'));
				 }
		}






please help to get ou this problem that pdf is not opening

narfbg:
@Johnwilliams1 This is a six year old pull request, and you're just randomly asking for help on a unrelated problem that you have ... that's not cool.

Ask your questions on our [forums](https://forum.codeigniter.com/) instead and you should be able to get some help.

PR #1332, opened 2012-05-06
url_title() accents support

PhiloNL:
At the moment url_title() breaks when using accents.
For example:

``` php
<?php
$string = 'Café';
echo url_title($string, '-', TRUE); // Returns caf
echo convert_accented_characters($string); // Returns Cafe
```

This change will convert the accents so a proper url title is returned.

``` php
echo url_title($string, '-', TRUE); // Returns cafe
```

### DIFF
--- a/system/helpers/url_helper.php
+++ b/system/helpers/url_helper.php
@@ -463,12 +463,21 @@ if ( ! function_exists('prep_url'))
  * @param	string	the string
  * @param	string	the separator
  * @param	bool
+ * @param	bool
  * @return	string
  */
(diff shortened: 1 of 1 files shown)


imabot2:
I totally agree with the fact that url_title() should convert accented characters. I don't see any case where such characters shouldn't be converted.

narfbg:
> I totally agree with the fact

The worlds big problem today - people think facts are things they agree with. Literally everything about this issue is opinions, preferences, agreements and disagreements; not facts.

For the last time - just use the two functions together if you need to.

PR #1448, opened 2012-06-08
html5 form inputs

cinan:
form_input() can be used for creating input[type=email] or input[type=<whatever>] html tags by changing its arguments, but using form_email()/form_url()/form_search() is more obvious in code (code is more readable).

Codeigniter should support new html5 tags.


SteeveDroz:
Where is that *feature/html5* branch? This issue is 6.5 years old and still nothing in CodeIgniter allows to do that simply. Responsive web **REQUIRES** input types !

PR #1522, opened 2012-06-24
[library:log] info and debug log level index swaped

arunoda:
This is related to codeigniter Log librarary.

It has given more priority to DEBUG logs rather than INFO logs.
So this commit will fix it.

Now final log levels looks like this

` protected $_levels		= array('ERROR' => 1, 'INFO' => 2,  'DEBUG' => 3, 'ALL' => 4);`


narfbg:
With #1528 / 90726b8c769ea75aec34814ddfa91655d488e6c3, this should now be closed.

Expect major changes in CI4 (CI3 is mostly just a very large update for the CI2 code), so the actual levels order may get changed then, but for now it stays as it is, for reasons already explained.

PR #1528, opened 2012-06-25
Changed select CodeIgniter log messages from 'debug' to 'info'

alexbilbie:
Changed select CodeIgniter log messages from 'debug' to 'info' so the debug log isn't junked up

In response to discussion in #1522


narfbg:
Manually applied, see commit referenced above.

PR #1750, opened 2012-08-31
Simplification in Text Helper's character_limiter()

vlakoff:
Because the "\s" regex character class includes \r and \n,
there is no need for the str_replace() part


vlakoff:
For the record, @nicolas-grekas [brought](https://github.com/symfony/symfony/pull/32440#pullrequestreview-271193477) the following upon this cursed land:

`/(?:\s{2,}+|[^\S ])/`

PR #1818, opened 2012-09-21
HMVC with Core Unit Tests

dchill42:
This is the HMVC solution from #357, further refined and with thorough unit tests added.

The former incarnation was recommended by @philsturgeon, who suggested several features and refinements along the way. All of the substantive changes - especially the CodeIgniter class and modifications to Config, Router, and Loader - have been stable and working in a production environment for over a year and a half now.

Almost all core classes have unit test coverage, and specifically those with new features or significant changes are all covered, with strong isolation and virtually every code path tested. Convenience functions for vfsStream have also been added to CI_TestCase.

A new README in the system directory enumerates the entire application lifecycle in detail to aid in comparison with the existing bootstrap process.

This implementation aims to be fully backward compatible with version 2.1. Great care has been taken to ensure that upgrading users can continue to use the framework in the same way they used to, even if they don't take advantage of the HMVC feature.

P.S. - Don't be frightened by the number of new lines of code in this request - 2/3 of that is new unit test code.


FDiskas:
@ivantcholakov you are wrong 
HMVC
http://en.wikipedia.org/wiki/Hierarchical_model–view–controller

ivantcholakov:
@FDiskas What is this on the picture, a Java application?

narfbg:
Closing this ... even if we get HMVC in CI4, it will be on top of completely different code.

FDiskas:
@ivantcholakov The project http://bit.ly/1XPUw9q made on top of this pull request. And there is no memory issue or any other problems to call controller within other controller. It helps to easy setup an A/B testing, different design and work flows. Probably I don't get you idea.

FDiskas:
Great - on CI4 there is still no HMWC :) :shipit:

lonnieezell:
@FDiskas Not by name, but pretty damn close, anyway: 

https://bcit-ci.github.io/CodeIgniter4/general/modules.html

PR #2047, opened 2012-12-02
Active Record Call Stored Procedure

ilk33r:
calling mysql stored procedure and fix Commands out of sync problem.
This feature only support mysqli driver.


narfbg:
Closing this now, will be resolved via #3981.

Issue #1762, opened 2012-09-03
$this->dbforge->add_key for Foreign Key


andreustimm:
Still do not have this method? Why? I can create?

rejoan:
Hi, 
I have created a method add_foreign_key() for adding foreign key in existing table in dbforge library. Its working nicely, tested. Would CI authority review it? Will I submit my code here?

Thanks

EpicKris:
Will this be available in upcoming releases of CodeIgniter 3?

rejoan:
should be. donno CI authority will considering it or not

dillix:
Why AndrewPodner commits not accepted by CI Team?

THEMBOCHARLES:
To add a foreign key to the schema do the following;
Add these methods in system/database/DB_forge.php

Declare on top after primary key 
public $foreign_keys = array();//for the foreign key

`/*** FOREIGN KEY****/

	public function add_foreign_key($key, $reference_table, $reference_column, $on_update = 'NO ACTION', $on_delete = 'NO ACTION')
	{
		if (empty($on_update))
		{
			$on_update = 'NO ACTION';
		}

		if (empty($on_delete))
		{
			$on_delete = 'NO ACTION';
		}

		$foreign_key = array(
			'key'              => $key,
			'reference_table'  => $reference_table,
			'reference_column' => $reference_column,
			'on_update'        => $on_update,
			'on_delete'        => $on_delete
		);

		$this->foreign_keys[] = $foreign_key;

		return $this;
	}

  public function add_foreign_key_to_table($table, $key, $reference_table, $reference_column, $on_update = 'CASCADE', $on_delete = 'CASCADE')
	{
		$this->add_foreign_key($key, $reference_table, $reference_column, $on_update, $on_delete);

		$sql = 'ALTER TABLE ' . $this->db->escape_identifiers($table) . ' ADD ' . $this->_process_foreign_keys($table,
				TRUE);

		$this->_reset_foreign();

		if ($this->db->query($sql) === FALSE)
		{
			return FALSE;
		}

		return TRUE;
	}



	protected function _process_foreign_keys($table, $first = FALSE)
	{
		$sql = '';

		foreach ($this->foreign_keys as $foreign_key)
		{
			if ( ! $first)
			{
				$sql .= ',';
			}

			$sql .= PHP_EOL . "\t" . 'CONSTRAINT '
[...]

PR #2096, opened 2012-12-25
issue #1762 : Forge / Add Foreign Keys

AndrewPodner:
implements ability to add foreign keys in the Database Forge


colonelclick:
Consider that you also need a reliable way to delete foreign keys as well, if you want this to be useful for migrations (need a down as well as an up). In order to that correctly, I think you will need syntax to specifically name the foreign keys, otherwise you may not know the keyname to remove the key.

`CONSTRAINT my_unique_user_selected_key_name_here FOREIGN KEY...`

PR #2102, opened 2012-12-26
Extra data parameter for hidden input type.

arekpietrzyk:
Extra data parameter for hidden input type.

Example use: ID parameter for field

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -128,7 +128,7 @@ if ( ! function_exists('form_hidden'))
 	 *
 	 * @param	mixed	$name		Field name
 	 * @param	string	$value		Field value
-	 * @param	string	$extra		Extra
+	 * @param	string	$extra		Extra data
 	 * @param	bool	$recursing
 	 * @return	string
 	 */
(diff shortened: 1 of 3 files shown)


Calico90:
Ancient post but still relevant. @narfbg I noticed an inconsistency (or bug) between form_hidden and form_input (while their basically the same HTML input field). The inconsistency is also described in the documentation. https://www.codeigniter.com/userguide3/helpers/form_helper.html. 

I couldn't agree more with "array-only approach." but even with accepting the alternative (two string options) it's inconsistent to other options.

I'd like to support
`echo form_hidden(array('name' => 'fingerprint', 'id' => 'fingerprint'))`
to result in:
`<input type="hidden" name="fingerprint" id="fingerprint">`

Instead it now results in 
`<input type="hidden" id="fingerprint">`
`<input type="hidden" name="fingerprint">`

It seems this can easily be fixed with backwards compatibility since if function parameter 1 and 2 are strings, we keep the same behavior as now. If parameter 1 is an array and 2 is not set, we use the suggested array-only approach.

narfbg:
@Calico90 Nope, that won't work ... It is already explicitly made to ignore the second parameter if the first is an array.

It's stupid, very stupid. I don't know who thought what while implementing this, but it is obviously intentional.

Issue #217, opened 2011-08-20
Parser becomes a driver


philsturgeon:
Support for more systems (smarty, dwoo, etc).

PR #2158, opened 2013-01-14
Feature/parser driver - Issue #217 Parser becomes a driver

ragboyjr:
Hola brochachos! If this meets CI specs, then we just need to update the userguide src, and we be good.

RJ

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -491,6 +491,17 @@ $config['time_reference'] = 'local';
 */
 $config['rewrite_short_tags'] = FALSE;
 
+/*
+|--------------------------------------------------------------------------
+| Parser Variables
+|--------------------------------------------------------------------------
+|
+| 'parser_driver'			= The driver to load: simple driver (default)
+| 'parser_valid_drivers'	= additional valid drivers which may be loaded
+|
+*/
+$config['parser_driver']		= 'simple';
+$config['parser_valid_drivers']	= array();
 
 /*
 |--------------------------------------------------------------------------
--- a/system/libraries/Driver.php
+++ b/system/libraries/Driver.php
@@ -181,7 +181,7 @@ class CI_Driver_Library {
 				show_error($msg);
 			}
 		}
-
+		
 		// Instantiate, decorate and add child
 		$obj = new $class_name();
 		$obj->decorate($this);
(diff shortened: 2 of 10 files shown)


kabirpust:
When CodeIgniter alpha_numeric is used to validated a user name like kabir12 or kabir_sd
then if I put like 123 then It takes my input because alpha_numeric try alpha or numeric.

So I hope codeigniter v3 will be added to validated a good scope to validated a user name so that the user can not put down 123 but he can kabir123.

jim-parry:
@kabirpust What does your Feb 24 comment have to do with a parser becoming a driver?

kabirpust:
@jim-parry, Sorry sir, I am not clear your question.

jim-parry:
@kabirpust your comment sounds like you are requesting some new validation feature in codeigniter, which appears to have nothing to do with this PR. what does alpha_numeric or validation have to do with a parser or a driver?

dmyers2004:
I saw this request again and did a little thinking. How would a end user use a "module" or "package" (libraries, modules, and VIEWS which I will just call a "package" in the rest of the comment) from another  user IF this was pluggable and the user that created package foo used parser A while package bar used parser B? 

Sure the pluggable driver would use a "interface" of sorts to make sure the driver implemented the correct methods like any other driver. That's not the issue.

The issue is what happens when the tags and syntax that each driver processes in the "view" string sent in are completely different? Worse when certain features aren't in parser A which are in parser B? As a end user to use package foo and package bar in your application you would need to some how use "both" parsers. Of course the package creator could just use the CI parser but again maybe in my program I am plugging in parser C. The system would need to support multiple parsers at the same time? Maybe like what was done for caching? 

http://www.codeigniter.com/user_guide/libraries/caching.html?highlight=cache#redis-caching

ie specific a specific parser library.

Something like:

$this->parser->ci->process($view,$data); /* built in ci parser */
$this->parser->twiggy->process($view,$data); /* one called twiggy */
$this->parser->process($view,$data); /* default */

Then a "package" could use a specific one regardless of what else is loaded.

narfbg:
Even if we were to merge something like this, it'd need to be rebased, so closing.

#217 is still open as the feature request.

PR #2540, opened 2013-07-19
Ability to add HTML attributes to a table row such as class or id

dalehurley:
Updated the `_prep_args` method so data can be passed for a row intended to have attributes such as class and id.

Updated the `generate` method so attributes such as class and id can be added to rows.

``` php
    //add the row to the rows array
    $rows[]=array('data'=>$row, 'class'=>'warning');
    echo $this->table->generate($rows);
```

or

``` php
    $this->table->add_row(array('data'=>$row, 'class'=>'warning'));
```


mubeshersajid1:
Thanks, very usefull

PR #2684, opened 2013-10-16
Added possibility clear out the cached variables from the Loader.

slax0rr:
I was in a situation where I was loading a view multiple times with different vars each time, and old vars were still there, so I needed a way to remove them. Might be useful for someone else as well?


AllanJeremy:
@narfbg  Hey, I know this is an old issue, anyway, seen the fix and figured I'd add the option to disable caching for all views since calling clear_vars() after every load gets tiring at times.

In case anyone prefers to use the `clear_vars()` method, they can simply leave the caching on and use clear_vars when they'd like to. This simply adds the option to clear_vars() for every loaded view which is the scenario I've found myself in in the last 3 projects.

## Current solution
Clear the cache on every load if we have disabled variable caching

## Optimal solution
This would mean preventing the actual variable caching from happening. Since I wasn't sure whether this would break things, I used the first approach.

### Will the optimal solution break the usability?
In future, the actual usability won't be affected, the only thing that will change is where the check for caching/clearing cache occurs

Sending in a PR, hopefully it helps someone

PR #3106, opened 2014-06-16
Update Image_lib.php

linauror:
Only wm_x_transp and wm_y_transp greater than 0,function "imagecolortransparent" with execution.
Because in wm_type = overlay, wm_x_transp = 0, wm_y_transp = 0, the image have partially transparent.


ivantcholakov:
a - a.jpg
b - b.jpg

The images from the original report have been lost. I am attaching two of them again for easing reproduction and testing.

PR #3223, opened 2014-09-05
Tweak in Lang class

dmmihaylov:
Changed function line() a bit, so we can pass parameters to be parsed by translation. For example:

$this->lang->line('name', 'John');

Where:

$lang['name'] = 'How is %s?';

Result:

'How is John?'


ivantcholakov:
forum.codeigniter.com/thread-63168-post-323781.html

PR #3273, opened 2014-10-15
set_item in another index

jensdeschrijver:
Added the ability to set an item inside another index.


narfbg:
Also, it has been suggested before and rejected due to parameter creep and still being a limited functionality (there may be countless nesting levels in an array).

Issue #92, opened 2011-08-19
get_dir_file_info() the return array's key needs to be unique


bitbucket-import:
The get_dir_file_info() function in the file helper has a bug in the return array being generated.

$file is being used as the return arrays key. But this in not unique enough in that If I have 2 similarly named files in different folders, the first files data is overwritten by the secong eg.

**I have the following files:**

assets\bingopizza_dbn\language\english\sitetext_lang.php

assets\bingopizza_dbn\language\spanish\sitetext_lang.php

**The english files data is missing.**

**If I call the function as per below:**

get_dir_file_info('assets\bingopizza_dbn\language\', $top_level_only = FALSE)

**The result returned is:**

Array
(
    [sitetext_lang.php] => Array
    (
        [name] => sitetext_lang.php
        [server_path] => ...assets\bingopizza_dbn\language\spanish\sitetext_lang.php
            [size] => 31
            [date] => 1285930111
            [relative_path] => ...assets\bingopizza_dbn\language\spanish\
    )
)

gadelat:
My opinion is that this helper should not use filenames as keys. Accessing filename from array is easy enough even without this "feature". It should be changed to use numeric keys as is proposed by @Indigo74 . Yes, it will be BC-break. I still think this should be done.

PR #3347, opened 2014-11-20
#92 `directory_map()` alternative for `get_dir_file_info()`

mwhitneysdsu:
- Added optional parameter to `directory_map()` to call
`get_file_info()` on each file and add the result to the function's
output.
- Documented limitation of `get_dir_file_info()` when multiple files in
the directory hierarchy share the same name, and recommended use of
`directory_map()` as a possible solution.

This is not necessarily a fix for #92, but a work-around to maintain
backwards compatibility (in both functions) and make it relatively easy
to gain access to this information under these conditions.


narfbg:
... and with no feedback other than mine, I guess it's up to me.

A better solution would be a separate `get_dir_file_info()`-like function ... the existing file and directory helper functions suffer from parameter creep and adding work-arounds via new parameters to them only makes it worse.

PR #3450, opened 2014-12-29
Bump docs to suggest 5.3 is used

philsturgeon:
A while back when I was on the team I suggested we moved from 5.1 to 5.2. This made sense two or three years ago when this branch was meant to be "out soon" but time has moved on a little bit. 

When people first started grumping about 5.2 being used I explained that I didn't think it was so bad for various reasons. I know that people use CodeIgniter on versions of PHP that are WAY out of date, and I know that even 3 years ago moving from 5.1 to 5.2 pissed people off. 

What has made me spin on a dime here is that the .travis.yml does not actually test against 5.2 at all, and of course it can't because vfStream has a [5.3 dependency](https://github.com/mikey179/vfsStream/blob/fc0fe8f4d0b527254a2dc45f0c265567c881d07e/composer.json#L7). 

On top of that, PHPUnit probably needs 5.3 too depending on the version travis is supplying you with.

SO! 

If you aren't testing against 5.2, you really really cannot release with docs saying that you support 5.2. CodeIgniter was always a cowboy framework with zero tests and that is turning around. In the past were dangerous untested releases done without warning. It was a bit of a nightmare, and I had to run back from the pub one day to patch and retag a version that somebody at EllisLabs had released without warning which had some bugs on one version of PHP. Without 5.2 in the `.travis.yml` version matrix once again I'd say you absolutely cannot say you support 5.2.

Instead of trying to find a way to test on 5.2, I'd urge you just to pop up the requirements to 5.3.

This may well still work on 5.2 and the only place it will fail is if a developer tries to run `composer update`, but that is only for testing, and the tests wont work anyway.

I'm not going crazy here. I think PHP 5.4 would be nice, but I know that these things need to be done slowly. If you lot want to go further than carry on, but everyone else should try and avoid complaining that the level of progress is not enough.


narfbg:
During our internal discussion, we've decided to advertise PHP 5.4+ as recommended, but still keep 5.2.4 as the absolute minimum requirement.

This is kind of the middle ground and I don't expect everybody to understand and/or approve the decision, but I'm sure those familiar with CodeIgniter's history and current situation would do so.

Frankly, if we were to drop support for versions lower than 5.4.x, knowing that if we had done this from the start we'd have a completely different framework now, I'd rather throw away the whole release. That would be more harm done to those who have waited for it basically forever.

Just so it is clear, I'm not a fan of supporting old versions forever, so I'll be happy to put more emphasis on recommending 5.4 throughout the documentation.

philsturgeon:
I am not super happy with this result. While I do not care about 5.3 or 5.4 being a minimum, you cannot reasonably say to people that you support PHP 5.2.4 when you have no idea if it will work or not. 

I HATED this about CodeIgniter for so long. It was a nightmare working on a project that had no unit testing, but I didn’t have the months required to work on it. 

Now that you have a test suite it is finally time for CodeIgniter to be a little bit more grown up. You can reliably say “We know for a FACT this test suite passes on 5.3, 5.4, 5.5 and 5.6. In the past we had to make speed-patches to fix stuff after updates broke things on a specific version in really bizarre ways.

So please, please, PLEASE reconsider writing anywhere in documentation that you support 5.2.4, because without tests you 
actually don’t and can’t. 

You are telling the CodeIgniter community and all of the students involved with maintaining it that “It works on my machine.” is an ok way to ship software, and it REALLY REALLY is not. At all.

Please merge this pull request.

benedmunds:
@narfbg can correct me if I'm wrong but v3 does pass on 5.2. Just not with Travis.

GrahamCampbell:
@bobdenotter How can it when phpunit requires php 5.3...

GrahamCampbell:
Unless the test suite still works on phpunit 3.6?

philsturgeon:
If that was the case somebody probably should have mentioned it WAY sooner in the conversation. 

But yes, PHPUnit requires 5.3 and unless you want to roll back PHPUnit to an earlier version and update the whole test suite to work without any of the features lost in that rollback, then… it has to be 5.3.something.

benedmunds:
Hmm there was talk internally of it working and I assumed it was the tests. Maybe they just tried it local.

Can you chime in here @narfbg since I'm not around my laptop atm

narfbg:
@benedmunds What I said in the internal discussion was that it (the test suite) *could* work on 5.2, but it would require a major effort that's not worth in practice.

@philsturgeon

There are better arguments for dropping 5.2.x support than this, @ircmaxell's being the most compelling one. I'd love to have it all covered by unit tests, but not having them doesn't mean that I "don't and can't support" something. Not to me anyway ... TDD isn't my religion.

As far as I'm concerned, any increase in the required version will result in a "PHP 6" scenario.

philsturgeon:
@narfbg this has nothing to do with TDD.

TDD is test first process, where the tests literally drive your development. 

There are other test based methodologies such as "spike and stabilize," which means you work on your tests once a reasonable chunk of work has been done, and get your coverage up before you ship. 

They both assume that at one point you are actually running tests. That is the main thing. **Run your tests.**

For 5.2, you are not running tests. This is almost identical to **not having tests.**

You are releasing software to a huge number of people and saying "I have no idea if this works."

You are telling the PHP community that you think "I mean... I clicked around a bit locally" is a good way to release software.

That moronic cowboy-coding mentality worked for EllisLab when they were building this between 2006-2011, but it should **not** work for you now this product has started active development once again and is being released in 2015.

Especially now as you've all spent the time to build such a huge test suite. It blows my mind that you would be happy with making all of these tests, then release the software on an untested version, utterly negating everything you worked so hard to achieve in the 3.x branch over the last few years. 

Dropping PHP 5.2, wether due to being utterly untested and having no idea if it works at all, or wether its due to security as Anthony points out, is an absolute must here. Use whichever reason you would like to use more. Hey, use both reasons. But you should _absolutely_ drop 5.2. 

I don't care if you go 5.4, but releasing on 5.2 is completely feckless behavior and I cannot condone it in the slightest. It'll be just one more terrible decision in CodeIgniters long and colorful history of making completely awful decisions. 

Do not do this.

GrahamCampbell:
I totally agree with @philsturgeon. This decision makes totally no sense to me. Supporting 5.2 is defined by your test suite running on it.

narfbg:
Phil, I know what TDD is ... it's the notion of automated testing being the most important thing that I meant with that.

Yes, it sucks that the lowest required version is not automatically tested. But you know what? The test suite sucks anyway. It sucked back when you merged it and it is almost untouched since you left the project because it's a PITA to work with.
FFS, it had assertions that *validated wrong results* because the guy who wrote it just copy-pasted the return values that were current at the time. I for one don't trust it, with the exception of a few tests that I wrote myself.

I too am unhappy with some of EllisLab's decisions, but you can't override them post factum when the hard work has already been done and *based* on them. Now you're saying that this *will be* another bad decision, but you know probably better than I do when exactly the decision to write code for PHP 5.2.4 *was* made. And that's exactly what has been done for the past few years - write code for PHP 5.2.4.

Countless patches and feature requests have been rejected over that. A lot of time spent trying to make something work on 5.2 will be wasted. CodeIgniter has been called dead, irrelevant (this one by yourself), a "shit project" and whatnot, because it's not "modern" like the other 100 frameworks out there.

Do I like it? Absolutely not.
Do I agree that testing is important, as well as using a maintained version of PHP? Absolutely, that's why we agreed to recommend usage of 5.4.

And I am biased, but I'm going to be extremely unhappy if this code gets released with the stamp "requires PHP 5.[3-4].x", yet it doesn't use any of the awesome features that a more recent version provides.
I will favor a PHP 6 situation over that. Call me whatever you wish, but seriously, I'd rather revert all breaking changes and release what's left of it as version 2.4, because that's all that it is anyway. All the hype is literally about what number is written somewhere and you all probably wouldn't give a shit if we were talking about CI 2.4.

philsturgeon:
> Phil, I know what TDD is ... it's the notion of automated testing being the most important thing that I meant with that.

That is absolutely not what TDD is. 

> Test-driven development (TDD) is a software development process that relies on the repetition of a very short development cycle: first the developer writes an (initially failing) automated test case that defines a desired improvement or new function, then produces the minimum amount of code to pass that test, and finally refactors the new code to acceptable standards. [Wikipedia](http://en.wikipedia.org/wiki/Test-driven_development)

Anyway, enough of semantics.

> Yes, it sucks that the lowest required version is not automatically tested. But you know what? The test suite sucks anyway. It sucked back when you merged it and it is almost untouched since you left the project because it's a PITA to work with.

I am amazed to hear that tests have been ignored throughout the development progress in the two years since I left.

> FFS, it had assertions that validated wrong results because the guy who wrote it just copy-pasted the return values that were current at the time. I for one don't trust it, with the exception of a few tests that I wrote myself.

It's pretty common to validate what the system is doing, then check those assumptions are correct later on. 

Whilst that does sound awful, I would absolutely prefer to have a test suite (shitty or not) that can be improved over time, rather than flippantly just saying "yeah it probbly works" and throwing it out in front of thousands of people with its zipper undone. 

> I too am unhappy with some of EllisLab's decisions, but you can't override them post factum when the hard work has already been done and based on them. Now you're saying that this will be another bad decision, but you know probably better than I do when exactly the decision to write code for PHP 5.2.4 was made. And that's exactly what has been done for the past few years - write code for PHP 5.2.4.

Lets be fairly clear on this specific part, EllisLab didn't make any decisions about PHP version usage other than "Ok ok we'll drop PHP 4 now." is about 2010. 

I pushed the decision to jump to PHP 5.2 because it was absolutely time. There were enough people using it, enough distros on it, and it was fine to do so for a major version.

I also made that call f**king ages ago, and made it before the test suite was merged.

I am not saying you lot are wrong to continue along the goals of what 3.x was for a while, but I am saying this:

1. 5.2 was never set in stone when the 3.0 branch was conceived. It changed based on the community and other various factors.
2. A 2 year old decision does not need to be maintained if other valid arguments are made against it.
3. Releasing code without ANY tests is dangerous, and thats an extremely valid argument.

> Countless patches and feature requests have been rejected over that. A lot of time spent trying to make something work on 5.2 will be wasted. CodeIgniter has been called dead, irrelevant (this one by yourself), a "shit project" and whatnot, because it's not "modern" like the other 100 frameworks out there.

I think you've missed the point of what im getting at. "Supported version" and "Minimum version" can both be different here.

You can absolutely say "This might still work on PHP 5.2, but it is not officially supported."

I am saying that without tests, its not appropriately supported, and thats the whole damn point here. It's about setting expectations. 

If you say "It might work on 5.2 but we don't really know, do so at your own risk." then the expectations you set are in line with what you can accurately promise people.

You cannot promise people your code works on 5.2, so you really should not try.

> And I am biased, but I'm going to be extremely unhappy if this code gets released with the stamp "requires PHP 5.[3-4].x", yet it doesn't use any of the awesome features that a more recent version provides.

I know, I argued that very case on the last PHP Town Hall, and have written in the past that bumping versions without utilizing features is a mess, but those opinions are totally overridden when faced with the alternative: lying to a huge community of people, saying your software works on a version of PHP whilst having no idea if it does or not.

That might have been acceptable in 2006-2011, but it is absolutely not acceptable now.

And again, a crappy but improving over time test suite is a bazillion times better than no test suite at all.

dbernar1:
@philsturgeon I don't understand why you're getting all worked up about this. @narfbg says he wants to stay at 5.2.4 for now.

If you have a problem with how unit testing/QA is done, make a pull request for that. No?

philsturgeon:
I'm not at all worked up, I am trying to explain why saying "It works on this version" whilst not actually having a clue if it does or not is not how things need to be done in the community.

If you can get the tests working on PHP 5.2 I'll be perfectly content, but that does seem like an exercise in futility when you could just merge this and tell people:

> Maybe it'll work on 5.2, but if you can upgrade to 5.3 or 5.4 you'll have a much better chance of success. Oh and they're more secure too so double win.

GeeH:
How to lose all credibility in an open source project in one pull request.

dbernar1:
How to feed the trolls...

philsturgeon:
Who exactly is a troll in this instance?

GrahamCampbell:
> Who exactly is a troll in this instance?

Good point.

ircmaxell:
escalated_quickly

ivantcholakov:
@philsturgeon

http://blog.ircmaxell.com/2014/12/being-responsible-developer.html

I can see the number about unit tests effecitveness on defect detection, it is 30%. So, these test are not definitive for a claim "It supports whatever". I understand "It supports" as a promise eventually - if a bug appears there will be somebody to fix it. There is no drama.

GeeH:
@dbernar1 I completely understand if you think I am trolling, but I'm genuinely not, it makes me feel sad when I see a credible OS project making bad decisions and losing that credibility.

Granted, this isn't really the place for random opinions or observations.

GrahamCampbell:
@GeeH Agreed.

ircmaxell:
For the record, I would like to comment on @narfbg 's original close reason:

> During our internal discussion, we've decided to advertise PHP 5.4+ as recommended, but still keep 5.2.4 as the absolute minimum requirement.

I think this is a somewhat reasonable compromise. The only reason I say somewhat is that I still believe actually advertising 5.2.* as supported is still negligent.

However, if you're not willing to raise that requirement, then recommending 5.4+ is a good step in the right direction.

Thank you for the consideration.

philsturgeon:
@ivantcholakov If people are 30% confident about the tested versions, that's 30% better than the rest.

> We are 30% confident that this software works on PHP 5.3, 5.4, 5.5 and 5.6, but we haven't got the slightest of foggy clues about PHP 5.2 because we don't even run our test suite there, but... I mean go ahead and use it it's fine.

That is the message CodeIgniter is sending to developers.

Wonderful.

Ocramius:
Related: zendframework/zf2#7095

philsturgeon:
@Ocramius I'd like to try and avoid crossing the streams here. The arguments for bumping CI to 5.3, and to bump ZF to 5.4 are a little different. This is less about hand holding people and not letting them use a version of software thats pretty damn old now, its more about being accurate with the portrayal of which versions of PHP this software will even work on.

(That said a jump to 5.4 would be nice too, but lets not try and worry about that just yet. Dropping 5.2 is much more important.)

narfbg:
@philsturgeon 

> Whilst that does sound awful, I would absolutely prefer to have a test suite (shitty or not) that can be improved over time, rather than flippantly just saying "yeah it probbly works" and throwing it out in front of thousands of people with its zipper undone.

Agreed. I mentioned the same argument in our internal discussion.

> Lets be fairly clear on this specific part, EllisLab didn't make any decisions about PHP version usage other than "Ok ok we'll drop PHP 4 now." is about 2010.
>
> I pushed the decision to jump to PHP 5.2 because it was absolutely time. There were enough people using it, enough distros on it, and it was fine to do so for a major version.
>
> I also made that call f**king ages ago, and made it before the test suite was merged.
>
> I am not saying you lot are wrong to continue along the goals of what 3.x was for a while, but I am saying this:
>
> 1. 5.2 was never set in stone when the 3.0 branch was conceived. It changed based on the community and other various factors.
> 2. A 2 year old decision does not need to be maintained if other valid arguments are made against it.
> 3. Releasing code without ANY tests is dangerous, and thats an extremely valid argument.

It doesn't matter if EllisLab made the decision or if they silently agreed with yours (actually, that is one good thing that they often did).

As you said, the decision was made ages ago; and the point is - job has already been done, it's too late to change it. The difference in what you and I believe this change would imply is the key - your POV affects documentation only, mine affects the codebase.

> I think you've missed the point of what im getting at. "Supported version" and "Minimum version" can both be different here.
> 
> You can absolutely say "This might still work on PHP 5.2, but it is not officially supported."
>
> I am saying that without tests, its not appropriately supported, and thats the whole damn point here. It's about setting expectations.
>
> If you say "It might work on 5.2 but we don't really know, do so at your own risk." then the expectations you set are in line with what you can accurately promise people.
>
> You cannot promise people your code works on 5.2, so you really should not try.

Well, "supported version" and "minimum version" are the same to me.

That being said, wording *can* be changed. Saying something in the lines of "It should work on PHP 5.2.4, but do so on your own risk because we can't guarantee it" is fine with me (similar to how telecoms advertise maximum downloads speed, but explicitly say that they can't guarantee them). That's what recommending 5.4 means to me and I already said in my first reply that I'm all for putting more emphasis on that.

The only promise that we give about 5.2 is that we will *try* making it work and non-optional functionality that doesn't work on 5.2 will be considered a bug.

@ircmaxell

Thank you for noting that, you're the first person here not involved in the project that actually tries to see both sides of the coin. I appreciate that.

Ocramius:
@philsturgeon I actually linked it in because:

 * people are not that worried about deploying 5.4 to production, they actually see it as a responsible thing to do
 * the discussion about semantic versioning that we had: bumping requirements doesn't need a major version bump, therefore there is no need for that big fuss about dropping 5.2 completely.

dbernar1:
Take a look at this: https://travis-ci.org/dbernar1/CodeIgniter

Looks like it might be possible to have the unit tests run on 5.2 as well. I will update you on the progress I make.

philsturgeon:
@narfbg we're getting closer to the same points. :)

> As you said, the decision was made ages ago; and the point is - job has already been done, it's too late to change it. The difference in what you and I believe this change would imply is the key - your POV affects documentation only, mine affects the codebase.

Leveraging PHP 5.3 functionality and saying that you will only offer active support for 5.3 and up are two totally different things. That is the crux of our difference on this view.

I'm just saying you should not call it a supported version, and warn folks away from using 5.2. Security is one argument as others have said, but the fact your test suite doesn't run against it is another. 

You don't need to get into the nitty gritty of why in your docs, but you really can't just sit there and say "It's fine on 5.2." when you don't know that. 

1. Warn people away from 5.2
2. Soft push towards 5.4
3. Work out if you are comfortable saying 5.2 is supported without tests

It sounds like 2 is being done already and that makes @ircmaxell happy. If instead of this PR you implement 1 with some wording in the docs then I'll be happier too.

3... seems like your mind is made up.

ircmaxell:
@philsturgeon It doesn't make me happy. But it's a better situation than existed a few days ago. I acknowledged the step. That doesn't mean it's an end-game or an ideal. Just that it's a step in the correct direction.

philsturgeon:
@dbernar1 I don't think anyone doubted it was possible, but I can already see you making changes to the test suite to rip bits out and make the tests run and that's exactly the sort of wasted effort I'm talking about.

Again:

> Maybe it'll work on 5.2, but if you can upgrade to 5.3 or 5.4 you'll have a much better chance of success. Oh and they're more secure too so double win.

Not those exact words obviously, but that general message. It's the perfect solution. Your 5.2 users are not shunned, they just have realistic exceptions, you're pushing people in the right direction softly, and if they dont bother to upgrade thats up to them.

An even better message would be:

> Server Requirements
> *******************
>		 
> PHP version 5.3.3 is the lowest supported version.
>
> CodeIgniter is tested on PHP 5.3, 5.4, 5.5, 5.6 and the latest version of HHVM, but PHP 5.4 or above is recommended. PHP 5.2 may also work, but is wildly outdated, untested and using it can lead to security issues or unexpected issues.

dbernar1:
@philsturgeon It's almost done now, and very little effort was put into it. Much less effort than you've expended to write all the text you sent to this thread.

As far as "making changes to tests to rip out bits", I encourage you to take another look at the diff: https://github.com/dbernar1/CodeIgniter/commit/5684027b247b55e4fbe7caa341eb646606ad33be because it is just one little thing. Not anything else.

It is just that quoted_printable_encode() stuff. What is that anyway? I will put that back in, it was just a temporary convenience so I can see whether there is a lot of work needed to get 5.2 testing working. Now that I know that is the only test in the whole suite that has an issue, I will go back and fix it up.

dbernar1:
OK, so this is a test that never actually ran before, since it is a test for versions lower than 5.3.

```
if (is_php('5.3'))
{
  return $this->markTestSkipped('quoted_printable_encode() is already available on PHP 5.3');
}
```

I'm gonna suggest removing the part of the test that is failing in the pull request I will make for adding 5.2 unit testing support, and continue the conversation there, regarding whether that is the right approach for that particular test to avoid hijacking this thread further.

philsturgeon:
@dbernar1 remember when I said "I don't think anyone doubted it was possible." I also didn't think it would be a months worth of work. I did however point out it would be an exercise in futility when you could just put a note on your documentation pointing out that using PHP 5.2 might just be a shit idea.

So long as your supported version list matches up with what you are actually testing against then fine, but this outcome is much like picking the nicest looking turd in the dog park.

Do as you wish. I'm out.

narfbg:
We'll get 2 for 1 then it seems ... We'll improve the wording and @dbernar1 has the tests working on 5.2.

PR #3452, opened 2015-01-01
Update Cache_memcached.php typo error

cweric:


system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -295,4 +295,4 @@ public function is_supported()
 }
 
 /* End of file Cache_memcached.php */
-/* Location: ./system/libraries/Cache/drivers/Cache_memcached.php */
\ No newline at end of file
+/* Location: ./system/libraries/Cache/drivers/Cache_memcached.php */
```
narfbg: Please remove the empty line here (you have to use a local editor, can't happen through GitHub's).

ZombieProtectionAgency:
@narfbg Sorry if i shouldn't ask this here, but I have been wondering since yesterday and since it came up....
>you have to use a local editor, can't happen through GitHub's 
 
How exactly do you do that? Is there any way to edit a single file with a local editor or do you have to install git and everything locally first?

vlakoff:
This issue should have been kept open, @narfbg you might want to directly fix the bug. It's the most straightforward.

narfbg:
No shit. :)

PR #3465, opened 2015-01-11
Added composer.lock to .gitignore

dbernar1:
Hi,

Is this file supposed to be in .gitignore? It seems it should be in there, but I'm not sure.

If you do want to merge this in, please let me know if I should change anything before you can accept it.

Thanks for sharing your software.

Signed-off-by: Dan Bernardic <dan.bernardic@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ user_guide_src/cilexer/dist/*
 user_guide_src/cilexer/pycilexer.egg-info/*
 /vendor/
 /nbproject/
+composer.lock


narfbg:
If you're not sure, don't suggest it ...

ivantcholakov:
@dbernar1 Better don't suggest this.
https://philsturgeon.uk/php/2014/11/04/composer-its-almost-always-about-the-lock-file/

dbernar1:
@ivantcholakov Thanks. I did actually know what @philsturgeon wrote there from before. Also from a ruby project I did a few years back :)

So we are looking at CI as an application rather than as a component. OK. Then the .lock file should be committed in the repository, yet it isn't.

@narfbg If I make a pull request with the .lock file, are you going to merge that one in?

narfbg:
And what exactly are you going to put in it?

dbernar1:
@narfbg I think I am starting to see why this is a non-issue to you. The composer.json file specifies the vfs library, which is used only by the tests, which you don't actually run locally at all. Is that about right?

That's cool, I'm still just trying to understand the project and don't know what matters and what doesn't yet. Thanks for your patience and time.

Just in case, to answer your question, the .lock file is generated when I run `composer install`, and it contains some composer jibberish.

narfbg:
I know what gets generated in it, and yes - CI doesn't depend on any of it. That's why I asked.

philsturgeon:
It’s really not gibberish, its a way to specify exactly which version of the code should be installed, from which source, and using which checksum to check 100% integrity of the files. 

CodeIgniter does not need this for roughly the same version a simple component does not. The requirements should be a little loser as its designed to work with a wide range.

On top of that, all you lot use are testing things so its exceptionally unimportant in this instance.

Keep ignoring it :)

ivantcholakov:
@philsturgeon

Let me explain may hesitation, I have a question about https://philsturgeon.uk/blog/2012/05/composer-with-codeigniter/ ?

If somebody did in the past these steps with CI2, then what is he/she going to do on upgrade to CI3? To overwrite the files? Then ignoring does not seem as an option.

Or maybe on upgrade it would be better he/she to move the whole thing under application/

philsturgeon:
@ivantcholakov If somebody already has a `composer.json` then they will of course need to merge the two files. If they already have a `composer.lock` (of course they will) then that doesn't matter, as it will be overridden when they run `$ composer update`

If you are concerned that you putting `composer.lock` into your `.gitignore` file will have any negative affect on those end users, then forget about it. If the end user has already added their `composer.lock` to their version control (here's hoping they have!!) then Git will continue to track changes to that file. If they have not added `composer.lock` to their `.gitignore` then nothing will happen. 

You really do not want to have two `composer.json` files and two sets of `vendor/`. That sounds awful. :)

ivantcholakov:
@philsturgeon Thank you. I've just noticed that /vendor/ is inside .gitignore file, probably only the component for testing is meant to be used. In this case it seems to me that composer.lock should be added to .gitignore too.

dbernar1:
@ivantcholakov awesome. I guess we've all learned something here then ;P

PR #3468, opened 2015-01-11
Added support for running unit tests on PHP 5.2

dbernar1:
Signed-off-by: Dan Bernardic <dan.bernardic@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 php:
+  - 5.2
   - 5.3.3
   - 5.4
   - 5.5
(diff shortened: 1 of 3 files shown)


dbernar1:
@narfbg I know this is not very high on your priorities list, but FWIW.

narfbg:
Priorities or not, this is fucking great in comparison to the #3450 discussion (and to how it exploded in my face when I tried to run the tests on 5.2).

Why are you changing DB_driver.php and the quoted_printable_encode() test though?

dbernar1:
I'm really glad you find it useful :)

DB_Driver's implementation of the field_data method had the signature `public function field_data($table)`, and all the sub-classes had the signature `public function field_data($table='')`.

When I ran the tests with 5.2, PHP complained about the mismatch. The change should be a "true refactoring", by which I mean that there is no way that the change would change existing functionality at all.

The quoted_printable_encode test was written wrongly, which just never came up before because the test is skipped on all versions higher than 5.2.

I hope that answers it. Otherwise, I can submit a pull request without those changes, and then you could review the Travis log, and see in what way the build fails without those changes.

narfbg:
Yes, please revert the field_data() and quoted_printable_encode() changes.

The former is a recent change and any extending classes should be changed instead, while the latter is copied from PHP's own test cases, so chances are - it's the implementation that's wrong and not the test.

dbernar1:
OK, I did, and the tests fail for those two reasons. Do you want me to change all the sub-classes? 

That is a more risky change, as I understand it, because any code that relies on that default parameter will then break.

Did you write the implementation of quoted_printable_encode() yourself, or did you copy it from somewhere?

dbernar1:
It looks like the test is actually wrong. Here's a codepad I prepared to test that:

http://codepad.viper-7.com/8INDk3

narfbg:
The implementation is mine, only the test is copied. I'll look into that myself though ... thanks for the whole effort.

narfbg:
dbf3460bc7c94bf49db7dff3525906cae53a81ac

PR #3471, opened 2015-01-12
Correctly determine setting of display_errors configuration variable.

trinet-at:
display_errors is a string since PHP 5.2.4, so on certain occasions a disabled display_errors setting could be misinterpreted as enabled.

Signed-off-by: Thomas Rieschl <thomas@trinet.at>

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -615,7 +615,7 @@ if ( ! function_exists('_error_handler'))
 		$_error->log_exception($severity, $message, $filepath, $line);
 
 		// Should we display the error?
-		if (ini_get('display_errors'))
+		if (_get_display_errors())
 		{
 			$_error->show_php_error($severity, $message, $filepath, $line);
 		}
@@ -650,12 +650,42 @@ if ( ! function_exists('_exception_handler'))
 		$_error->log_exception('error', 'Exception: '.$exception->getMessage(), $exception->getFile(), $exception->getLine());
 
 		// Should we display the error?
-		if (ini_get('display_errors'))
+		if (_get_display_errors())
 		{
 			$_error->show_exception($exception);
 		}
 
 		exit(1); // EXIT_ERROR
+	}
+}
+
+// ------------------------------------------------------------------------
+
+if ( ! function_exists('_get_display_errors'))
+{
+	/**
+	 * display_errors ini setting
+	 *
+	 * Gets the real setting of the display_errors configuration variable.
+	 * parse_ini_string() is only available with PHP 5.3.0 or greater
+	 *
+	 * @return	boolean
+	 */
+	function _get_display_errors()
+	{
+		if (function_exists('parse_ini_string'))
+		{
+			$display_errors = parse_ini_string('display_errors = ' . ini_get('display_errors'));
+			return (bool)$display_errors['display_errors'];
+		}
+
+		$display_errors = strtolower(ini_get('display_errors'));
+		if (empty($display_errors) OR in_array($display_errors, array('null', 'off', 'no', 'false', '0')))
+		{
+			return FALSE;
+		}
+
+		return TRUE;
 	}
 }
 


narfbg:
The value is overriden in index.php, it's not worth it.

trinet-at:
Well, not always
I usually set the `display_errors` value at Apache vhost level or via FPM pool settings as `php_admin_value`, so it's not allowed to be changed on runtime. The setting in `index.php` is then ignored.
I discovered the problem on a production website, where the setting was `php_admin_value display_errors off` in the vhost.
CodeIgniter displayed the errors anyway, because `if (ini_get('display_errors')) { }` evaluated as `true`, because the value was `string(3) "off"`
This is a quite sensitive setting so it should be ensured, that the correct value is determined, don't you think? `parse_ini_string()` is the safest method for this.

narfbg:
Well, the way I see it, this is rather an edge case caused by misconfiguration ...

The `display_errors` option was *never supposed* to be used as a string, it's a boolean value that got changed only to make it possible to redirect errors to STDERR or STDOUT. "off" (as a string) will only be returned if you quote the value when you set it. No quotes around "off" (or any of the other negative values) will result in '0' or an empty string being returned.

I'd rather encourage proper configuration than write code to satisfy every single edge case.

That being said, (string) 'off' is probably a more common case than any of the other values that your patch tries to work with, so a simple `str_ireplace('off', '', ini_get('display_errors'))` should fix 99% of the false positives, but the rest IMO is just bloat.

trinet-at:
On a Debian Squeeze machine with Apache and PHP from the repository, the setting `php_admin_value display_errors off` (without the quotes) in Apache vhost will result in a `(string) off`, so I don't think it's that much of an edge case.

Either way, as per PHP configuration, it's allowed to use no, off, 0, false and so on, even without the quotes as a valid value, so it's no misconfiguration, no matter if it never was supposed to be a string.
If I write `php_admin_value display_errors no` in my vhost, it's perfectly valid and PHP interprets it to not display errors, and so should CodeIgniter, because I expect it to behave the same was as PHP does.
If you want, I can add a check for `empty()` as first condition for the 5.3. version. It's a bit faster than the `parse_ini_string()`

If you don't want to implement that, would you consider updating the documentation and mention, that if errors shouldn't be displayed, the `display_errors` setting has to be `0`, and not `off` or anything else?

As for 99% of the false positives it'll be `off`, that might be true, but also checking the other values just takes a few chars and doesn't *bloat* the code and as I mentioned before, those are all valid settings.

narfbg:
> Either way, as per PHP configuration, it's allowed to use no, off, 0, false and so on, even without the quotes as a valid value, so it's no misconfiguration, no matter if it never was supposed to be a string.
If I write `php_admin_value display_errors no` in my vhost, it's perfectly valid and PHP interprets it to not display errors, and so should CodeIgniter, because I expect it to behave the same was as PHP does.

It's valid as a by-product of how PHP's INI parser works, but any documentation will suggest usage of 0/1 or On/Off. And I've personally never seen anybody use any other values.
I'm not saying CodeIgniter should ignore how PHP works, but you want to go into extreme lenghts for that, no matter how useless it may be.

> If you want, I can add a check for `empty()` as first condition for the 5.3. version. It's a bit faster than the `parse_ini_string()`

I'm not concerned about performance when displaying errors. :)

> If you don't want to implement that, would you consider updating the documentation and mention, that if errors shouldn't be displayed, the `display_errors` setting has to be `0`, and not `off` or anything else?

A note about `php_admin_value` would be OK (as that's the directive causing the issue, standard INI file settings behave fine), and one about values in quotes too. 0 instead of Off in general would be too much.

> As for 99% of the false positives it'll be `off`, that might be true, but also checking the other values just takes a few chars and doesn't *bloat* the code and as I mentioned before, those are all valid settings.

Based on my `str_ireplace()` suggestion - sure, fitting those few more characters in there is fine. Writing 10 more lines of code to handle values that nobody uses is what I call bloat. :)

trinet-at:
> It's valid as a by-product of how PHP's INI parser works, but any documentation will suggest usage of 0/1 or On/Off. And I've personally never seen anybody use any other values.

That might be right, but when I use `Off`, CodeIgniter behaves incorrectly at the moment, because `(string) Off` evaluates as `true`

> I'm not saying CodeIgniter should ignore how PHP works, but you want to go into extreme lenghts for that, no matter how useless it may be.

Don't you think it's worth it, if it prevents 1%, of all CodeIgniter users to accidentally leak possibly sensitive information, including myself? Isn't it one of the advantages of using a framework that it catches such quirks so that the developer isn't bothered with it? Look what jQuery had to do to overcome lots of browser quirks. It's essentially the same.
*Extreme lenghts* may be exaggerated :smirk: It's really just one line of code. But because of the support for PHP 5.2.* there's the fallback, because `parse_ini_string()` isn't available until 5.3.0, but it's the correct function to use, because it's PHP's own ini parser. So, if future PHP versions change the behavior, the code still works. And, if I check the output of `phpinfo()` it's *always* the same as from `parse_ini_string()`

> A note about php_admin_value would be OK (as that's the directive causing the issue, standard INI file settings behave fine), and one about values in quotes too.

Yes, `php_admin_value` is causing the issue (the *admin* portion), but I think it's often used. I've seen some hosting providers where you can set that in the admin backend and it's put in the Apache vhost (not `php.ini` because of the shared environment). And it's the *_admin_* setting that you'll want to set in the vhost, because otherwise you could just put it in the `.htaccess` file or some bootstrap file. It's the fact, that it **can't** be overwritten somewhere else, because it's still sitting as development setting somewhere in the code.

> 0 instead of Off in general would be too much.

It wouldn't be enough because currently using `Off` would *not* disable outputting the errors.

narfbg:
You're arguing for the sake of it. This is as far as I go: 38666660a0d3b9a91cffb27c0fb519340e76ba5d

trinet-at:
> You're arguing for the sake of it.

That's not true :smirk: 
It's just that it happened to myself and it took me a long time to figure out why errors are still displayed. I just wanted to prevent others from having the same problems. When it comes to security related stuff, I tend to implement it more thoroughly.
Your implementation of course also fixes the bug, so.. thanks :smile:

PR #3474, opened 2015-01-13
Strings are truncated when containing %E*

erwandf:
forcing utf-8 so the strings are no more truncated when passed to rawurldecode, we must transform it before so it will be interpreted the good way.

This is an error we had with french characters at Whaller (https://whaller.com)

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -29,10 +29,10 @@ If you change anything that requires a change to documentation then you will nee
 
 ### Compatibility
 
-CodeIgniter recommends PHP 5.4 or newer to be used, but is is compatible with
-PHP 5.2.4 so all code supplied must stick to this requirement. If PHP 5.3
-(and above) functions or features are used then there must be a fallback for
-PHP 5.2.4.
+CodeIgniter recommends PHP 5.4 or newer to be used, but it should be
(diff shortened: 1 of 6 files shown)


narfbg:
Who says they are always UTF-8?

thomasfaure:
@narfbg what do you think if we add [mb_check_encoding](http://php.net/mb_check_encoding) function in order to test if the string is a UTF8?

narfbg:
If anything should be tested, it's the configured application charset and the incoming request headers.

And now that I have a second look at this, I don't see the difference at all:

<pre>
$ php -a
Interactive mode enabled

php > var_dump(rawurldecode(ord(utf8_decode('%EE'))));
string(2) "37"
php > var_dump(rawurldecode(ord('%EE')));
string(2) "37"
</pre>

('%E*' is certainly not even passed through `rawurldecode()`, only strings matching `/%[0-9a-f]{2,}/i` are.

Also, the most upvoted comment about `utf8_encode()`, `utf8_decode()` says that doing this is absolutely wrong: http://php.net/utf8_encode#104906

narfbg:
On another note, judging by the commit messages, the OP has issues with the XSS filtering on input ... this shouldn't be done in the first place. XSS must only be applied to output.

PR #3475, opened 2015-01-13
var_export in DB.php outputting string

jdlferreira:
When a DSN string is supplied and it contains parameters such as "db_debug=TRUE", var_export will currently output the parsable string representation of the variable. By setting the second parameter to TRUE, the output is killed.

### DIFF
--- a/system/database/DB.php
+++ b/system/database/DB.php
@@ -130,7 +130,7 @@ function &DB($params = '', $query_builder_override = NULL)
 			{
 				if (is_string($val) && in_array(strtoupper($val), array('TRUE', 'FALSE', 'NULL')))
 				{
-					$val = var_export($val);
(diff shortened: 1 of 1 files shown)


system/database/DB.php
```diff
@@ -217,4 +217,4 @@ class CI_DB extends CI_DB_driver { }
 }
 
 /* End of file DB.php */
-/* Location: ./system/database/DB.php */
\ No newline at end of file
+/* Location: ./system/database/DB.php */
```
narfbg: Remove the empty line here please.

jdlferreira: Done.

narfbg: ... not, nothing is changed.

In case you're editing online through github's editor - you can't do that, you'll have to edit locally.

PR #3477, opened 2015-01-13
Add different Environmental variable for errors and config loading

dmyers2004:
Since we have updated the config loading to cascade being able to have the config file loading not tied directly to the way the errors are displayed comes in handy. You could set your local environment to "sandbox" and debug to "development" for example and then on the development server "development" for both.

This also safely falls back to it's present workflow.


narfbg:
You can do the same with more than 2 environments.

PR #3479, opened 2015-01-14
var_export in DB.php outputting string

jdlferreira:
When a DSN string is supplied and it contains parameters such as
"db_debug=TRUE", var_export will currently output the parsable string
representation of the variable. By setting the second parameter to TRUE,
the output is killed.


narfbg:
Another tip for your future contributions - you don't have to resubmit, just commit & push to the same branch.

PR #3488, opened 2015-01-20
Removed comments at EOF

Razican:
As suggested by #3476. This removes the comments at the end of the file and adds a new line at the end of each file.

Issue #3476, opened 2015-01-13
Put an EOL at end of files


vlakoff:
By removing the closing tags, https://github.com/bcit-ci/CodeIgniter/commit/a3ffbbb75ab9403941e4f810703313432b3993cc was actually an improvement.

Now, more than 6 years later, it's time to get rid of these blocks. Just remove these blocks and put an EOL. It's a no-brainer.

PRs have been so much polluted because of this.

<br>On a side note, the licence block at the top of the files can be cumbersome. I mainly think of the [default controller](https://github.com/bcit-ci/CodeIgniter/blob/develop/application/controllers/Welcome.php) and the [config files](https://github.com/bcit-ci/CodeIgniter/tree/develop/application/config). It would be great to remove them, if this is possible. @jim-parry

vince83110:
This should be a request not an issue no ?

narfbg:
Not against it, although it would look a bit like we've just trolled anybody who has ever submitted a PR.

kakysha:
@narfbg .. some day you need to stop trolling people with that, really. GitHub editor is unusable now and it makes this not so necessary rule more useless.

narfbg:
Did you not read my comment? I said I'm not against the proposal.

I don't appreciate you implying that I do troll people with it, I'm simply applying the same rule that was forced on me when I first started contributing.

Razican:
I have created a PR for addressing it, if you finally decide to do it. About the license, with the @license sentence is covered, without the need for all the comment block.

vlakoff:
I had prepared changes for this too, but locally... I'll get in touch a bit later.

narfbg:
@jim-parry Would you give a +1/-1 on this?

I'd make the call alone, but I suspect you care a bit more about this one.

jim-parry:
+1 Remove closing blocks

The closing block at the end of PHP files has always felt contrived to me.
I am in favor of removing it//.

jim-parry:
The comment block at the beginning of source code files feels awkward too, but it serves a useful purpose according to the guidance I have been able to find ... it is part of asserting "collective copyright", on behalf of BCIT, because BCIT would be the body to defend any copyright infringement action, and keeping it there kept our lawyers happy, and that was part of the long and convoluted struggle/process to get the CodeIgniter adoption blessed by BCIT's lawyers. What a mouthful!

[...]

PR #3490, opened 2015-01-20
Remove closing blocks at end of PHP files

vlakoff:
Here are my changes for addressing #3476. Should supersede #3488.

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -174,7 +174,3 @@ $autoload['language'] = array();
 */
 
 $autoload['model'] = array();
-
-
-/* End of file autoload.php */
-/* Location: ./application/config/autoload.php */
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -521,7 +521,3 @@ $config['rewrite_short_tags'] = FALSE;
 | Array:		array('10.0.1.200', '192.168.5.0/24')
 */
 $config['proxy_ips'] = '';
-
-
-/* End of file config.php */
-/* Location: ./application/config/config.php */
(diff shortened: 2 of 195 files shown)


narfbg:
I don't like commenting on commits ... the conversations get lost easily. So, continuing here from https://github.com/bcit-ci/CodeIgniter/commit/4606ac3c6b3c512f10820c6568c1080ec08ca182#commitcomment-9363377

I thought you proposed no comments at all at EOF?

Razican:
The question is about the docs. I think that with putting the code with no EOF comments would be better than putting the EOF comment only in the docs.

narfbg:
@Razican I don't really understand your last comment ...

Anyway, it looks ambigous in the docs and I don't personally see a way of making it otherwise, so I'd rather just remove that particular 'incorrect vs. correct' example and leave it just to wording.

Btw, I'd also change the wording a bit ...

Before:

    For this reason, all PHP files should **OMIT** the closing PHP tag. You should also put one empty line at the bottom of the file.

After:

    For this reason, all PHP files MUST OMIT the closing PHP tag and end with a single empty line instead.

I don't care much about the styling, which is why I didn't bold anything myself - your call on that.

Razican:
What I meant was that the EOF comment in the docs is only for clarifying purposes.

vlakoff:
Agree with your comments, PR updated.

ivantcholakov:
Is such end-of-file cleaning to be applied on codeigniter3-translations project?

narfbg:
You can raise that as an issue there, or better yet - submit a pull request. I don't see a reason not to, but it's a separate repository, which I'm not involved with.

PR #3493, opened 2015-01-21
Update Session_database_driver.php

desmowow:
The column named 'id' of ci_sessions table is wrong.
The correct name is 'session_id'


narfbg:
It's not wrong.

PR #3494, opened 2015-01-21
Fix Sphinx broken path

vlakoff:
A few additional notes:
* For robustness and simplicity, we may replace relative paths `<../libraries/loader>` with absolute paths `</libraries/loader>`.
* @narfbg I was planning to end compiling the documentation by myself and just use the online version. You should verify that your changes compile correctly.


narfbg:
I can't compile the docs for every commit ...

PR #3495, opened 2015-01-21
Proposal: new script_tag

josewilsoncc:
Added a new function to html helper, which can generate script tags


narfbg:
#2335 #3062 

Please use the search box first.

josewilsoncc:
Please, could you tell me where I am wrong?

This feature supports all the parameters of the tag and it is documented.

narfbg:
You're wrong in the assumption that we want the feature included ... I posted those links so that you follow them, not ignore them.

PR #3496, opened 2015-01-21
Change all :doc: links to absolute paths

vlakoff:
This should improve maintainability.

### DIFF
--- a/user_guide_src/source/changelog.rst
+++ b/user_guide_src/source/changelog.rst
@@ -51,18 +51,18 @@ Release Date: Not Released
    -  Path constants BASEPATH, APPPATH and VIEWPATH are now (internally) defined as absolute paths.
    -  Updated email validation methods to use ``filter_var()`` instead of PCRE.
    -  Changed environment defaults to report all errors in *development* and only fatal ones in *testing*, *production* but only display them in *development*.
-   -  Updated *ip_address* database field lengths from 16 to 45 for supporting IPv6 address on :doc:`Trackback Library <libraries/trackback>` and :doc:`Captcha Helper <helpers/captcha_helper>`.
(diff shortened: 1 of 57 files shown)


narfbg:
I don't think that altering the existing content solves anything ... it's a matter of regular error checking.

vlakoff:
It reduces work to do and possibility of errors in case the structure is changed.

narfbg:
How exactly does changing the existing paths do that?

Making a policy to use absolute paths might, but we don't have such and frankly, I don't care enough to implement one. It's documentation, gets updated once in a while and errors appear exactly then (when necessary), it's not production code.

vlakoff:
I've already made all the necessary replacements. Nevertheless, it would introduce a *de facto* standard. Users looking at other parts of the code when making changes. Yep, we could add a "contributing to the documentation" documentation, but don't care much too.

Whatever, just merge it or close it. I'm strongly +1 for the former.

vlakoff:
As you cherry-picked the first commit, can I safely assume you don't plan to merge the main commit?

narfbg:
Well, the first one is a *fix*, so it had to be applied anyway.

I've not yet decided whether to merge it or not, although you know I'm leaning towards the latter ...

vlakoff:
I'm fine if you decide not to merge it; after all, the changes are numerous, which has drawbacks.

Juuust a nitpick, could you please take https://github.com/vlakoff/CodeIgniter/commit/d166c546b5b0b2727f1db02159d64e1cff12d632?

narfbg:
Done.

vlakoff:
Thanks. I'm closing this PR, as the remaining changes most probably won't be taken.

PR #3497, opened 2015-01-22
repair autoload.php when environment exists

avenirer:
autoload.php didn't load properly if an enviroment directory exists (develpment directory).

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -1232,10 +1232,13 @@ class CI_Loader {
 	 */
 	protected function _ci_autoloader()
 	{
-		include(APPPATH.'config/autoload.php');
 		if (file_exists(APPPATH.'config/'.ENVIRONMENT.'/autoload.php'))
 		{
 			include(APPPATH.'config/'.ENVIRONMENT.'/autoload.php');
+		}
+		else
+		{
+			include(APPPATH.'config/autoload.php');
 		}
 
 		if ( ! isset($autoload))


narfbg:
You're breaking, not repairing it - it should override, not replace the main one.

avenirer:
but it gives error if no autoload.php is in config directory. Should we keep an autoload.php directly inside config directory, separated from development and production directories?

narfbg:
Of course you should ... that's why it's there by default.

avenirer:
ok. thank you.

avenirer:
But wouldn't that mean that that rule should also (for example) apply to config.php and routes.php (as in, copied, not moved)? For example, as I am working in development and not production, I've moved config.php and routes.php inside the development directory, and no error was reported. But when I've moved the autoload.php didn't work. For this one I have to copy and not move. Does autoload.php have a special status? An autoload.php being in config directory and also in a potential "production" and/or 'development' directory would introduce confusion, the only solution imposed by the framework being to use only one autoload.php (the one in config) throughout all the application.

narfbg:
Well, if that's the behavior for other configs - sure, we can patch it to only load config/autoload.php if it exists, but otherwise it's not about copying or moving ... it's about the order in which they are loaded, (and they are all always loaded).

ivantcholakov:
I think, code is fine as it is now. The logic in pseudo-code is something like this:

```php
$autoload = array_merge_recursive($autoload, $autoload_development);
```

It is fine.

@avenirer Don't "move" autoload.php file, under the directory config/development/ create a new file autoload.php that contains only the options you want to override in development environment.

avenirer:
Logically speaking, you can't have an autoload.php for development environment and another autoload.php for the production environment (if you develop something in development environment so that it will later be implemented in the production environment you assume the same libraries, helpers, packages will be autoloaded).

Hence, it is only logical to have only one autoload.php (the one inside config directory); but then the application shouldn't look for an autoload inside the environments directories.

On the other hand, if someone decided to have a special autoload.php for an environment (development) it is only logical for the framework to first look for that autoload.php inside that directory and only if it doesn't exist, to look for the autoload.php in the config directory.

But the decision is yours.

ivantcholakov:
Under config/ENVIRONMENT/ there is no full set of the configuration files. There are only some chosen files that contain only differences. This is logical too. An it is wise, you would not need to support (compare?) full (and modified) copies of the configuration files.

The status quo if fine.

narfbg:
You're both overthinking it.

avenirer:
Happy... happy... joy... joy :+1:

ivantcholakov:
Ok, this does not break anything.

Issue #3498, opened 2015-01-22
Remove license headers at top of files


vlakoff:
Follow-up to #3476.
Separate issue as we are now specifically talking about the (huge) license headers.

<br>I think that as long as we have these info in each file we are fine:
* What license? MIT
* To who? BCIT

The relevant licensing info, straightforward. Two lines.

The rest is unnecessary redundancy. These belong to the root directory (you know, the usual *README*, *LICENSE*, *AUTHORS*...):
* Full license text (thankfully it is short)
* Authors listing (EllisLab, BCIT)
* Project description (website, etc.)

In fact, doing this way would be better for the lawyers, because the "metadata" included in each file is the only relevant one about licensing, so that it focuses on it.

<br>A few additional comments:
* @narfbg Remember that `/application` was previously inside `/system` ;-)
* @webvitaly We already have a license file in the root directory.

Razican:
As I said in the other issue, my opinion is that having the reference to the license and the authors is enough, and that we do not need to copy the license in each file. If it was GPL, would you copy the whole license in each file? that would double the project size.

narfbg:
@vlakoff 

I don't understand why you insist on removing or shortening the comment blocks under system/ ... It has never bothered me and nobody spends nearly as much time as myself looking at those files.

Are you scrolling until you find what you want or ...? Really, I don't get it.

@Razican 

Thankfully, it's not GPL. If it was completely up to me, it would've been the ISC license, which is even shorter (but with literally the same meaning). Still, MIT is short enough not to care about that.

ivantcholakov:
I think, full license headers protect owner's rights better. The files may be used outside CodeIgniter framework, people may bundle specific libraries and drivers within different projects. Assuming this possibility it would be better these full headers to stay.

vlakoff:
@narfbg In fact I agree with you:

* Those in `/system` shouldn't annoy users, and so can be kept.
* Those in `/application` are annoying, and should be trimmed as I posted above, or removed (there isn't much, if anything, to "protect"). I'm for the latter.

narfbg:
@vlakoff 

I guess I misunderstood your reasoning to open another issue then ... it was just a matter of @jim-parry giving another +1 in the previous thread.

vlakoff:
Long story short: #3499. Feel free to suggest any improvement.

jim-parry:
The existing comment blocks in sorce code files under system/ and in index.php are appropriate.
Quoting OSI:
[...]

PR #3499, opened 2015-01-22
Remove license blocks in /application files

vlakoff:
Refs #3476 and #3498.


jim-parry:
See my comments for #3498.

Issue #3431, opened 2014-12-20
Session path determination in upcoming session library


vlakoff:
Opening a new issue to not clutter #3073.

Trying the feature/session branch some time ago, I encountered the same issue as in https://github.com/symfony/symfony/issues/4908. I remember having had spent a bit of time to figure out the problem.

I think we should either:
- Detect if invalid path, and if so provide the user with helpful message. The solutions are to change PHP session path, or better, specify a session path in CI config.
- Implement a mechanism like in https://github.com/symfony/symfony/pull/5016/files.

npulidom:
I came up with this issue in OSX 10.10.1 with Apache 2.4, PHP 5.6 & CI 3.0 dev. 
Folder `/var/tmp/` must have write perms (chmod -R 777)
In environment config, for OSX platform, make sure is set `session_save_path()` is set.

```php
//CI environment conf
if (defined('ENVIRONMENT'))
{
	switch (ENVIRONMENT)
	{
		case 'development':
			ini_set('display_errors', 1);
			error_reporting(E_ALL);
			//for OSX platform
            if(php_uname('s') == "Darwin")
            	session_save_path("/var/tmp/"); 
        break;
		case 'testing':
            //....
        break;
    }
}
```

vlakoff:
@npulidom Not really the best place to put the code…
@narfbg #3503 should be quite beneficial.

narfbg:
@ both of you ...

The only reason you encountered the error is because you didn't have a `$config['sess_save_path']`.

thiagopalooza:
I did have configured $config['sess_save_path'], but I still get the error...

narfbg:
Having it set to NULL does not mean it's configured ...

jouwstrm:
Please tell me you're not the author of codeigniter.

1. You're way too condescending... Grow Up.
2. sess_save_path should be setup already, or at least come with an example.

jouwstrm:
If you can't talk to users of codeigniter with a little respect, and realize there are bugs in the software still, please don't respond... AND PLEASE DON'T CONTRIBUTE.

sess_save_path should already be setup upon installation or atleast come with an example path.

Grow up.

narfbg:
If you think my comments were disrespectful, you should read what you just wrote yourself.

jouwstrm:
like I said... Grow up.

narfbg:
[...]

PR #3503, opened 2015-01-23
Better error reporting if problematic session path

vlakoff:
Addresses #3431. Benefits only.

* warning thrown as previously, but with more useful message
* logged too


narfbg:
https://github.com/bcit-ci/CodeIgniter/issues/3431#issuecomment-71276025
https://github.com/bcit-ci/CodeIgniter/issues/3431#issuecomment-67773640

+ we never use `trigger_error()` (except in compat/, where the goal is to emulate PHP as close as possible). Also, the intent is that a message should *not be shown* at all, but I just don't want to use error suppression to avoid that when it's clearly the user's fault.

PR #3506, opened 2015-01-24
Changes in file headers

vlakoff:
Addresses #3498, based on @jim-parry's comments.


narfbg:
EllisLab is still the author and Jim's comments were about application/ only ...

Please don't open any new PRs about this ... we now have ~10 issues and pull requests for this single thing. We'll do what's necessary once we decide it.

PR #3510, opened 2015-01-26
Session writeup error?

jim-parry:
This mistake was pointed out to me ... it looks like the session table create command suggests "session_id" as the primary key, when it should actually say "id".

### DIFF
--- a/user_guide_src/source/libraries/sessions.rst
+++ b/user_guide_src/source/libraries/sessions.rst
@@ -569,7 +569,7 @@ For PostgreSQL::
 		"ip_address" varchar(45) NOT NULL,
 		"timestamp" bigint DEFAULT 0 NOT NULL,
 		"data" text DEFAULT '' NOT NULL,
-		PRIMARY KEY ("session_id")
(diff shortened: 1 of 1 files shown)


Razican:
AFAIK "id" was renamed to "session_id" in CI3

jim-parry:
The problem is the contradictory SQL...
CREATE TABLE "ci_sessions" (
        "id" varchar(40) NOT NULL,
        "ip_address" varchar(45) NOT NULL,
        "timestamp" bigint DEFAULT 0 NOT NULL,
        "data" text DEFAULT '' NOT NULL,
        PRIMARY KEY ("session_id")
);

CREATE INDEX "ci_sessions_timestamp" ON "ci_sessions" ("timestamp");

There is no "session_id" field created. I don't have any strong feeling for one or the other, just that they should match :-/

narfbg:
You haven't pulled last week's changes ... 3053b81940f2b5e8f608290a43a77886ebc70ae4

narfbg:
... or you have, but my search for 'session_id' failed for some reason.

PR #3511, opened 2015-01-26
Update CI_VERSION from 2.2.0 to 2.2.1

ColonelMoutarde:


narfbg:
Did that manually ... don't target master.

PR #3518, opened 2015-01-27
On email validation check for MX records

george-dragnea:
We can use getmxrr to check if domain has MX records.


kakysha:
From http://php.net/manual/en/function.getmxrr.php:
`Note:
This function should not be used for the purposes of address verification. `

Issue #3520, opened 2015-01-27
[Proposal] Extending the Form Validation


foxlance:
I extended the validation class to support more functionality such as date validity, in and not in, etc. 

Would you look at 

https://github.com/foxlance/CodeIgniter/compare/bcit-ci:develop...develop 

and see if its worth creating a pull request?

Thanks!

narfbg:
At first glance - 'in' might be useful.

The rest can either be done though already existing rules or are prone to bugs and reports about them will surely follow.

I will have to ask you to go straight with a pull request in the future though. If you've already commited code in a fork, it's probably easier to submit it as a PR than to write a description about it in an issue.

PR #3521, opened 2015-01-28
Extending form validation methods

foxlance:
Added support for the following functions:

Between - numeric value should be within a given range
In / Not In - string should be in/not in a list of allowed values
Date - string should be a valid date
Before Date / After Date - date should fall between a given date range

### DIFF
--- a/system/language/english/form_validation_lang.php
+++ b/system/language/english/form_validation_lang.php
@@ -65,4 +65,4 @@ $lang['form_validation_less_than_equal_to']	= 'The {field} field must contain a 
 $lang['form_validation_greater_than']		= 'The {field} field must contain a number greater than {param}.';
 $lang['form_validation_greater_than_equal_to']	= 'The {field} field must contain a number greater than or equal to {param}.';
 $lang['form_validation_error_message_not_set']	= 'Unable to access an error message corresponding to your field name {field}.';
-$lang['form_validation_in']		= 'The {field} field must be one of: {param}.';
+$lang['form_validation_in_list']		= 'The {field} field must be one of: {param}.';
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -1455,7 +1455,7 @@ class CI_Form_validation {
 	 * @param	array
 	 * @return	bool
 	 */
-	public function in($value, $list)
+	public function in_list($value, $list)
 	{
 		return in_array($value, explode(',', $list), TRUE);
 	}
--- a/user_guide_src/source/changelog.rst
+++ b/user_guide_src/source/changelog.rst
@@ -350,10 +350,10 @@ Release Date: Not Released
       -  Removed method ``is_numeric()`` as it exists as a native PHP function and ``_execute()`` will find and use that (the **is_numeric** rule itself is deprecated since 1.6.1).
       -  Native PHP functions used as rules can now accept an additional parameter, other than the data itself.
       -  Updated method ``set_rules()`` to accept an array of rules as well as a string.
-      -  Added support for ``in`` method that test if a given value is within a given list.
       -  Fields that have empty rules set no longer run through validation (and therefore are not considered erroneous).
       -  Added rule **differs** to check if the value of a field differs from the value of another field.
       -  Added rule **valid_url**.
+      -  Added rule **in_list** to check if the value of a field is within a given list.
       -  Added support for named parameters in error messages.
       -  :doc:`Language <libraries/language>` line keys must now be prefixed with **form_validation_**.
       -  Added rule **alpha_numeric_spaces**.
--- a/user_guide_src/source/libraries/form_validation.rst
+++ b/user_guide_src/source/libraries/form_validation.rst
@@ -961,7 +961,7 @@ Rule                      Parameter  Description                                
                                      not numeric.
 **less_than_equal_to**    Yes        Returns FALSE if the form element is greater than the parameter value,                        less_than_equal_to[8]
                                      or not numeric.
-**in**                    Yes        Returns FALSE if the form element is not within a predetermined list.                         in[red,blue,green]
+**in_list**               Yes        Returns FALSE if the form element is not within a predetermined list.                         in_list[red,blue,green]
 **alpha**                 No         Returns FALSE if the form element contains anything other than alphabetical characters.
 **alpha_numeric**         No         Returns FALSE if the form element contains anything other than alpha-numeric characters.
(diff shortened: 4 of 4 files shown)


narfbg:
I believe I told you in #3520 that only the 'in' rule may be useful ... This PR also has conflicts with our current code.

foxlance:
I understand. I'm fine with only the 'in' rule making it. I'll close this request and create a new one instead? (or is there s better way of deleting those other commits from here?)

narfbg:
You don't have to submit a new pull request, but you do have to sync your code with changes in the origin repository ... otherwise it's not mergeable.

You'll also have to add a changelog entry and I'll leave some inline comments on the code.

system/libraries/Form_validation.php
```diff
@@ -1438,6 +1438,22 @@ public function less_than_equal_to($str, $max)
...
 	/**
+	 * Value should be within an array of values
+	 *
+	 * @param	string
+	 * @param	array
+	 * @return	bool
+	 */
+	public function in($value, $list)
+	{
+		$list = array_map('trim', explode(',', (string) $list));
```
narfbg: This is completely unnecessary, and both of the casts too ... the whole method can be reduced to:

    public function in($value, $list)
    {
        return in_array($value, explode(',', $list), TRUE);
    }

system/language/english/form_validation_lang.php
```diff
@@ -53,6 +53,7 @@
 $lang['form_validation_less_than_equal_to']	= 'The {field} field must contain a number less than or equal to {param}.';
 $lang['form_validation_greater_than']		= 'The {field} field must contain a number greater than {param}.';
 $lang['form_validation_greater_than_equal_to']	= 'The {field} field must contain a number greater than or equal to {param}.';
+$lang['form_validation_in']		= 'The {field} field must be within a predetermined list: {param}.';
```
narfbg: A more friendly message:

    The {field} field must be one of: {param}

system/libraries/Form_validation.php
foxlance: True, however, if the developer accidentally coded something like

```php
$this->form_validation->set_rules('color', 'Color', 'in[red, blue]');
```
the space will be taken into account, and the code will fail. I thought running trim would help a bit. What do you think?

system/language/english/form_validation_lang.php
foxlance: Got this.

system/libraries/Form_validation.php
narfbg: I think it *should* fail.

foxlance:
This should cover all the necessary changes.

user_guide_src/source/changelog.rst
```diff
@@ -350,6 +350,7 @@ Release Date: Not Released
       -  Removed method ``is_numeric()`` as it exists as a native PHP function and ``_execute()`` will find and use that (the **is_numeric** rule itself is deprecated since 1.6.1).
       -  Native PHP functions used as rules can now accept an additional parameter, other than the data itself.
       -  Updated method ``set_rules()`` to accept an array of rules as well as a string.
+      -  Added support for ``in`` method that test if a given value is within a given list.
```
narfbg: See 2 lines below how other new rules have been noted.

On another note however, might be better to change the rule's name - in_array or in_list?

ivantcholakov:
Perhaps @jim-parry should check the language line whether it sounds well in English.

narfbg:
@ivantcholakov 

If you don't think my English is good enough to handle that on my own, you should've said that years ago. Don't take this the wrong way, but doing it now makes me think that you're just looking for an excuse to say something, anything ... you don't have to leave a comment on every single issue.

ivantcholakov:
@narbg

I am sorry for making such an impression, yet, maybe it would be useful for you to have a look at this: https://github.com/bcit-ci/codeigniter3-translations/pull/134#issuecomment-70388488

Razican:
If you think some language strings could be improved, you can do a pull request. Personal issues do not fit in an open source project.

narfbg:
@ivantcholakov So ping me from the relevant repo/issue/whatever if necessary, don't hijack other threads ...

@Razican There's no personal issue. There's an off-topic issue and you just contributed to it.

I value both of you guys for your contributions and enthusiasm about the project, but you're too chatty at times. Let's keep it on-topic now, shall we?

@foxlance Sorry about that distraction.

Can you rename the rule to 'in_list' and alter the changelog message in a manner similar to the one for 'differs'?

foxlance:
@narfbg I believe this should suffice

system/libraries/Form_validation.php
```diff
@@ -1449,6 +1449,20 @@ public function less_than_equal_to($str, $max)
 	// --------------------------------------------------------------------
 
 	/**
+	 * Value should be within an array of values
+	 *
+	 * @param	string
+	 * @param	array
```
sergeyklay:
```php
* @param	string|array
```

system/libraries/Form_validation.php
```diff
@@ -1449,6 +1449,20 @@ public function less_than_equal_to($str, $max)
...
+	 * @return	bool
+	 */
+	public function in_list($value, $list)
+	{
+		return in_array($value, explode(',', $list), TRUE);
+	}
```
sergeyklay: Or:

```php
public function in_list($value, $list)
{
    if (is_string($list)) {
        $list = explode(',', $list);
    }
    return in_array($value, $list, TRUE);
}
```

sergeyklay:
```sh
$ php -r 'var_dump(in_array(0, ["a", "b", "c", "d"]));';
bool(true)

$ php --version
PHP 5.4.35-0+deb7u2 (cli) (built: Nov 19 2014 07:56:24) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2014 Zend Technologies

```

narfbg: 1st comment reply: FV rules are only strings.
2nd comment reply: ... which is why TRUE is passed as the third parameter, which you haven't done.

system/libraries/Form_validation.php
narfbg: Should be only string actually.

system/libraries/Form_validation.php
sergeyklay: :+1:

system/libraries/Form_validation.php
sergeyklay: but in doc array

PR #3525, opened 2015-01-28
Exprot error to db lang

b1tc0re:
Perhaps it was meant to be more comfortable but I think so.


narfbg:
These custom messages are (I assume) leftovers from EllisLab's work on their own products, they are hardly worth the effort or maintenance in CI however ... I'd rather remove than add more of them.

PR #3527, opened 2015-01-29
input.rst : clarifies cookie_prefix is not added

vtual:
The CI manual states that the get_cookie() function "is an alias to $this->input->cookie()." I believe this is an erroneous as $this->input->cookie() does not prepend the cookie prefix a user might have set in its config.php file, while get_cookie() does.
This is the source of many headaches... ;)
This change strives to clarify the difference in the way these two functions work for the benefit of all users.

### DIFF
--- a/user_guide_src/source/libraries/input.rst
+++ b/user_guide_src/source/libraries/input.rst
@@ -81,6 +81,8 @@ The main methods are:
 -  ``$this->input->cookie()``
 -  ``$this->input->server()``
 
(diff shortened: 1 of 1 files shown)


narfbg:
I had some nitpicky suggestions about the patch, so I committed the changes myself (same goes for #3528), I hope you don't mind.

For future reference however, you should submit all *related* changes at once instead of with separate pull requests. You can also update a PR with new changes once it is submitted.

vtual:
Thanks! Will take note of your remark.

PR #3531, opened 2015-01-29
fix typo

galdiolo:


narfbg:
It seems that you're looking all over the place for typos like that - thanks for the effort.

However, you might want to take your time and submit all of them at once when you're done with that - typos happen and the content is huge ... we might end up with tens of these pull requests if you submit them individually.

galdiolo:
I submitted this individual one on purpose to be easy for you to refuse the pull request if necessary, because I was not sure if the file changelog.rst was generated by a tool or manually edited.
I'm not searching for typos methodically (I'm just studying the code), but I'll try to group them as much as possible. Thanks for your huge effort on CodeIgniter.

narfbg:
We don't put auto-generated stuff in the repo ... that doesn't make sense.
Also: http://keepachangelog.com/ :)

PR #3553, opened 2015-02-04
put cookie_prefix into input class

fuxu:
`input->cookie` and `input->set_cookie` have a different implemtation while `cookie_prefix` was not an empty string.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -279,7 +279,9 @@ class CI_Input {
 	 */
 	public function cookie($index = NULL, $xss_clean = NULL)
 	{
-		return $this->_fetch_from_array($_COOKIE, $index, $xss_clean);
+		is_bool($xss_clean) OR $xss_clean = (config_item('global_xss_filtering') === TRUE);
+		$prefix = config_item('cookie_prefix');
+		return $this->_fetch_from_array($_COOKIE, $prefix . $index, $xss_clean);
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


narfbg:
This is intentional and documented.

fuxu:
but why it is designed to be different? I want to know the reason.
in my opinion, `cookie_prefix` should not be cared by engineers, just like `table_prefix`.

narfbg:
Cookies are user input, a database table prefix is not ...

fuxu:
so, it means, I don't need to care about it when I write a cookie, but I must remember there is a `prefix` when I read it. I don't think it is a good implementation~

narfbg:
Well, would it be a good implementation if you weren't able to read a non-prefixed cookie if you had to?

fuxu:
en, I see what you mean. but I think if the `set` method is implemented with prefix, non-prefixed cookies will exists~~ am i right?

narfbg:
No, not really. Another application on the same domain may set a non-prefixed cookie. Even your application can do that if you really want it to.

fuxu:
or, in another way. I think it should be implemented without prefix in `input->cookie` and `input->set_cookie`, and with prefix in `set_cookie` and `get_cookie`?

narfbg:
Well, in practice, that could've been the only choice if `CI_Input::set_cookie()` didn't exist. It shouldn't have existed in the first place ... it's a class dealing with *input*, and that method does *output*.

The problem is, it's been there for a long time and we're stuck with it for BC reasons alone, so it doesn't make much sense to change it's behavior - that would be a BC break in its own right.

Documentation improvements (if and where necessary) are our only choice at this point.

fuxu:
I see. thank you for your patience :smiley: .

PR #3557, opened 2015-02-04
enabled incremental bcc() calls\n

druu:
so far multiple calls to bcc() replaced the existing bcc-array.
this way, the blind recipients will be amended on each call

Signed-off-by: David Wosnitza <david@druul.in>

### DIFF
--- a/system/libraries/Email.php
+++ b/system/libraries/Email.php
@@ -671,11 +671,19 @@ class CI_Email {
 
 		if ($this->_get_protocol() === 'smtp' OR ($this->bcc_batch_mode && count($bcc) > $this->bcc_batch_size))
 		{
-			$this->_bcc_array = $bcc;
+			if (count($this->_bcc_array))
+			{
+				$this->_bcc_array = array_merge($this->_bcc_array, $bcc);
+			}
+			else
+			{
+				$this->_bcc_array = $bcc;
+			}
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


narfbg:
It would make more sense to make it accept an array IMO.

druu:
that, it already supports

narfbg:
So how is this an issue then?

druu:
not an issue, just an enhancement regarding expected behaviour.

narfbg:
OK, I'll rephrase then ... why should addition instead of replacement be the expected behavior? :)

druu:
AH now I got what you're after.
depending on the situation and conditions (e.g environments, debugging and whatnot)  it would be cleaner to see multiple calls to ->bcc()  instead of having to prepare an array before/throughout the process of setting everythign up.

as soon as I wrote this, I realized I forgot to equalize this behavior on ->cc()   which I shall add as soon as I conviced you ;)

narfbg:
That would be more verbose, but not necessarily cleaner - that's down to personal preference.

The more important thing is that this change would be a BC break ... even if you could convince me to do that, we're already in an RC stage, so ... no, can't happen.

druu:
ah, fair enough. didn't consider BC in this case being an issue at all.
maybe next time ;)

vlakoff:
Methods `add_cc()` and `add_bcc()` could be fine.

PR #3559, opened 2015-02-04
Session cookies driver.

kakysha:
The cookies driver saves session data in user's browser cookie. It is the old CI2 'Session Library' that was rewritten as a driver to meet new CI3 Session Library requirements.

The downside of the cookies driver is lack of locks (the only one non-blocking driver) and the presence of second cookie for data storage besides the first one that stores session ID. Integrity and security check of input cookies are done via HMAC signatures. This means that you *MUST* set ``$config['encryption_key']`` in order to make this driver work. Also driver supports optional data encryption. 

P.S. this PR is made only for those who still in love with the good old session library. I know that it never will be merged for the absence of locks, but it still be googlable and usable. Thanks.


narfbg:
Are you fucking kidding me?! My last commit is a section in the session docs explaining why not using locks in sessions is the worst decision ever.

You know this will never get merged, you know the documentation clearly states that this will never be again supported, you're aware we're trying very hard discourage users to use github as a place to get help and you're intentionally *exploiting* GitHub's page rank to promote this pile-of-shit thing.

I can't fucking believe this. You've just lost any respect I've ever had for you.

PR #3560, opened 2015-02-04
Use session cookie name in gc check

Nibbler999:
I noticed this check in the session gc code isn't particularly reliable. First, it ignores the cookie name which is the obvious thing to use to check if the file is a CI session file. Secondly it assumes the session id is a 40 character hex string which is not a valid assumption.

### DIFF
--- a/system/libraries/Session/drivers/Session_files_driver.php
+++ b/system/libraries/Session/drivers/Session_files_driver.php
@@ -107,7 +107,7 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 	 * Sanitizes the save_path directory.
 	 *
 	 * @param	string	$save_path	Path to session files' directory
-	 * @param	string	$name		Session cookie name, unused
+	 * @param	string	$name		Session cookie name
 	 * @return	bool
 	 */
 	public function open($save_path, $name)
@@ -332,10 +332,16 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 
 		$ts = time() - $maxlifetime;
 
+		$pattern = sprintf(
+			'/^%s[0-9a-f]{%d}$/',
+			preg_quote($this->_config['cookie_name'], '/'),
+			($this->_config['match_ip'] === TRUE ? 72 : 40)
+		);
+
 		foreach ($files as $file)
 		{
 			// If the filename doesn't match this pattern, it's either not a session file or is not ours
-			if ( ! preg_match('/(?:[0-9a-f]{32})?[0-9a-f]{40}$/i', $file)
+			if ( ! preg_match($pattern, $file)
 				OR ! is_file($this->_config['save_path'].DIRECTORY_SEPARATOR.$file)
 				OR ($mtime = filemtime($this->_config['save_path'].DIRECTORY_SEPARATOR.$file)) === FALSE
 				OR $mtime > $ts)


narfbg:
It is a [very valid assumption](https://github.com/Nibbler999/CodeIgniter/blob/session-fixes/system/libraries/Session/Session.php#L322).

I don't mind including the cookie name in there as well, but your solution currently isn't more reliable either ... If you take for example a CI session cookie named 'foo', your solution will also delete sessions for cookies named 'foobar'.

Nibbler999:
Ah, I didn't notice that. My apologies.

I have put back the regex check, but the foo/foobar problem remains.

narfbg:
You might as well eliminate the strpos() call and combine the two in a single regular expression. :) But yes ... the foo, foobar issue remains regardless of that.

There is a 100% sane way - combining all the factors (cookie name, match_ip dependency, session ID pattern) in one. Truth is, I intended to that in the first place but wasn't sure how match_ip would fit in back when I wrote this, so I just left the bits that won't break if anything changes.

Give it a try if you wish so, I'll commit it in a few days if you don't.

vlakoff:
To avoid repetition and improve clarity, it would be great to assign the following (3 occurrences) to a variable:

    $this->_config['save_path'].DIRECTORY_SEPARATOR.$file

narfbg:
Agreed on assigning to a variable, but about the regular expression. Should be something like this:

    $pattern = sprintf(
        '/^%s[0-9a-f]{%d}$/',
        preg_quote($this->_config['cookie_name'], '/'),
        ($this->_config['match_ip'] === TRUE ? 72 : 40)
    );

The `$this->_config['save_path'].DIRECTORY_SEPARATOR.$file` part on the other hand is constructed inside a loop, so I'm not sure about that ... I personally find it clear enough and not really that long. It's the regular expression that becomes barely readable at this point.

vlakoff:
* I don't really like this sprintf proposal :\
* "The (...) part on the other hand is constructed inside a loop"... are you talking about performances? 1) irrelevant. 2) saves property accesses so might be faster actually. 3) don't even care to benchmark that. 4) irrelevant, seriously.

Additional notes:
* We can simplify `'(?:[0-9a-f]{32})'` to `'[0-9a-f]{32}'`
* Then, I would rather suggest:

        $pattern = '/^'.$cookie_name
            .($this->_config['match_ip'] ? '[0-9a-f]{32}' : '')
            .'[0-9a-f]{40}$/i';

narfbg:
I don't care that much about performance, but it always rings a bell when I'd have to change code inside a loop for readability alone ... the more important part was that it's currently readable enough IMO. Either way, I don't care that much about it.

And speaking of readability, I'm not a `sprintf()` fan either and my initial thought was what you're now suggesting (except you're missing a `preg_quote()` call), but you have to admit ... it's just ugly.

vlakoff:
Looks like we have different perceptions. But whatever, it's just about coding style. I leave the rest up to you to decide, you have all the elements.

Nibbler999:
This is the cleanest way I can see. Using the = character as separator because it's not valid for the cookie name but still valid as a file name.

narfbg:
It still needs a regular expression ... why do you not want it to be just a single regexp?

Nibbler999:
I don't understand why the regular expression is required.

narfbg:
Because that's the only way to match a random hex-encoded string that is the session ID ... :)

Nibbler999:
Why do you want to match the random part? Checking the prefix tells us it's a CI session file and that's all we need to know.

narfbg:
It's less of a chance to destroy another app's valid session, but still incomplete logic. If somebody reports a bug that the GC will erase a file that they named 'foo=bar.txt' in that dir, it will be a valid bug report.

Nibbler999:
Changed to use the sprintf suggestion.

narfbg:
Great, thanks.

PR #3562, opened 2015-02-05
Allow not-escaping the value in set_value()

avenirer:
Wouldn't it be a good idea to allow for not escaped html in set_value()?

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -676,17 +676,18 @@ if ( ! function_exists('set_value'))
 	 *
 	 * @param	string	$field		Field name
 	 * @param	string	$default	Default value
+	 * @param	bool	$escaped	Escaped value
 	 * @return	string
 	 */
-	function set_value($field, $default = '')
+	function set_value($field, $default = '',$escaped = TRUE)
 	{
 		$CI =& get_instance();
 
 		$value = (isset($CI->form_validation) && is_object($CI->form_validation) && $CI->form_validation->has_rule($field))
 			? $CI->form_validation->set_value($field, $default)
 			: $CI->input->post($field, FALSE);
-
-		return html_escape($value === NULL ? $default : $value);
+		$value = ($value === NULL) ? $default : $value;
+		return $escaped ? html_escape($value) : $value;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


system/helpers/form_helper.php
```diff
@@ -676,17 +676,18 @@ function form_prep($str)
 	 *
 	 * @param	string	$field		Field name
 	 * @param	string	$default	Default value
+	 * @param	bool	$escaped	Escaped value
 	 * @return	string
 	 */
-	function set_value($field, $default = '')
+	function set_value($field, $default = '',$escaped = TRUE)
```
narfbg: Needs a space after the comma, and should be named `$escape` (no 'd').

system/helpers/form_helper.php
```diff
@@ -676,17 +676,18 @@ function form_prep($str)
...
 	{
 		$CI =& get_instance();
 
 		$value = (isset($CI->form_validation) && is_object($CI->form_validation) && $CI->form_validation->has_rule($field))
 			? $CI->form_validation->set_value($field, $default)
 			: $CI->input->post($field, FALSE);
-
-		return html_escape($value === NULL ? $default : $value);
+		$value = ($value === NULL) ? $default : $value;
```
narfbg: Might as well change that to:

    isset($value) OR $value = $default;

Also, please don't remove the empty line above it.

system/helpers/form_helper.php
```diff
@@ -676,17 +676,18 @@ function form_prep($str)
...
+		return $escaped ? html_escape($value) : $value;
```
narfbg: When using a ternary with a variable, please wrap it in parenthesis - makes it more readable.

narfbg:
Also needs a changelog entry and a documentation update.

system/helpers/form_helper.php
```diff
@@ -676,17 +676,19 @@ function form_prep($str)
 	 *
 	 * @param	string	$field		Field name
 	 * @param	string	$default	Default value
+	 * @param	bool	$html_escape	HTML escaped value
```
narfbg: "Whether to escape HTML special characters" or something like that ... "HTML escaped value" means a completely different thing.

avenirer:
done...

narfbg:
Great, just one change on the docblock remains ... :)

avenirer:
@narfbg I really don't envy you. You must have nerves of steel to be so patient... :+1:

narfbg:
Thanks. :)

PR #3563, opened 2015-02-05
session regenerate correction

avenirer:
The session didn't regenerate on the if condition.


narfbg:
10 minutes prior to you submitting this: 395f92882afada9701ab2384772ce331ae9b3f9d

... and if you look at your code, it doesn't match the comment description because it makes the first part of the check irrelevant.

PR #3565, opened 2015-02-05
update documentation for set_value()

avenirer:
Updated documentation according to pull request #3562

### DIFF
--- a/user_guide_src/source/helpers/form_helper.rst
+++ b/user_guide_src/source/helpers/form_helper.rst
@@ -571,28 +571,25 @@ The following functions are available:
 		// Would produce:  </form> </div></div>
 
 
-.. php:function:: set_value($field[, $default = ''])
+.. php:function:: set_value($field[, $default = '', $html_escape = FALSE])
 
 	:param	string	$field: Field name
 	:param	string	$default: Default value
+	:param	bool	$html_escape: HTML escape value
 	:returns:	Field value
 	:rtype:	string
 
 	Permits you to set the value of an input form or textarea. You must
 	supply the field name via the first parameter of the function. The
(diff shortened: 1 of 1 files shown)


user_guide_src/source/helpers/form_helper.rst
```diff
@@ -571,28 +571,25 @@ The following functions are available:
 		// Would produce:  </form> </div></div>
 
 
-.. php:function:: set_value($field[, $default = ''])
+.. php:function:: set_value($field[, $default = '', $html_escape = FALSE])
```
narfbg: Default is TRUE, not FALSE.

And optional arguments need to be wrapped in square brackets:

    [, $html_escape = TRUE]

user_guide_src/source/helpers/form_helper.rst
```diff
@@ -571,28 +571,25 @@ The following functions are available:
...
 	:param	string	$default: Default value
+	:param	bool	$html_escape: HTML escape value
 	:returns:	Field value
 	:rtype:	string
 
 	Permits you to set the value of an input form or textarea. You must
 	supply the field name via the first parameter of the function. The
 	second (optional) parameter allows you to set a default value for the
-	form.
+	form. The third (optional) parameter allows you to choose if you want the html_escape() to be applied to the value.
```
narfbg: The third (optional) parameter allows you to turn off HTML escaping of the value, in case you need to use this function in combination with i.e. :php:func:`form_input()` and avoid double-escaping.

However, please keep it up to 76 characters per line - just use the previous lines as an example for where to split.

narfbg:
And as with your changelog PR ... just add this to #3562.

PR #3572, opened 2015-02-05
Fix preg_replace issue in remove_evil_attributes

CronLogic:
If there are too many evil attributes passed in a POST array variable, the original preg_replace('') expression will fail because the resulting pattern becomes far too long.  This patch uses preg_replace_callback instead and hopefully should be PHP 5.2 compatible (if needed.)

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -774,44 +774,32 @@ class CI_Security {
 	{
 		$evil_attributes = array('on\w*', 'style', 'xmlns', 'formaction', 'form', 'xlink:href');
 
-		if ($is_image === TRUE)
-		{
-			/*
-			 * Adobe Photoshop puts XML metadata into JFIF images,
-			 * including namespacing, so we have to allow this for images.
-			 */
-			unset($evil_attributes[array_search('xmlns', $evil_attributes)]);
-		}
+	        if ($is_image === TRUE)
+	        {
+	            /*
+	             * Adobe Photoshop puts XML metadata into JFIF images,
+	             * including namespacing, so we have to allow this for images.
+	             */
+	            unset($evil_attributes[array_search('xmlns', $evil_attributes)]);
+	        }
 
-		do {
-			$count = 0;
-			$attribs = array();
-
-			// find occurrences of illegal attribute strings with quotes (042 and 047 are octal quotes)
-			preg_match_all('/(?<!\w)('.implode('|', $evil_attributes).')\s*=\s*(\042|\047)([^\\2]*?)(\\2)/is', $str, $matches, PREG_SET_ORDER);
-
-			foreach ($matches as $attr)
-			{
-				$attribs[] = preg_quote($attr[0], '/');
-			}
-
-			// find occurrences of illegal attribute strings without quotes
-			preg_match_all('/(?<!\w)('.implode('|', $evil_attributes).')\s*=\s*([^\s>]*)/is', $str, $matches, PREG_SET_ORDER);
-
-			foreach ($matches as $attr)
-			{
-				$attribs[] = preg_quote($attr[0], '/');
-			}
-
-			// replace illegal attribute strings that are inside an html tag
-			if (count($attribs) > 0)
-			{
-				$str = preg_replace('/(<?)(\/?[^><]+?)([^A-Za-z<>\-])(.*?)('.implode('|', $attribs).')(.*?)([\s><]?)([><]*)/i', '$1$2 $4$6$7$8', $str, -1, $count);
-			}
-		}
-		while ($count);
-
-		return $str;
+	        do {
+	            $GLOBALS['_remove_evil_attributes_count'] = 0;
+	
+	            //php 5.2 compatible
+	            $replace_func = create_function ('$match', '$GLOBALS["_remove_evil_attributes_count"]++; return "";');
+	
+	            // find occurrences of illegal attribute strings with quotes (042 and 047 are octal quotes)
+	            $str = preg_replace_callback('/('.implode('|', $evil_attributes).')\s*=\s*(\042|\047)([^\\2]*?)(\\2)/is', $replace_func, $str);
+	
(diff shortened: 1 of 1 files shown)


narfbg:
Hopefully ... what a horrible moment for Travis to drop 5.2, [literally during our last build](https://travis-ci.org/bcit-ci/CodeIgniter/builds/49658590).

Anyway ... `create_function()` and `$GLOBALS` are not really pleasant to review. Would you rather explain the issue in detail?

CronLogic:
Sorry about that, would prefer to use 'function closures' and 'use' but didn't think this was supported in 5.2.  I am not seeing what else would cause a problem w/5.2. In any case, the goal was to keep the safety counter variable in there.  The issue is that if the preg_replace string/pattern becomes too long, a PREG_BACKTRACK_LIMIT_ERROR is thrown.

narfbg:
Yea, 5.2 doesn't have closures and the use keyword.

How long exactly? pcre.backtrack_limit is configurable, we can simply change that.

narfbg:
Uh oh ... should've thought twice before hitting the button - configurable answers my "how long question". :)

Do you have it set to something lower than the default value? How did you encounter this?

CronLogic:
We haven't really touched the backtrack_limit.  We ran across this in the context of posting in a CMS scenario, where we may have a large chunk of HTML to send through.  Given enough attributes it would just error out and return a null string.

narfbg:
OK, can you test if adding another zero to your current value solves it for you? According to the docs, the default is 100k on versions prior to 5.3.7 and 1m after that.

CronLogic:
Sorry about the delay in getting back to this.  I've submitted another potential commit here that doesn't use GLOBAL and create_function.  I now see the trick to use array($this, 'function_name') that you've used elsewhere in this file.  I have tested the php.ini backtrack_limit...honestly we never actually specifically defined it so we have been using default.  I would maintain that using preg_replace_callback in this case would be more efficient for longer chunks of text and the php.ini would not have to be dealt with.  It was not immediately apparent why we were getting just blank POST variables so this patch could also avoid that confusion altogether for the developer.

narfbg:
I wasn't suggesting that you should just change your php.ini settings. Instead, I was thinking an `ini_set('pcre.backtrack_limit', <some high value>)` call in the class constructor would be good enough.

If you think your solution here is better, that's ok, but you'll have to explain your changes reeeally detailed if you want this to get merged.

You're replacing a very complex `preg_replace()` rule and the only thing you're doing to compensate that is adding a negative look-behind for word characters ... I don't think that amounts to the same effect.
I'd nitpick on other stuff as well, but that is the most important thing ... we're in a RC stage right now and this piece of code is very sensitive and carefully crafted.

CronLogic:
Hey Narf, thanks for the feedback.  Looking at the original code, it seems as if it is doing two lookups for all the attributes (quotes / no quotes).  The preg_replace looks like it is finding start and end HTML tags at the beginning and end of a string, and then restoring those tags without the evil attribute.  However, the pattern is such that pretty much all of these tags are optional.  If I have a string that is purely attributes (not inside an HTML tag), these will still get stripped out.  If the intention was to only look for attributes inside HTML tags then I could see how this would be beneficial, although it could still probably be done with preg_replace_callback.  With preg_replace_callback we are just replacing attributes with empty strings as we find them.  The original patterns to find the attributes are not changed.

It also occurred to me that the patterns could be combined here to do one loop.  This new code seems to be more efficient (and could be even more if patterns are combined) and potentially less problematic than doing a mass preg_replace.  

Basically this came about because we were unaware of what was going on at first -- the string would just be set to empty in the POST array and was concerning.  We spent a fair amount of time tracking it down, so my hope is that if it does a sufficient job then it might save others some time.  The original code isn't 'broken' and I do understand that you are in an RC stage ... and I also maybe missing something stupidly obvious so feedback is welcome :)  Just wanted to bring a possible alternative to the table.

andrefrcastro:
Not sure if it is related, but found this on the logs:

>ERROR - 2015-02-13 13:27:18 --> Severity: Warning  --> preg_replace(): Compilation failed: regular expression is too large at offset 76051 /system/core/Security.php 681

>ERROR - 2015-02-13 13:27:49 --> Severity: Warning  --> preg_replace(): Compilation failed: regular expression is too large at offset 76051 /system/core/Security.php 681

CI 2.2.1

narfbg:
@CronLogic 

Yeah, I was looking for more of a technical explanation on the regexp instead of rationale ... :)

The proposal does make sense, but there are just too many little, but also important details. Otherwise it would've been merged already (after some trivial changes). I've got a lot on my hands right now, but I'll try to review it more carefully in the next few days.

@andrefrcastro 

Yes, it is related - that's the problem this PR attempts to fix.

narfbg:
OK, I spent some time playing with your patch, comparing it to the old code and analyzing both patterns ...

You're correct in saying that the pattern tries to look for tags, but at the same time makes that condition optional ... that's a flaw and it does result in raw text being treated as HTML attibutes and subsequently modified.

Your code is indeed more efficient, but also over-intrusive in that it includes even less conditions. For example, it would remove something like this (while the old patterns wouldn't):

    jsonp = fancy

I wouldn't have noticed that myself if I didn't recall trying to avoid it previously (b69103e8ab0c646d01f5e97ef6a255293de1e60e, dbd999f33374f6541f167e3d77a3e80a991b301a).

You also seem to have mistaken `preg_replace_callback()` in general as a more efficient solution. It's not ... extra function (callback) calls alone make it slower. The issue comes from the fact that all input attributes are collected and stored in memory first instead of replacing them as they are found - this is something that you got right. However, the same effect is achievable via `preg_replace()` much easier and results in better performance.

See the commit that I've referenced above, which also makes tag matching non-optional, so attribute-like strings outside of HTML tags won't be modified. :)

Thanks for reporting the issue and for being so patient about it.

PR #3574, opened 2015-02-06
If we have is_unique() why not also have exists()?

avenirer:
This rule will verify if a value already exists in a specified database table field.


narfbg:
You can do the same with the new 'in_list' rule after pre-fetching the valid values.

roboticsexpert:
performance of count query is very better than select ...

gastongr:
> You can do the same with the new 'in_list' rule after pre-fetching the valid values.

However having a validation rule like this would be cleaner. It also helps with code reuse.
Please re consider it.

narfbg:
I'll probably sound like an asshole, but I guess after 5 identical PRs I do have to say this ... Here's why it won't be reconsidered:

- It is not FV's job to know that a database even exists, let alone be directly depend on it.
- The only argument for `exists` is "we already have `is_unique`".
- Yes, the suggestion to use `in_list` was just a brush-off excuse.
- Because of the first bullet point, the only related thing I'd consider is deprecating and subsequently removing `is_unique`.

Sorry.

PR #3575, opened 2015-02-06
correction on delete() return value

avenirer:
If no row was deleted the delete() method returned TRUE. This way we only return TRUE if at least a row was deleted.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -2117,7 +2117,15 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			$this->_reset_write();
 		}
 
-		return ($this->return_delete_sql === TRUE) ? $sql : $this->query($sql);
+		if($this->return_delete_sql === TRUE)
+		{
+			return $sql;
+		}
+		else
+		{
+			$this->query($sql);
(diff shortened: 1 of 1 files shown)


narfbg:
In SQL, that doesn't make sense. TRUE is a success, FALSE is a failure ... not finding a row is not a failure.

avenirer:
It is a failure if you want to delete something that is not there. If there is no row deleted when you asked to delete a row, then that it is a failure.

narfbg:
No, it's not. If it was, the query itself would trigger an error.

avenirer:
if in the model you try:

if($this->db->delete('table', array('id'=>$nonexistent_id))
{
echo 'ok';
}
else
{
echo 'false';
}

...it will echo 'true' even if that row doesn't exist and, hence, is not deleted.

kakysha:
Builder methods return value is to indicate the success of query execution. FALSE means that something is wrong on the way to db engine as it can not execute your query. 0 affected rows means SUCCESS of query execution, that's is why it return TRUE. Affected_rows() designed specially to use in your models, use it manually.

avenirer:
@kakysha you may be right... but is not reasonable to say that is a success somehing that is not done. Thank you for explanation.

PR #3585, opened 2015-02-09
Language loading 

jim-parry:
Include application/language in the loading of a language file, for consistency with the other application components (config, libraries, etc).

At the moment, language files are loaded from system/language, and then over-ridden in a lang->load call or else in one of the package paths. This change would load them from system/language, over-ride them from application/language, and then finally from a path specified in the lang->load call or else in one of the package paths.

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -123,6 +123,14 @@ class CI_Lang {
 		{
 			include($basepath);
 		}
+		
+		// Override language settings with any found in application/language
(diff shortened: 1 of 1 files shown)


narfbg:
I'm quite sure APPPATH is included in the package paths array ...

jim-parry:
Looking into Loader, it looks like the package paths are only set when add_package_path is called. Will test.

jim-parry:
I don't see where it is in the code, but it does indeed load a language file from APPPATH. My bad!

PR #3588, opened 2015-02-11
Update input.rst

BillClinton:
Added example for returning the entire input stream.


narfbg:
That particular piece of documentation is meant only to demonstrate that the method works the same as ``post()`` and ``get()``. Other than that, you shouldn't really need to use xss_clean with it and your example indirectly encourages that.

If you want to add an example about fetching the entire (decoded) stream, you can add to the method description in the class reference, but please omit the second parameter.

PR #3598, opened 2015-02-16
Added Missing Permitted Uri chars

riwakawebsitedesigns:
I added more permitted_uri_chars because some times developers need the & and =.


narfbg:
They can also sometimes add it ... Everybody needs to allow something *sometimes*, we might as well just remove the filter if that was our criteria.

PR #3601, opened 2015-02-18
Fix mysqli_real_escape_string

andreas22:
When string is empty or connection is not available do not run function mysqli_real_escape_string


narfbg:
So, if there's no connection - don't escape at all? Sorry, but no ...

sarciszewski:
The safest way to secure a database from SQL injection is to disconnect it from every network.

Issue #3255, opened 2014-09-26
User Authentication


code2be:
Hello,

Why not having a User Authentication library which handles the process of Registering, Logging in, Resetting Password of Users ?

I know there are many libraries for Codeigniter which does the same functions and more like ionAuth. But, Being a native function bundled with the framework is better as far as I think.

What do You think ?

Issue #1557, opened 2012-07-02
cart format_number returns empty string for 0


obozdag:
Cart library format_number function returns empty string for given 0. I want to show 0.00 for promotion in customer's cart. But it returns empty string.

PR #3609, opened 2015-02-19
Delete space

robinsere:


jim-parry:
Seriously?

obozdag:
Being polite or formal (when necessary)  is better attitude especially for instructors. You could point out the idea behind the change of starting "if ( ! defined..." statements in a new line or give a link to the conversation like via #xxxx.

jim-parry:
Ok, sorry.
Eliminating a single line as proposed does not solve any bug or issue.
Using most IDEs, reformatting the text will introduce the blank line, as a formatting rule, undoing any such change.
Eliminating a single line in a single file, and not in all of the other PHP files in application/ or in the project, especially given the above, seems like either a mistake or a silly (non-serious) proposal.
Not providing a description for the proposed change, or referencing an issue that it fixes, and having this pull request appear "out of the blue" as a result, leads me to conclude that the pull request was either a mistake or a non-serious request, which we have had our share of recently.
I could have just closed it, without any comment, but wanted to make a point.
I guess I made the wrong point, in your eyes.

narfbg:
Oh, the etiquette police caught us ...

@obozdag 

You're on the internet, where nobody is paid to be polite to others, and you're being waaay too sensitive. Especially for somebody with the nerve to say [this](https://github.com/bcit-ci/CodeIgniter/issues/3255#issuecomment-56993475), and to ask for [personal treatment](https://github.com/bcit-ci/CodeIgniter/pull/3213#issuecomment-54031929) from a non-profit project, but [can't even be bothered](https://github.com/bcit-ci/CodeIgniter/issues/1557) to check if their issue hasn't already been resolved.

I'm also quite certain that you have no idea what it is to manage a project like CI, especially in your spare time, so let me tell you - it's not humanly possible to always remain calm, polite and be as descriptive as you suggest.

Sorry, but you're in no position to give attitude advices to other people. Please stop doing it.

PR #3613, opened 2015-02-22
When using post_controller_constructor hook - Undefined variable: _SESSI...

alexvandervegt:
...ON

<h4>A PHP Error was encountered</h4>

<p>Severity: Notice</p>
<p>Message:  Undefined variable: _SESSION</p>
<p>Filename: Session/Session.php</p>
<p>Line Number: 730</p>

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -724,13 +724,19 @@ class CI_Session {
 		{
 			foreach ($key as $k)
 			{
-				unset($_SESSION[$k]);
+				if(isset($_SESSION[$k]))
+				{
+					unset($_SESSION[$k]);
+				}
 			}
 
 			return;
 		}
 
-		unset($_SESSION[$key]);
+		if(isset($_SESSION[$key]))
+		{
+			unset($_SESSION[$key]);
+		}
 	}
 
 	// ------------------------------------------------------------------------


narfbg:
The error means that you've already unset these variables, but you didn't unmark them as temp/flash data. So why didn't you do that?

alexvandervegt:
Please explain?

Also i think its always wise to check if a variable exists before unsetting it.

narfbg:
The code that you're changing simply unsets variables that were previously marked as 'tempdata' or 'flashdata'. These markers wouldn't exist if you didn't create them via `set_flashdata()`, `set_tempdata()` or `mark_as_temp()`, `mark_as_flash()`.
Therefore, if the `unset()` call is reached for them *and* the actual variables don't exist in `$_SESSION`, that means that you've marked them as flash/temp data, but you've manually unset them without `unset_flashdata()`, `unset_tempdata()` or `unmark_flashdata()`, `unmark_tempdata()`, like this:

    $this->session->set_flashdata('foo', 'bar');
    unset($_SESSION['foo']);
    // next request will try to unset($_SESSION['foo']) again

So, wise to check for that - sure, but it currently shows a logical error in your code, so in fact the error message is helping you.

alexvandervegt:
This is not done in the code. I just updated my system from the V3 Beta to V3 RC2 before i was using the buildin native session driver, this is not available anymore so i switched to the files driver.

And now this error appears, also i get a lot empty sessions when using the files system. So i switched back to the beta version again.

Hope you guys can do something with this feedback and update the session library to a stable one before a releasing a stable version.

narfbg:
It is quite stable, but if you've depended on some specific behavior of the "native" driver, it may not exist with the new library.

One thing I can tell you for sure is that you certainly *are* manually deleting data that's been previously set as flashdata/userdata ... even if not with the exact same code I showed. It's simply impossible to get these error messages otherwise.

alexvandervegt:
Its working now indeed, thanks for your replies :)

Issue #3558, opened 2015-02-04
dead parameter in form_upload


gadelat:
There is parameter $value in form_upload which is not used and does not make sense for file input tag. This parameter should be removed. Hovewer, this may cause breakage in user applications, so I think this should be changed in 3.1

narfbg:
Agreed, it's a stalemate right now.

narfbg:
There's nothing that can be done about this without breaking BC and that shouldn't happen in minor versions such as 3.1, so it stays as is.

narfbg:
On another note however, SemVer won't be benefitial as we move forward to CI4 ... it's going to be a rewrite anyway, so holding deprecated functionalities and such until then is pointless.

PR #3617, opened 2015-02-22
Issue #3558

pedrohills:
I just fix the problem.


narfbg:
You're only creating a BC break, not fixing anything.

PR #3620, opened 2015-02-23
Cache_memcached: Improve

Fieah:
Reduce unnecessary checks.

_setup_memcached($is_memcached):
1. Once extension loaded, the class will be automatically enable, so it is
unnecessary to check class_exists again.
2. Added 'is_memcached' into $_memcache_conf for further usage.

save():
Use 'is_memcached' instead of get_class


narfbg:
The name `$is_memcached` is ambiguous, especially for a config option - I have a bad feeling about such a setting.

I'm not convinced this is the right approach, if a change is necessary at all ... setting a property from inside `is_supported()` would be more simplistic and efficient.

Fieah:
This changes has performance difference, but i totally agree with your points.
Will work on it later.

PR #3623, opened 2015-02-23
DB reconnect(): improvements

madwings:
DB reconnect() function is useless in current form. "mysqlnd" driver disabled reconnection through ping method, long ago, even if you turn it on in settings - https://bugs.php.net/bug.php?id=52561. I do not now what is the status in cubrid and postgre, but ping was never a reliable way for testing or reestablishing connection. In PDO it is removed altogether and for a good reason. With the change it will work for all the drivers. In the future DB driver could be modified to automatically detect lost connection and reconnect.

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -475,16 +475,15 @@ abstract class CI_DB_driver {
 	/**
 	 * Reconnect
 	 *
-	 * Keep / reestablish the db connection if no queries have been
+	 * Reestablish the db connection if no queries have been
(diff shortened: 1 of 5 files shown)


narfbg:
This is a BC break for anybody who uses `reconnect()`, even if it doesn't work with mysqlnd. I'm also not keen on removing it for other drivers where you too are saying that you don't know how it works.

paranic:
great solution
i had the same problem in the past
don't you think we need a reliable way of ping in order to use reconnect() ?

PR #3628, opened 2015-02-26
Update  docblock

wphp:
Add active_record auto-complete support when using IDE


narfbg:
No, only bug fixes for the 2.2-stable branch please.

PR #3621, opened 2015-02-23
Common: set_status_header: Remove tabs in $stati

Fieah:
Weird. 
It only visible on Github.

PR #3630, opened 2015-02-27
common.php: log_message: Improvements

Fieah:
1. log_message: Check if logging enabled before call to CI_Log, i think
there's a difference.

2. Remove unnecessary property $_threshold_max

3. Grammar fix

4. Merged changes from #3621 

:)


narfbg:
It's the Log class' job to determine the threshold ... I know you're aiming for a performance improvement, but it shouldn't come in the form of duplications.

Fieah:
Well... OK.

PR #3636, opened 2015-03-02
Update CodeIgniter.php

piece601:
I think defult vendor directory in project root directory.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -166,8 +166,8 @@ if ( ! is_php('5.4'))
 	{
 		if ($composer_autoload === TRUE)
 		{
-			file_exists(APPPATH.'vendor/autoload.php')
-				? require_once(APPPATH.'vendor/autoload.php')
+			file_exists(APPPATH.'../vendor/autoload.php')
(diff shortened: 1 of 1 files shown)


narfbg:
... if that is your project root. You like it that way, others don't.

piece601:
but default composer.json in project root directory.

narfbg:
Yes, but that's for CI's own development, and at this time - only for running the test suite.

Otherwise, you can copy the application/ dir for multiple projects and manage different dependencies for them via composer, while still using the same system/ dir.

davidgv88:
The solution:

In application/config/config.php
$config['composer_autoload'] = FCPATH.'vendor/autoload.php';

piece601:
@davidgv88 Thanks.

It's good idea~

davidgv88:
Yesterday I wrote the same:
http://forum.codeigniter.com/thread-1429.html

narfbg:
Please keep it on the forums ... GitHub is NOT the place to ask or search for tips.

arikardnoir:
Im installing composer in Codeigniter and the files are going to dir "libraries", it's not creating the dir "vendor" in my app. When Im creating manual, and then make update of composer, he's creating files again in libraries.

PR #3638, opened 2015-03-02
travis: bash cleanup

TomasVotruba:


narfbg:
... look at the build for this PR.

TomasVotruba:
What does closing mean?

narfbg:
Why do you want to change it at all?

TomasVotruba:
Better readability.

narfbg:
I don't see much point in that for 4 lines in a bash script that's only used for tests. It's not like it's unreadable and barely anyone needs to look at it anyway.

TomasVotruba:
I see

PR #3640, opened 2015-03-03
Parser library variables escape & template includes

vov1:
Added functionality for escaping data and including template in to template.

{var} - prints escaped data
{{var}} - prints unescaped data

@include(template_name) - includes one view to another

### DIFF
--- a/system/libraries/Parser.php
+++ b/system/libraries/Parser.php
@@ -139,6 +139,8 @@ class CI_Parser {
 		{
 			return FALSE;
 		}
+        
+        $template = $this->_parse_includes($template, $data);
 
 		$replace = array();
 		foreach ($data as $key => $val)
@@ -189,7 +191,10 @@ class CI_Parser {
 	 */
 	protected function _parse_single($key, $val, $string)
 	{
-		return array($this->l_delim.$key.$this->r_delim => (string) $val);
+		return array(
+            $this->l_delim.$this->l_delim.$key.$this->r_delim.$this->r_delim => (string) $val,
+            $this->l_delim.$key.$this->r_delim => (string) htmlspecialchars( $val )
+        );
 	}
 
 	// --------------------------------------------------------------------
@@ -208,7 +213,7 @@ class CI_Parser {
 	{
 		$replace = array();
 		preg_match_all(
-			'#'.preg_quote($this->l_delim.$variable.$this->r_delim).'(.+?)'.preg_quote($this->l_delim.'/'.$variable.$this->r_delim).'#s',
+			'#'.preg_quote($this->l_delim).'?'.preg_quote($this->l_delim.$variable.$this->r_delim.$this->r_delim).'?(.+?)'.preg_quote($this->l_delim).'?'.preg_quote($this->l_delim.'/'.$variable.$this->r_delim.$this->r_delim).'?#s',
 			$string,
 			$matches,
 			PREG_SET_ORDER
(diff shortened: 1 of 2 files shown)


narfbg:
I see what you're doing with escape by default, but that's a huge BC break and the escaping itself is not really sensible. Even otherwise, you should use a templating engine like Twig if you want such functionality.

CI's Parser class is meant to be dead simple and if we start accepting changes like these, we're also dedicating ourselves to adding and supporting more and more templating features and that's just beyond our targets. In fact, I've thought a few times already about dropping the Parser class in favor of external templating engines for those same reasons, but that will probably happen in CI4.

kakysha:
@narfbg for what reason does CI need templating engine at all? Why not just use PHP in views, really? One more absolutely useless abstraction layer. For whom it is intended? Designers? Pfff. Replacing $var with {{var}} and require_once() with @include() makes makes no sense for me.

vojtatranta:
@kakysha well, I don't want to create flame here, but php framework that does not use templating engine that cares about escaping can be hacked under 5 seconds. 
Funny thing is that even sophisticated templating engine that by default everything escapes cannot ensure that attacker wont be able to run his code on your site via XSS or something.
The true purpose of existence of templating engines is not simplification of the way we write html but lack of security layer when printing stuff via PHP to html.

LouisMilotte:
@vojtatranta citation for hack claim?

vojtatranta:
@LouisMilotte well, you need citation for  it? 
If you do not use templating engine, you have to manually escape every string in every template, you can't even use <?= ?>.
If someone saves his username as <script>alert('Sucessfull xss')</script> then mostly this code will be executed (or it might com as parameter from URL).
Even Codeigniter's official tutorial does not escape strings.
Thats extremely dangerous, we all should know about that.

mwhitneysdsu:
@vojtatranta CI's Parser library doesn't escape the output for you, either. It's relatively simple to override it to do so, but, in the end, you're really left to your own devices to secure output in CI.

LouisMilotte:
@vojtatranta  that's not a citation; citation please.

vojtatranta:
@mwhitneysdsu thats what I am talking about.
@LouisMilotte Well I hope that explanation would be enough, XSS is usually used to steal session of logged user, I think you can find yourself some articles about it.
If you really want a citation I can write article about how you can simply steal session from some administrators using XSS and Javascript.

LouisMilotte:
@vojtatranta more so as to why template engines resolve that. At this point it's probably best to just point me at one of these "engines" that achieve what you're talking about rather than flooding this issue log.

vojtatranta:
@LouisMilotte Very well, then mostly commonly use in PHP world is propabably [Smarty](http://www.smarty.net/) or you can also use [Mustache/handlebars](https://github.com/XaminProject/handlebars.php) Symfony relies on [Twig](http://symfony.com/doc/current/book/templating.html) Laravel has nice engine called [Blade](http://laravel.com/docs/4.2/templates) and last but not least, czech framework Nette has awesome templating library which I use in CI if I have something in legacy code, it's called [Latte](https://github.com/nette/latte).
All of them can be installed via composer. There is also fork of smarty that is [meant to be used in CI](https://github.com/Vheissu/Ci-Smarty).
Using bold PHP (unescaped, manually escaped) HTML I consider huge hole in security, so huge that I would fear to use such app.

narfbg:
> well, I don't want to create flame here, but

"I don't want to ... but here it is". You resurrected a 4-months old discussion for the sake of argument.

Please don't use our issue tracker as a forum board.

PR #3646, opened 2015-03-06
Fix #3642

### LINKED ISSUE #3642: Can't load config file with same name in Package after add_package_path()
class CI_Config {
function load($file = '', $use_sections = FALSE, $fail_gracefully = FALSE){
$file = ($file === '') ? 'config' : str_replace('.php', '', $file);
...
}
}

For example:
i have a config.php file in - PHP\application\third_party\my_app\config\config.php

So i add this package, and try to load the file
$this->load->add_package_path(APPPATH . 'third_party/my_app');
$this->config->load('config');

But in this case, CI will always return TRUE on line 131 in CI_Config and skip to load package/config
Because, PHP\application\config\config.php is loaded at the beginning.

I don't know this is a bug or not. I fix it by create a function pass the package config path to load the config file.

wai:
Fix 1. when load the main config file, it will not add into CI_Config::is_loaded. once load other package/config.php, the application/config/config.php will include again and merge into current config. if application/config/config.php base_url is empty, here will jump over automatically set the base_url if none was provided in config class constructor and override the base_url to empty.
Fix 2. when use the same name, it will stop end quit the function when the file use a same file, change return true to continue to make sure check all config paths.

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -247,12 +247,14 @@ if ( ! function_exists('get_config'))
 			{
 				$found = TRUE;
 				require($file_path);
+				$config['__is_loaded'][] = $file_path;
 			}
 
 			// Is the config file in the environment folder?
 			if (file_exists($file_path = APPPATH.'config/'.ENVIRONMENT.'/config.php'))
 			{
 				require($file_path);
+				$config['__is_loaded'][] = $file_path;
 			}
 			elseif ( ! $found)
 			{
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -102,7 +102,14 @@ class CI_Config {
 
 			$this->set_item('base_url', $base_url);
 		}
-
+		
+		// Add loaded confing path to list 
+		if (isset($this->config['__is_loaded'])) 
+		{
+        		$this->is_loaded = $this->config['__is_loaded'];
+        		unset($this->config['__is_loaded']);
+        	}
+        	
 		log_message('info', 'Config Class Initialized');
 	}
 
@@ -128,7 +135,7 @@ class CI_Config {
 				$file_path = $path.'config/'.$location.'.php';
 				if (in_array($file_path, $this->is_loaded, TRUE))
 				{
-					return TRUE;
+					continue;
 				}
 
 				if ( ! file_exists($file_path))
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -103,7 +103,7 @@ class CI_Config {
 			$this->set_item('base_url', $base_url);
 		}
 		
-		// Add loaded confing path to list 
+		// Add loaded config path to list 
 		if (isset($this->config['__is_loaded'])) 
 		{
         		$this->is_loaded = $this->config['__is_loaded'];


vkeranov:
This is already fixed. Check this:
https://github.com/bcit-ci/CodeIgniter/commit/137aa20e0b0fd71ff8f672c57c07c4972c91c6a4

wai:
I have tried, and i found still have some problem on it.
This patch is base on 137aa20, and try to fix on it.

narfbg:
#3642 is already fixed, what you've described here is an issue specifically with base_url.

But the solution is quite hacky ... I'd rather revert the fix for #3642 than accept it. Will see if I can come up with a better one later.

wai:
Thanks for your work, i want to point out #3642 still need . 45c15b5 (Fix 2 ) to make sure check all config paths. 
.137aa20 fixed that when i add all third party, and call load config, it will load all once. 
(BUT it is because missing PHP\application\config\config.php in CI_Config::is_loaded so i can use config.php to load it in first time only)
So if i have 3 third party added with same config file name but load it one by one, it will only load the first config only.

Here is the test case for ref.

core/Config.php
120: is_loaded Array ( ) 
128: Try to check and load =>PHP\application\config\other_config.php
139: include = PHP\application\config\other_config.php
159: config array merge
128: PHP\application\config\development\other_config.php
135: file_exists = PHP\application\config\development\other_config.php

MY TEST (MY_APP1)- PHP\application\third_party\MY_APP1 config->load('config')
120: is_loaded Array ( [0] => PHP\application\config\other_config.php ) 
128: Try to check and load =>PHP\application\config\config.php
139: include = PHP\application\config\config.php
139: config array merge
128: PHP\application\config\development\config.php
135: file_exists = PHP\application\config\development\config.php
128: PHP\application\third_party\MY_APP1\config\config.php
139: include = PHP\application\third_party\MY_APP1\config\config.php
159: config array merge
128: PHP\application\third_party\MY_APP1\config\development\config.php
135: file_exists = PHP\application\third_party\MY_APP1\config\development\config.php
FINISH - PHP\application\third_party\MY_APP1

MY TEST (MY_APP2)- PHP\application\third_party\MY_APP2 config->load('config')
120: is_loaded Array ( [0] => PHP\application\config\other_config.php [1] => PHP\application\config\config.php [2] => PHP\application\third_party\MY_APP1\config\config.php ) 
128: Try to check and load =>PHP\application\config\config.php
130: in_array(file_path, this->is_loaded, TRUE) return true [Baceuse]=> PHP\application\config\config.php
FINISH - PHP\application\third_party\MY_APP2

MY TEST (MY_APP3)- PHP\application\third_party\MY_APP3 config->load('config')
120: is_loaded Array ( [0] => PHP\application\config\other_config.php [1] => PHP\application\config\config.php [2] => PHP\application\third_party\MY_APP1\config\config.php ) 
128: Try to check and load =>PHP\application\config\config.php
130: in_array(file_path, this->is_loaded, TRUE) return true [Baceuse]=> PHP\application\config\config.php
FINISH - PHP\application\third_party\MY_APP3

Sorry for my hacky suggestion, Thanks again.

narfbg:
That is intentional, all configs should only be loaded once.

wai:
Yes, configs should loaded only once.
But, current situation is application\third_party\config\config.php will never load when use load function independently.

Your fix in this case is ok
$this->load->add_package_path(APPPATH . 'third_party/MY_APP1');
$this->load->add_package_path(APPPATH . 'third_party/MY_APP2');
$this->load->add_package_path(APPPATH . 'third_party/MY_APP3');
$this->config->load('config');

finally loaded result -> 
application\config\config.php
application\third_party\MY_APP1\config.php
application\third_party\MY_APP2\config.php
application\third_party\MY_APP3\config.php

But...
$this->load->add_package_path(APPPATH . 'third_party/MY_APP1');
$this->config->load('config');
$this->load->add_package_path(APPPATH . 'third_party/MY_APP2');
$this->config->load('config');
$this->load->add_package_path(APPPATH . 'third_party/MY_APP3');
$this->config->load('config');

finally loaded result -> 
application\config\config.php
application\third_party\MY_APP1\config.php

track->
$this->load->add_package_path(APPPATH . 'third_party/MY_APP1');
$this->config->load('config');
---import application\config\config.php & MY_APP1\config.php > base_url empty happen

$this->load->add_package_path(APPPATH . 'third_party/MY_APP2');
$this->config->load('config');
---import nothing, return true on line 131 
>will not import MY_APP2\config.php because application\config\config.php in array!

$this->load->add_package_path(APPPATH . 'third_party/MY_APP3');
$this->config->load('config');
---import nothing, return true on line 131

So i want to point out to solve this by change one thing.
45c15b5 Config.php Line 131 return TRUE; should be continue; for keep going other paths.
Line 105-112 ['__is_loaded'] just my hacky idea, not in this case.

narfbg:
The more I think about it, the less I see a point in "overwriting" config.php at runtime ... that's the one config that *must* be present for your application to function at all, and decisions based on it have already been made before you could ever call add_package_path().

So why do you want to do that?

PR #3655, opened 2015-03-09
Updated CI Security unit test 

ahnh:
Improved Code Coverage from 54.55% to 77.27% for CI_Security.

### DIFF
--- a/tests/codeigniter/core/Security_test.php
+++ b/tests/codeigniter/core/Security_test.php
@@ -73,44 +73,44 @@ class Security_test extends CI_TestCase {
 
         // --------------------------------------------------------------------
 
-        public function test_xss_clean_string_array()
+	public function test_xss_clean_string_array()
 	{
-                $harm_strings = array(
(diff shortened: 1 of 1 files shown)


tests/codeigniter/core/Security_test.php
```diff
@@ -71,12 +71,69 @@ public function test_xss_clean()
...
+                $harm_string = '<img src="test.png">';
+
+		$xss_clean_return = $this->security->xss_clean($harm_string, TRUE);
+
+		$this->assertTrue($xss_clean_return);
+	}
+
+        // --------------------------------------------------------------------
+
+        public function test_xss_clean_image_invalid()
```
narfbg: Looks like you've used spaces for indentation for everything but the curly braces.

Can you convert those to tabs? That's what we use for indentation, it's explained in our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html).

ahnh: I converted to tabs.

PR #3665, opened 2015-03-13
Update Security_test.php

voku:
Heyho, 

only for explanation => I forked the "Security"-Class yesterday and I created a new repo: "https://github.com/voku/anti-xss" and I added some more tests, that maybe not 100% compatible with CodeIgniter. However, these attacks should also intercept, or?!

Mfg Lars


narfbg:
Well, if you've forked it, I suppose it is to apply fixes for these errors and not just the test. If so, why didn't you submit them?

voku:
... because I thought that I changed to much code :P by extracting the "XSS"-functions out of the "Security"-Class that do much more then anti-xss ...

voku:
@narfbg one question, is the "CI_Utf8"-Class ready for production code? The reason I ask is because before this commit is seems that it's only used for some testing-code ... I have also added some more tests for the UTF-8 Class but they failed -> https://github.com/voku/CodeIgniter/commit/25c175f83673e275b6b398c0f467a55c847d2935

What do you think about a external UTF-8 Class? e.g.: https://github.com/voku/portable-utf8 ;)

system/core/Security.php
```diff
@@ -362,6 +362,12 @@ public function xss_clean($str, $is_image = FALSE)
 		// Remove Invisible Characters
 		$str = remove_invisible_characters($str);
 
+		// Removes all non-UTF-8 characters
+		// && Remove NULL characters (ignored by some browsers).
+		$utf8 = new CI_Utf8();
```
narfbg: What has that got to do with XSS?

system/core/Security.php
```diff
@@ -390,6 +396,9 @@ public function xss_clean($str, $is_image = FALSE)
 		// Remove Invisible Characters Again!
 		$str = remove_invisible_characters($str);
 
+		// Remove Netscape 4 JS entities.
```
narfbg: Netscape 4?!

system/core/Security.php
```diff
@@ -664,15 +673,21 @@ public function entity_decode($str, $charset = NULL)
 
 				$replace = array();
 				$matches = array_unique(array_map('strtolower', $matches[0]));
-				for ($i = 0, $c = count($matches); $i < $c; $i++)
+
+				$c = count($matches);
```
narfbg: Why all of these changes?

system/core/Utf8.php
```diff
@@ -79,6 +83,60 @@ public function __construct()
...
 
+
+	/**
+	 * normalize whitespace
+	 *
+	 * @param string $str The string to be normalized.
+	 *
+	 * @return string
+	 */
+	public function normalize_whitespace($str)
```
narfbg: What is this for?

system/core/Utf8.php
```diff
@@ -59,6 +59,10 @@ class CI_Utf8 {
 	 */
 	public function __construct()
 	{
+		if (defined('UTF8_ENABLED')) {
```
narfbg: That UTF8_ENABLED is FALSE doesn't mean that you shouldn't be able to use the class ...

narfbg:
>  is the "CI_Utf8"-Class ready for production code? The reason I ask is because before this commit is seems that it's only used for some testing-code ...

It might be the case that we've eliminated its usage, but that doesn't mean it's not production ready.

> What do you think about a external UTF-8 Class? e.g.: https://github.com/voku/portable-utf8 ;)

Bundle it with the framework? No.

Anyway, please keep this PR to XSS-related changes only.

system/core/Utf8.php
voku: I see the error-msg that the constant "UTF8_ENABLED" is already defined ... so I skipped the constructor at this point ...

system/core/Security.php
voku: Because I see a php-warning "Undefined offset" in this code-part ...

system/core/Security.php
voku: ported from Drupal.org ;) -> https://api.drupal.org/api/drupal/core!lib!Drupal!Component!Utility!Xss.php/function/Xss%3A%3Afilter/8

system/core/Security.php
voku: "ignored by some browsers" -> e.g.: "<scr\0ipt>" 

... or take a look at the test-cases e.g: https://github.com/voku/CodeIgniter/blob/develop/tests/codeigniter/core/XssTest.php#L570

system/core/Utf8.php
voku: Normalize UTF-8 "NULL" / Space" characters, so we can search for e.g.: "\<script\>" ...

system/core/Security.php
narfbg: That's what the `remove_invisible_characters()` call is for.

system/core/Security.php
narfbg: We don't copy something just because Drupal or another framework has it ... if you want to change or add something, especially in a security sensitive piece of code, you have to know why it is necessary and explain it.

system/core/Security.php
narfbg: And you see it even with this patch from a month and a half ago?

074a214ee829e2169058b7f07efdd44edcc3fc4f

system/core/Utf8.php
narfbg: You only see it because you're instantiating the class twice, which you're not supposed to do.

system/core/Utf8.php
narfbg: As I said, `remove_invisible_characters()`.

narfbg:
You assuming that everything is UTF-8, which is not necessarily the case.

You're also quite strangely using 2 spaces for indentation in one file (we use tabs), tabs or a different number of spaces in another, moving tests from one file to another (making it hard to spot the actual changes) ...

I've got the feeling we can finish this faster if you explain what you're trying to fix, so I can work on it as well.

system/core/Security.php
voku: But it dosn't work for UTF-8 Strings: remove it and run the test again ->

Expected :<p arg="">" style="background-image: url([removed]alert&#40;0&#41;);"<p>
Actual   :<p arg="�">" style="background-image: url(j�avas cript:alert&#40;0&#41;);"�<p>

... and some Browser (IE) seems to ignore this e.g. null-chars :/

narfbg: Null chars (\x0) are surely stripped, and the test that you're referring to uses \xe0, which is a letter character, not an invisible or invalid one. Also, we don't support Internet Explorer 6 (a comment in your testcase says that it only works on IE6), nor Netscape Navigator 4 (really? :)).

Now, surely if there's a valid flaw based on UTF-8 characters that we don't handle - we need to account for it, but you also can't assume that everything is UTF-8 which you are doing here.

voku: IE7, old versions of Firefox and Netscape 8 seems to be affected, too. :/ -> https://books.google.de/books?id=FKN5uL57tyAC&printsec=frontcover&hl=de#v=onepage&q&f=false

narfbg: Can you be a little more specific? You can't expect me to read the whole PDF. And if IE7 and some "old version of Firefox" is also affected, why does your comment (code comment) mention only IE6?

narfbg:
Added FSCommand and seekSegmentTime to evil attributes list: aebd039a61cb5135b31ab0b8d9d95ed3fb678c7b

However, none of your other proposed changes make sense for us (another of our team members has also reviewed your patch). I'll wait for your response a bit more before closing.

system/core/Security.php
voku: The php-warning still exists, because of "array_unique()" ...

e.g. 
-> array(1 => 'a', 2 => 'b', 3 => 'b', 4 => 'c')
-> array(1 => 'a', 2 => 'b', 4 => 'c')

... the problem arises when we use a "for"-loop, because there is !3 anymore. Maybe we can use a foreach loop here?

narfbg: The warning was caused by the count being computed before `array_unique()`, which is no longer the case after 074a214ee829e2169058b7f07efdd44edcc3fc4f

voku:
"I've got the feeling we can finish this faster if you explain what you're trying to fix, so I can work on it as well." ->

 ... I also think so, maybe my "thinking error" is / was that we can protect the string if we convert UTF-NULL-Chars into "normal"-chars. 

Demo: https://html5sec.org/test/?c=utf-8#61

  ```Test: <p arg=\"\xe0\">\" style=\"background-image: url(j\xe0avas\xc2\xa0cript:alert(0));\"\xe0<p>```
  ```Expected :<p arg="">" style="background-image: url([removed]alert&#40;0&#41;);"<p>```
  ```Actual   :<p arg="�">" style="background-image: url(j�avas cript:alert&#40;0&#41;);"�<p>```

But the old IE will ignore the UTF-NULL-Char and so you will still see the "xss-alert".

system/core/Security.php
voku: Maybe I'm wrong, but please read again what I have written above, thx! We shouldn't iterate (via "for"-loop) over a array from "array_unique()", because it's maybe not a "numeric array" anymore ...

narfbg:
I'll say it again - \xe0 is not a NULL character and we simply don't care about old IE or Netscape browsers.

system/core/Security.php
narfbg: Well, you are wrong ... There's no way that you pass a numeric array to `array_unique()` and get a non-numeric array, and that was never the cause for that notice anyway.

narfbg: On another note however, yes - I didn't consider that it doesn't re-index the keys.

voku: 1.) So if we use a "foreach"-loop the php-warning should disappear, however for which use-case is this code-section anyway? My unit-tests are also ok, if I remove the code for -> "// Decode standard entities, avoiding false positives" 

2.) And it looks like it didn't works anyway (for old php-versions) because there is a typo in ...
```
...
$_entities[':'] = '&colon;';
$_entities['('] = '&lpar;';
$_entities[')'] = '&rpar';
$_entities["\n"] = '&newline;';
$_entities["\t"] = '&tab;';
...
```

I think we need to remove the ";" at the end of the string, because in "array_search($matches[$i] . ';', $_entities, true);" it will be added again?

narfbg: There is ... not a typo, but a missing semicolon. The whole point is to add it in case that it's missing - that's what this whole method is made for.

068ab206d84bf1668832988932dc61cfb3103bf8

voku: Ok, now I see it! "$match" becomes the missing semicolon... and thanks for the quick-fix! :) I think we can close this "pull"-request.

PR #3675, opened 2015-03-16
Unit test ci input

ahnh:
Input.php " _fetch_from_array" function is fixed and now fully covered by unit test.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -191,25 +191,28 @@ class CI_Input {
 		{
 			$value = $array[$index];
 		}
-		elseif (($count = preg_match_all('/(?:^[^\[]+)|\[[^]]*\]/', $index, $matches)) > 1) // Does the index contain array notation
+		elseif (($count = preg_match_all('/(?:^[^\[]+)|\[[^]]*\]/', $index, $matches)) > 1) // Does the index contain array notation (Eg. Foo[Bar])
 		{
 			$value = $array;
+			$found = False;
 			for ($i = 0; $i < $count; $i++)
 			{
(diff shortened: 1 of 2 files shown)


narfbg:
What was broken that is being fixed?

ahnh:
Inside of the for loop, it wasn't checking all the items, if the first item was empty or it's not set.

narfbg:
I don't quite understand what you mean, but it is supposed to stop if it only sees brackets, as there's no sensible way to resolve them ... please revert that and leave only the tests.

Also, your last assertion doesn't represent "empty notation". It may seem as a nitpicky detail, but it's a non-existing value, which isn't the same thing. :)

ahnh:
With the original function, what do you expect the possible keys with the given input 'abc[def]'?

Currently if you give the 'abc[def]' the function will use "abc" as the first key. If the "abc" isn't in the data set, it will return NULL without checking the "def" as another key. This is why I changed so that it can also check the "def" as the key.

ahnh:
I'll revisit it and make another PR later.

narfbg:
It's a recursion of a kind ...

Let's assume we've got a `$this->input->post('abc[def]')` call. In the first iteration, `$value` will be `$_POST` and therefore we'll be checking if `$_POST['abc']` exists - return NULL if it doesn't or do `$value = $_POST['abc']` and go to the next key ... And because of that, `isset($value[$key])` will be in fact looking for `$_POST['abc']['def']`, which is what the whole thing is supposed to do. :)

ahnh:
I now see the point. Thanks Andrey, I'll rework on it.

PR #3678, opened 2015-03-17
Added CI_Form_validation unit tests

djwoods:
Added setup and tests for all rules specified in the documentation.
Several test cases currently fail. I'll create bug issues & fixes once this pull request is accepted.

Signed off:
djwoods <d.woods92@gmail.com>

### DIFF
--- a/tests/codeigniter/libraries/Form_validation_test.php
+++ b/tests/codeigniter/libraries/Form_validation_test.php
@@ -11,8 +11,7 @@ class Form_validation_test extends CI_TestCase {
 		$loader = $this->getMock('CI_Loader', array('helper'));
 		// At current, CI_Form_Validation only calls load->helper("form")
 		// Assert this so that if that changes this fails fast
-		$loader->expects($this->once())
-				->method('helper')
+		$loader->method('helper')
 				->with($this->equalTo('form'));
 		// Same applies for lang
 		$lang = $this->getMock('CI_Lang', array('load'));
@@ -34,8 +33,9 @@ class Form_validation_test extends CI_TestCase {
 
 	public function test___construct()
 	{
-		$this->assertNotNull($this->form_validation);
-	}
+		$form_validation = new CI_Form_validation();
+		$this->assertNotNull($form_validation);
+	}	
 
 	public function test_rule_required()
 	{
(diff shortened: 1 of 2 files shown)


tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+        $ldr->expects($this->once())
+                ->method('helper')
+                ->with($this->equalTo('form'));
+        
+        // Same applies for lang
+        $lang = $this->getMock('CI_Lang', array('load'));
+        
+        // Setup CI_Input
+        // Set config for Input class
+        $this->ci_set_config('allow_get_array',	TRUE);
```
narfbg: I'm pretty sure you don't need these.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+        $this->ci_set_config('global_xss_filtering', FALSE);
+        $this->ci_set_config('csrf_protection', FALSE);
+
+        $security = new Mock_Core_Security();
+
+        $this->ci_set_config('charset', 'UTF-8');
+        $utf8 = new Mock_Core_Utf8();
+
+        $inp = new Mock_Core_Input($security, $utf8);
```
narfbg: `$inp` is a rather WTF name ... :)

It's not important, but you'd have to make changes to the PR anyway, so you might as well change that to `$input`, as well as `$ldr` to `$loader`.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+    
+    public function test_rule_required()
+    {
+        $this->assertTrue($this->run_rule('required', ' someValue'));
+        
+        $this->assertFalse($this->run_rule('required', ''));
+        $this->assertFalse($this->run_rule('required', ' '));
+    }
+    
+    public function test_rule_matches()
```
narfbg: This whole test is incorrect ... The 'matches' rule matches against another field's value, not an arbitrary string.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+    {
+        // Empty input should pass any rule unless required is also specified
+        $this->assertTrue($this->run_rule('matches[sample]', ''));
+        $this->assertTrue($this->run_rule('matches[s]', 's'));
+        
+        $this->assertFalse($this->run_rule('matches[Sample]', 'sample'));
+        $this->assertFalse($this->run_rule('matches[sample]', ' sample'));
+    }
+    
+    public function test_rule_differs()
```
narfbg: Same as `test_rule_matches()`.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+    {
+        // Empty input should pass any rule unless required is also specified
+        $this->assertTrue($this->run_rule('numeric', ''));
+        $this->assertTrue($this->run_rule('numeric', '0'));
+        $this->assertTrue($this->run_rule('numeric', '12314'));
+        
+        $this->assertFalse($this->run_rule('numeric', '123a'));
+    }
+    
+    public function test_rule_integer()
```
narfbg: Should also test if it accepts negative values (it should).

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+    }
+    
+    public function test_rule_valid_url()
+    {
+        // Empty input should pass any rule unless required is also specified
+        $this->assertTrue($this->run_rule('valid_url', ''));
+        $this->assertTrue($this->run_rule('valid_url', 'www.codeigniter.com'));
+        $this->assertTrue($this->run_rule('valid_url', 'http://codeigniter.eu'));
+   
+        $this->assertFalse($this->run_rule('valid_url', 'codeigniter'));
```
narfbg: I know this is strange, but 'codeigniter' is a valid URL according to RFC2396.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -0,0 +1,314 @@
...
+        $this->assertFalse($this->run_rule('valid_ip[ipv6]', '127.0.0.1'));
+        $this->assertFalse($this->run_rule('valid_ip', 'H001:0db8:85a3:0000:0000:8a2e:0370:7334'));
+        $this->assertFalse($this->run_rule('valid_ip', '127.0.0.259'));
+    }
+    
+    public function test_rule_valid_base64()
+    {
+        // Empty input should pass any rule unless required is also specified
+        $this->assertTrue($this->run_rule('valid_base64', ''));
+        $this->assertTrue($this->run_rule('valid_base64', 'string'));
```
narfbg: 'string' is not a valid Base64-encoded string ...

narfbg:
In addition to the inline comments, you should also read our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html) and adjust the patch according to it. For starters, we use tabs for indentation, not 4 spaces.

Also, you should test for 'required' only once and quite a lot of the other rules can be tested by directly calling them via `$this->form_validation->rule_name($input, $optional_parameter)` - you don't need to set them all up via `set_rules()` and such. In fact, the 'matches' and 'differs' rules are one of the few that require that because (as I said) they match against other fields.

tests/codeigniter/libraries/Form_validation_test.php
djwoods: The mock loader with the helper function stubbed is required because without it, CI_Form_validation is unable to load the form_helper helper. 
Similarly, the mock CI_Lang is needed to load form_validation

I've removed the unnecessary config stuff though (I'm guilty of copy/pasting the setup for CI_Input)

narfbg: Yes, I was referring to the config values ... :)

djwoods:
Aside from tabs and a handful of poor variable names, I didn't see anything in the style guide that I was violating. Am I missing something?

As for the ability to test rules by directly calling their methods, I feel the current tests provide a broader (and better) test since they verify the public facing API is working as expected. I understand this is more of a user scenario test than a unit test but I don't see as much value in testing the rules directly since most are 1 or 2 lines and simply call a native method. Would you like me to switch them to testing the rule methods directly?

tests/codeigniter/libraries/Form_validation_test.php
djwoods: Ah. It was hard to see that since it just shows me 16 lines of code and your comment. It doesn't link the specific lines to which you're commenting on

narfbg: Yeah, the comment is always under the line that it was intended for ...

narfbg:
> Aside from tabs and a handful of poor variable names, I didn't see anything in the style guide that I was violating. Am I missing something?

No, that's about it. You're still using spaces on lines 11 and 14-16 though ...

> As for the ability to test rules by directly calling their methods, I feel the current tests provide a broader (and better) test since they verify the public facing API is working as expected. I understand this is more of a user scenario test than a unit test but I don't see as much value in testing the rules directly since most are 1 or 2 lines and simply call a native method. Would you like me to switch them to testing the rule methods directly?

Yes, please. That most of them are 1-2 lines is one reason why tests for this library have never been our priority, but they still have to be tested separately. The benefit is that when a test fails you know *exactly* which piece of code has failed, while when you do a scenario test, a single error in e.g. `set_rules()` may trigger a failure in tens of other test cases, which are completely fine in isolation.

djwoods:
Sorry for the stream of commits. I hadn't realized it would automatically update the pull request (though in hindsight it seems pretty obvious!). 

I'll change the tests to test the rule functions directly sometime later this week.

Also, the matches and differs tests are still failing. It seems the code relies on a rule being set for every field.  $_POST data for fields that don't have a rule don't get copied into the field_data array, which is where the match and differ function look for data to compare to.

djwoods:
@narfbg I've made the changes you requested and added a handful of extra tests.
Once this has passed review & been merged, I'll add more test cases to improve coverage

narfbg:
A large improvement, indeed.

But you're still not correctly testing 'matches' and 'differs' ... they will always fail because their logic isn't to compare against `$_POST` like you've assumed it is. If that was the case, we wouldn't have `set_data()` (where the test case is also half wrong, empty arrays are ignored).

Only fields that *have rules defined* for them are taken into account, so if you want to test if 'foo' differs/matches 'bar', you have to define both fields. Of course, the case where a field being compared to is not defined should also be tested (and it should assert to FALSE), but it is still a different test case.

djwoods:
Looking at the code I can see that the test cases are "wrong". However, I'm trying to write test codes based on the documentation. There is no mention in the documentation that both fields have to have rules for match & differ to work. If this is the desired behaviour, it should be documented.

Likewise there is no mention of set_data() doing nothing when the array is empty. In fact, it seems to me that this could cause a very nasty hidden bug, since the previous data is still used, despite me calling reset_validation() and set_data() with an empty array (which may have been dynamically created, and not known to be empty). 

I'll make the changes as you've requested, but the documentation should notify the developer of these requirements / behaviours

narfbg:
It's implied ... why would you need to compare against a field that you're not validating?

djwoods:
What about a security question or security image? I'd want to match the user's input with information that I'm likely to grab from a database, put into an array along with the data from $_POST, and then validate using set_data()? This test case isn't testing this scenario, since it's using $_POST and not set_data(), but this case would still fail since set_data() relies on the same logic that both fields would have rules.

narfbg:
Not designed for captchas ...

djwoods:
Is there a reason you expect users to not have captchas on their forms?
Also, I've updated the tests as per your request

narfbg:
I'm not saying users wouldn't have captchas, I'm saying these rules serve for a different purpose. I'll review again and probably merge a bit later.

PR #3680, opened 2015-03-18
Fixed a typo in uri.rst documentation

schwemmer:
Changed "and" to "an". Editing the file online on github.com shows the last line of the file as changed, even though I didn't change it at all. It keeps doing that so I don't know how to avoid it. What I really changed is only line 113.

### DIFF
--- a/user_guide_src/source/libraries/uri.rst
+++ b/user_guide_src/source/libraries/uri.rst
@@ -110,7 +110,7 @@ Class Reference
 		:returns:	Associative URI segments array
 		:rtype:	array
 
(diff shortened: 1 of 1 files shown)


ZombieProtectionAgency:
Just FYI The github editor adds a new line to the end of the file if there isnt one already. I dont know what the style guidelines for documentation is and I don't think there is a way around it but, I think, for the actual source they switched from comment at EOF to new line at EOF.

schwemmer:
I see, that explains it. Thanks!

PR #3682, opened 2015-03-18
Updated CI_Input unit test and fixed error "undefined offset" 

ahnh:
The error was caused by using the same variable name, $i, twice for for loop inside for loop.

Signed-off-by:Heesung Ahn <ahn.heesung@gmail.com>

### DIFF
--- a/tests/codeigniter/core/Input_test.php
+++ b/tests/codeigniter/core/Input_test.php
@@ -250,9 +250,10 @@ class Input_test extends CI_TestCase {
 		$this->assertEquals('127.0.0.1', $this->input->ip_address());
 		
 		$this->input->ip_address = FALSE;
-		$this->ci_set_config('proxy_ips', 'http://foo/bar/baz, 127.0.0.1/1');
+		$this->ci_set_config('proxy_ips', 'http://foo/bar/baz, 127.0.0.2');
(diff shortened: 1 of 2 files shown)


tests/codeigniter/core/Input_test.php
```diff
@@ -206,9 +206,23 @@ public function test_set_cookie()
...
 	{
+		//TODO: Find a way to test HTTP headers
+		$this->assertNull($this->input->get_request_header('test'));
+	}
+	
+	// --------------------------------------------------------------------
+	
+	public function test_ip_address()
+	{	
+		$this->input->ip_address = TRUE;
```
narfbg: It's not supposed to have this value, ever ...

ahnh: I've updated test_get_request_header function.

narfbg: What has that got to do with ip_address?

ahnh: Updated to real ip address. 

The initial thought was because of the ip_address variable was set to "FALSE" initially and there is line check whether its FALSE, so I was testing it with TRUE. 
if ($this->ip_address !== FALSE)
{
            return $this->ip_address;
}

PR #3683, opened 2015-03-19
Enable sub-directory for default_controller

bcphung:
Not sure if a bug or it means to be, well, following the general changelog for CI 3, multiple levels of controller directory is on, and one piece still missing is `$route['default_controller']` out of sight.

Let's take an example:
`$route['default_controller'] = 'backend/dashboard';` will result that _backend_ is the controller (_Backend.php_) instead of _Dashboard.php_
So, the new method will working on purpose.

### DIFF
--- a/system/core/Router.php
+++ b/system/core/Router.php
@@ -295,13 +295,26 @@ class CI_Router {
 			show_error('Unable to determine what should be displayed. A default route has not been specified in the routing file.');
 		}
 
-		// Is the method being specified?
-		if (sscanf($this->default_controller, '%[^/]/%s', $class, $method) !== 2)
+		// Is there sub-directory included?
+		if (strpos($this->default_controller, '/') !== FALSE)
 		{
+			$segments = explode('/', $this->default_controller);
+			
+			// Get valid and return correct class name
+			$class = $this->_validate_request($segments)[0];
 			$method = 'index';
 		}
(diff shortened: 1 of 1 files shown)


narfbg:
The bug is that you were ever able to put a directory in there, not that you can't do it now.

bcphung:
With the release of rc3, I faced a problem:
`$route['default_controller'] = 'directory/controller';` this always parsed that `directory` is the controller here.

With this commit, I can do the magic.

narfbg:
I know what your commit does, but you're not supposed to do that "magic".

Again, that you were ever able to use it as directory/controller was due to a bug, and it has never been a fature. 'default_controller' was always a controller/method pair.

bcphung:
Then, index.php is the place that I have to come for.

Well, I will keep this for myself.

Thanks.

kuez:
I need to reo open this bug because i notice that the marge of code is not correct, i checked the file /system/core/Router.php with the commit ba71bee241065782d312afeecba1f4c2981bc213  and is not the same :( , i dowloaded CI v3.1.8 from github. Thanks

image

bcphung:
I'm sorry @kuez, you are looking at a closed PR, which was not approved.
If you want patching around for your own usage, just copy-n-paste as of the diff.

PR #3696, opened 2015-03-24
system/core/Common.php, function html_escape

logsdon:
Fix: array_fill() throws an error if count($var) is 0
Signed-off-by: Joshua Logsdon <logsdon.joshua@gmail.com>

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -741,7 +741,6 @@ if ( ! function_exists('html_escape'))
 	 */
 	function html_escape($var, $double_encode = TRUE)
 	{
-		// If empty, skip escaping
 		if (empty($var))
 		{
 			return $var;
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -743,6 +743,12 @@ if ( ! function_exists('html_escape'))
 	{
 		if (is_array($var))
 		{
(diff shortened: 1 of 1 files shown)


narfbg:
I wouldn't put this inside the `if (is_array($var))` block ... any empty value (strings, integers, etc. included) should be immediately returned if we're going to check if anything's empty. But the obvious question is - why are you passing an empty array to `html_escape()` in the first place?

Also, no spaces around the empty() call please.

logsdon:
This wasn't an issue in CI 2.x for me but the new CI 3.x $html_escape param for set_value() is what lead me to this. If I'm expecting set_value() to return an array and I pass the $default param as array() is when it's happening to me, because it gets passed on to html_escape.

I was doing empty($var) really to do a count($var) == 0 check, but it makes sense like you said to check $var at the top.

Worth fixing and pull requesting again?

narfbg:
Just commit the changes and push them to the same branch, no need for a separate pull request.

logsdon:
Excellent. That should be updated now. Thank you.

narfbg:
Um ... could you also remove the comment line please. It's quite obvious what's going on.

logsdon:
Good to go.

PR #3700, opened 2015-03-25
Fix language overriding for calendar library

gadelat:
Fixes http://forum.codeigniter.com/thread-1552.html
Broken since 8d5b24a8c55dc1ae7721e10de094c4aba2ca7eae

### DIFF
--- a/system/libraries/Calendar.php
+++ b/system/libraries/Calendar.php
@@ -131,7 +131,7 @@ class CI_Calendar {
 	{
 		$this->CI =& get_instance();
 
-		if ( ! array_key_exists('calendar_lang.php', $this->CI->lang->is_loaded))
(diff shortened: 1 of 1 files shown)


system/libraries/Calendar.php
```diff
@@ -131,7 +131,7 @@ public function __construct($config = array())
 	{
 		$this->CI =& get_instance();
 
-		if ( ! in_array('calendar_lang.php', $this->CI->lang->is_loaded, TRUE))
+		if ( ! array_key_exists('calendar_lang.php', $this->CI->lang->is_loaded))
```
narfbg: `isset()` instead of `array_key_exists()`

vlakoff:
Nice catch :)

narfbg:
Indeed.

PR #3704, opened 2015-03-26
Changes to stored_procedure with oracle

mariofertc:
We had a trouble with the stored procedure, first the function was
strict with the package name, with this implementation the package is
optional, and the second is that the bind_by_name was failt, because it
was assigning two ids of connection, one time at binding and other when
the query excecute.


narfbg:
I just pushed a bugfix for the binding bug (you can see the commit linked above). Yours was based on CI2 and because of that, it actually made more breaks.

However, the package parameter is a larger issue than it seems and allowing an empty string to be passed would be a band-aid on a bullet wound. See #3981 for more details.

PR #3706, opened 2015-03-26
Fix language overriding

gadelat:
I think this is supposed to be done for 8d5b24a8c55dc1ae7721e10de094c4aba2ca7eae. I don't think this is a good way to do it, though. I think lang library should be changed to support this natively. Not sure of the best way. Open to suggestions.


narfbg:
No, none of these libraries should even know about the existence of `CI_Lang::$is_loaded`. In fact, that's something that I should've told you for your previous PR.

narfbg:
Also, this isn't a fix ... nothing's broken.

PR #3710, opened 2015-03-27
Fix an "strpos(): Empty needle" warning

toulet:
Signed-off-by: Cyrille TOULET <cyrille.toulet@linux.com>

### DIFF
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -205,14 +205,17 @@ class CI_URI {
 		$query = isset($uri['query']) ? $uri['query'] : '';
 		$uri = isset($uri['path']) ? $uri['path'] : '';
 
-		if (!empty($_SERVER['SCRIPT_NAME']) && strpos($uri, $_SERVER['SCRIPT_NAME']) === 0)
-		{
-			$uri = (string) substr($uri, strlen($_SERVER['SCRIPT_NAME']));
-		}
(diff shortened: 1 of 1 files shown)


system/core/URI.php
```diff
@@ -205,11 +205,11 @@ protected function _parse_request_uri()
 		$query = isset($uri['query']) ? $uri['query'] : '';
 		$uri = isset($uri['path']) ? $uri['path'] : '';
 
-		if (strpos($uri, $_SERVER['SCRIPT_NAME']) === 0)
+		if (!empty($_SERVER['SCRIPT_NAME']) && strpos($uri, $_SERVER['SCRIPT_NAME']) === 0)
```
narfbg: Needs to have a space on each side of `!`. Also, `empty()` would catch '0' or similar as well, which might be a really weird case, but still possible.

And if you'll be putting that in both conditions, it should really be just one if in which you'd nest these, so ...

    if (isset($_SERVER['SCRIPT_NAME'][0]))
    {
        // strpos ifs
    }

That would be the optimal solution.

toulet:
This part of code has been updated.

narfbg:
Now you're using 4 spaces for indentation. We use tabs ...

toulet:
Oops, it's fixed ;~)

PR #3715, opened 2015-03-30
Fixes pagination with relative URL

infinit89:
When base_url is a URL based on protocol, like "//www.google.com" the double slash regex kills the first "//".


narfbg:
I'd rather remove that line altogether and if there is a case where double slashes are produced - find out where and fix that instead.

PR #3718, opened 2015-03-30
Fix logged path to match checked path in loader

mwhitneysdsu:
This is to fix a logged path missed in
8f5c1780706113c926bb7801db27dbae97d00fcf


ftwbzhao:
@narfbg 
“Fix whitespace in previous commit”, I Got it.
Thanks :rose:

narfbg:
@ftwbzhao That has nothing to do with this thread ... keep it on topic.

PR #3721, opened 2015-03-31
update installation link

galdiolo:


narfbg:
The links is fine, it's the content on the website that needs to be changed.

PR #3727, opened 2015-04-03
Lowercase $side variable for $this->db->like() in Query Builder

yahyaerturan:
```$this->db->like('name',$value,'AFTER')``` returns ```LIKE '%$value%'```. Safer to lowercase for coding speed and developer habits.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -925,6 +925,9 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 				? $this->_group_get_type('') : $this->_group_get_type($type);
 
 			$v = $this->escape_like_str($v);
+			
+			// lowercase $side for user habits to write in UPPERCASE
+			$side = strtlower($side);
 
 			if ($side === 'none')
 			{


narfbg:
This is a CI-specific feature, so how could you possibly have developed a habbit to write 'AFTER' instead of 'after'?

yahyaerturan:
You are right, but most of developers are using various frameworks as I do. To remember all is not possible. Using uppercase in queries  may help to develop that habbit. Anyways, please easy :) It's just an advise to do slightly better ( as you all did the magnificent work already).

narfbg:
My point is - it's not a part of a query that you're writing here with 'after' ... The whole point of the query builder is that you don't write any part of a query anywhere, you just pass data.

Anyway, the build is failing because there's a typo in your patch.

yahyaerturan:
Saw the typo now, sorry. OK then, I am closing, thanks.

narfbg:
Don't close it ... just push a fix for it.

yahyaerturan:
OK I'll build a new pull request without typos. Thanks

yahyaerturan:
Actually #3729 is without typo. Is it ok?

narfbg:
No ... #3729 works only with `like()`, leaving `or_like()`, `not_like()`, etc. unchanged. This here is the proper one, just fix the typo ...

You don't need to open a new PR either, just commit & push to the same branch.

yahyaerturan:
I thinks it's done, thanks for your time and follow up.

narfbg:
Eh ... you also needed to reopen the PR.

yahyaerturan:
Openned! No: #3739. Title: #3727 Lowercase $side variable for $this->db->like() in Query Builder

narfbg:
You had to **re**open this one ...

PR #3729, opened 2015-04-03
CI_DB_query_builder::like() strtolower for $side variable

yahyaerturan:
$this->db->like('name',$value,'AFTER') returns LIKE '%$value%'. Safer to lowercase for coding speed and developer habits.


narfbg:
Don't submit 2 requests for the same thing ...

yahyaerturan:
Wasn't meant to, sorry. First one is mixed with old CI Repo so TRAVIS failed. Thanks.

PR #3730, opened 2015-04-03
Matched root_path's slashes with the name being replaced in the Zip::read_dir() library

paralogizing:
This pull request solves an issue with the $root_path's slashes not matching up with the $name it is replacing when $preserve_filepath is set to FALSE.

### DIFF
--- a/system/libraries/Zip.php
+++ b/system/libraries/Zip.php
@@ -352,7 +352,7 @@ class CI_Zip {
 		// Set the original directory root for child dir's to use as relative
 		if ($root_path === NULL)
 		{
-			$root_path = dirname($path).DIRECTORY_SEPARATOR;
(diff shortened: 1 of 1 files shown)


narfbg:
So, you're saying that `dirname('foo\\bar/baz/qux'))` would return ``foo\\bar/baz`` on Windows?

paralogizing:
No, but Windows is the problem here.  If you set the ``$path`` to ``'foo/bar/baz/qux'`` then the ``$root_path`` becomes ``'foo/bar/baz\'`` and the ``$name`` becomes ``'foo\bar\baz\qux'`` which doesn't match up when we ``str_replace('foo/bar/baz\', '', 'foo\bar\baz\qux')``.

What I am proposing is that we put the ``$root_path`` through the same ringer as the ``$name`` so that ``str_replace('foo\bar\baz\', '', 'foo\bar\baz\qux')`` will have the desired effect.

PR #3732, opened 2015-04-04
Check if session key exists before unseting.

sv3tli0:
Method unset_userdata is public so it will be good if we check existence of any key before unsetting it.

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -720,13 +720,18 @@ class CI_Session {
 		{
 			foreach ($key as $k)
 			{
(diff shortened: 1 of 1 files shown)


narfbg:
http://3v4l.org/NO4QW

sv3tli0:
I have made idiotic pull request sorry.
My idea was not to prevent notices in php4 :) I had to set keys validation.
If clients pass array containing nulls or object, there are fails as this unset doesn't check what it tries to unset..

Is it a good option to set some kind of type validation, as:
if(is_string() || is_int()) {}

PR #3735, opened 2015-04-04
Update doc links to userguide3

gadelat:
It should be decided to either map CI 3 docs to user_guide on the website, or update all links in source to CI 3 docs.


jim-parry:
Working on a symbolic link for this, so the source doesn't need changing :)

PR #3741, opened 2015-04-06
Separate mobile cache feature

peter4548:
This feature allows the user to set the "separate_mobile_cache" config parameter.  When set to TRUE, CodeIgniter will server separate web page cache files for mobile devices.

To do this, when checking for the cache, a hashtag is added to the cache URI for mobile devices ("#mobile").  This way a separate cache file name will be generated and served for mobile requests.  To detect whether the request is mobile, an instance of CI_User_agent is created, and is later assigned to the controller automatically once it is initialized, whenever the feature is turned on.  This happens before the security classes are initialized, and so we may need to make sure this does not pose a security threat.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -296,6 +296,18 @@ $config['cache_query_string'] = FALSE;
 
 /*
 |--------------------------------------------------------------------------
+| Separate Mobile Cache
+|--------------------------------------------------------------------------
+|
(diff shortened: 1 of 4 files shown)


narfbg:
Sorry, but I don't see more than ~3 people benefiting from this addition, and I really don't think it's a proper solution at all ... this is not the Output class' job.

PR #3742, opened 2015-04-07
[feature] check redis auth failed

ftwbzhao:


narfbg:
You'll have to remove the redis config - it shouldn't be present by default.

Also, we use tabs for indentation, not spaces. Please read our [styleguide](http://www.codeigniter.com/user_guide/general/styleguide.html). A changelog entry should be added as well.

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -302,7 +302,13 @@ protected function _setup_redis()
 
 		if (isset($config['password']))
 		{
-			$this->_redis->auth($config['password']);
+			$success = $this->_redis->auth($config['password']);
```
narfbg: This is redundant ... you can do just `if ( ! $this->redis->auth($config['password']))`

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -302,7 +302,13 @@ protected function _setup_redis()
...
+
+			if ( ! $success)
+			{
+			    log_message('debug', 'Cache: Redis Auth Failed. Check the config.');
```
narfbg: 'Redis Auth Failed. Check the config.' should instead be 'Redis authentication failed.'.

PR #3749, opened 2015-04-08
Updated truncate() and empty_table() to accept an array of fields

gadelat:


narfbg:
No ... emptying a table is something that you should be careful with, not do in batches.

PR #3751, opened 2015-04-08
Logging into custom log file

MUlt1mate:
This feature will allow to separate logging data.
That's useful for searching and viewing logged information.
I'll add docs if you approve the feature.

There is also a security issue: directory traversal.
I think it's better to add warning message to docs rather than sanitize input.


MUlt1mate:
I also broke the indention. 
Maybe somebody familiar with the issue? 
I use PhpStorm and it seems fine there.

narfbg:
I don't really like this idea in the first place and you've pointed out enough flaws for yourself ...

PR #3753, opened 2015-04-09
use = instead of +=

ftwbzhao:


narfbg:
What's the point of this change? "is not rigorous" means nothing ...

ftwbzhao:
@narfbg 

```php
$this->cache->redis->save('zero', 0);
$this->cache->redis->get('zero'); // string(1) "0"

//now
$this->cache->redis->get_metadata('zero'); // return FALSE

//In fact
$this->cache->redis->get_metadata('zero'); // return  string(1) "0"
```

narfbg:
And what has that got to do with configuration?

ftwbzhao:
Two Commit Tow Things
> First
**use = instead of +=**

---

> Second
**[fix] redis get_metadata**

narfbg:
Look, I am not blind ... I see that there are two commits, that's not what I'm asking.

'use = instead of +=' doesn't explain **why** you want to use = instead of +=. Please be descriptive.

ftwbzhao:
$config is empty array (: Why Use += ?

narfbg:
Good point, but next time - explain that.

Remove the changelog entry, it's in the wrong place and it says nothing.

ftwbzhao:
Just Doing Now..

narfbg:
*Remove* it, don't change it.

ftwbzhao:
OK

narfbg:
Just **remove the changelog entry**!

ftwbzhao:
OK Now!

PR #3754, opened 2015-04-09
Update Session.php

plastiik:
Allow "set_userdata" to receive object to set sessions.

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -691,7 +691,7 @@ class CI_Session {
 	 */
 	public function set_userdata($data, $value = NULL)
 	{
-		if (is_array($data))
+		if (is_array($data) || is_object($data))
 		{
 			foreach ($data as $key => &$value)
 			{


narfbg:
Objects shouldn't represent lists ...

plastiik:
They shouldn't... but sometimes, they represent.

This will prevent any error, or no execution to be made (stop the code from being executed), and actually you will not need to "cast" the object to array, increasing performance, with is good, and then as know you save the Session in an array, and not an array.

Sometimes, you do a API request to some, service, and they return a JSON Object with a array of Objects, and if you dont need to save that information on the DB, you can use sessions to use it between pages.

For me it helped because I need to think in performance, and I dont need to cast any vars, its a quick fix, that does not hurt anybody, only prevent an kill of execution.

I did this, because I have a big APP, and I update to Codeigniter 3.0, and its insane to change every place this is happening.

I will leave this in my version at least :+1:

narfbg:
If you try to prevent every "kill of execution" at any cost, you're hiding a lot of potential logic flaws.

plastiik:
True :+1:  Still keep this in my APP :dancer:

PR #3755, opened 2015-04-10
[fix] migration function ``version``

ftwbzhao:
The migration function ``version`` didn't return true if already latest.

```php
<?php
//IF $target_version == 10 AND $current_version == 10
//Run code.
var_dump($this->migration->version(10)); //return string(3) "010"
//fix
var_dump($this->migration->version(10)); //return bool(true)
//fix migration function `latest`,`current` too
?>

### DIFF
--- a/system/libraries/Migration.php
+++ b/system/libraries/Migration.php
@@ -232,6 +232,11 @@ class CI_Migration {
 			return TRUE;
 		}
 
+		if ($target_version == $current_version)
+		{
+			return TRUE;
+		}
+
 		$previous = FALSE;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ftwbzhao** [comment]
When ``$target_version `` == ``$current_version``

```php
#step1
else
{
    // Moving Down, apply in reverse order
    $method = 'down';
    krsort($migrations);
}

//step2：Expression is always false
($method === 'down' && $number <= $current_version && $number > $target_version)

//at last：@return isn't true
return $current_version; 
```

Issue #3757, opened 2015-04-10
OCI8: db->_escape_char should be empty by default in CI 3.0


pabloveliz:
In CI 3.0 a change was introduced in **DB_driver.php**
```
	protected $_escape_char = '"';
```
In previous version wasn't protected, which allows me to workaround on oci8 driver from controllers/models without messing with the system folders.
Also, there are no getters or setters for this variable, and since we cannot extend the database class, and most of the drivers define their own *escapechar* (cubrid, mssql, mysql, mysqli, odbc, sqlsrv, pdo::[4d,cubrid,dblib,mysql,odbc,sqlsrv]). 

I propose to set it to empty in **DB_driver.php** :
```
	protected $_escape_char = '';
```
In this way, OCI8 will work as expected, and any new DB driver won't have to worry about defining this unless neccessary.

2) Or define it in **oci8_driver.php** as empty (like in odbc_driver.php)
```
	protected $_escape_char = '';
```

3) Or, just left unprotected in DB_driver.php as in previous version:
```
	$_escape_char = '"';
 ```

That last option would allow us to set to any char that works with the version we are working on ie. Oracle 10, Oracle 11, etc..

I prefer option 1, but option 3 would work for me and any other person that is working fine with current defaults.

**Comment by narfbg** [comment]
OCI8 doesn't define it because it inherits it from DB_driver, this is intentional.

**Comment by pabloveliz** [comment]
Then it's BUG... 

Using Oracle Instant Client on Oracle Database 10g Enterprise Edition Release 10.2.0.1.0 - Prod

When I run CI 3.0 with oci8 driver (as intentional) I get Error Number: 904
```
ORA-00904: "company": invalid identifier

SELECT "code", "description" FROM "SALES"."EXCHANGE" WHERE "company" = 'GOOG'
```
The problem is the escaping char, which should be empty, so the query would be
```
SELECT code, description FROM SALES.EXCHANGE WHERE company = 'GOOG'
```
In that case, I guess option 2) would be the recommended way, or maybe implement setter and getter for that variable (since it's a protected one).

narfbg:
No, it's not a bug and it's not the quotes that are your problem - it's case-sensitivity.

pabloveliz:
Quite right, the problem was case-sensitivity, so... migration from 2.x to 3.x is giving me problems because there are a lot of lowercase statements that were working with 2.x and the $_escape_char.
Is there a way to ALTER that variable without touching the source in "system"?

[...]

PR #3759, opened 2015-04-10
fix OCI8 driver default escape character, as stated in issue #3757

pabloveliz:
With the default DB escape character, the SQL statement generated by the oci8 driver was producing an **ORA-00904:invalid identifier** error. ie:
```
SELECT "code", "description" FROM "SALES"."EXCHANGE" WHERE "company" = 'GOOG'
```
The problem was that the escaping char was not defined as in the other drivers, which should be empty, so the query would be:
```
SELECT code, description FROM SALES.EXCHANGE WHERE company = 'GOOG'
```


narfbg:
No, it should not be empty and this is not a bug.

PR #3768, opened 2015-04-14
[log] use Constants

ftwbzhao:
```$config['log_file_extension'] ``` remove isset

> ```log_file_extension``` is the same as ```log_path```, so remove isset.


narfbg:
No.

PR #3777, opened 2015-04-15
Fix prefix and suffix bug.

wphp:
This bug occurs when using a custom prefix and suffix in the pagination class, $this->cur_page will always equal to 1.
$config['prefix'] = 'page_';
$config['suffix'] = '/';


ZombieProtectionAgency:
Wouldn't it be better to substring out starting at prefix + 1 going through length - suffix - prefix?

narfbg:
Only security fixes go into the 2.x version tree now, sorry.

PR #3783, opened 2015-04-18
Added the option to set a session CSRF cookie

achraf52:
I have added an option that lets the developer set CSRF cookies to expire at the browser session's expiration, this behavior is widely used by many popular websites and it doesn't affect or diminish the efficiency of CSRF protection, but prevents "unauthorized" errors from throwing up to the average user who left a form open for a few hours before trying to submit it

It can be enabled by setting $config['csrf_expire'] to 0 in the application config.php file.

I hope someone will document this new feature provided by the Security library.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -422,9 +422,14 @@ $config['global_xss_filtering'] = FALSE;
 |
 | 'csrf_token_name' = The token name
 | 'csrf_cookie_name' = The cookie name
-| 'csrf_expire' = The number in seconds the token should expire.
+| 'csrf_expire' = The number of seconds before token should expire.
 | 'csrf_regenerate' = Regenerate token on every submission
 | 'csrf_exclude_uris' = Array of URIs which ignore CSRF checks
+|
+| Options:
+|
+| Set 'csrf_expire' to 0 to mark the cookie as "session-only" which
+| will is a cookie that only expires when the browser is closed.
 */
 $config['csrf_protection'] = FALSE;
 $config['csrf_token_name'] = 'csrf_test_name';
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -258,7 +258,15 @@ class CI_Security {
 	 */
 	public function csrf_set_cookie()
 	{
-		$expire = time() + $this->_csrf_expire;
+		if ($this->_csrf_expire==0)
+			{
+				$expire = 0;
+			}
+			else
+			{
+				$expire = time() + $this->_csrf_expire;
+			}
+		
 		$secure_cookie = (bool) config_item('cookie_secure');
 
 		if ($secure_cookie && ! is_https())
--- a/user_guide_src/source/libraries/security.rst
+++ b/user_guide_src/source/libraries/security.rst
@@ -90,6 +90,20 @@ may alter this behavior by editing the following config parameter
 
 	$config['csrf_regeneration'] = TRUE;
 
+You can set the CSRF cookie to either expire after a set number of
+seconds or as a session-only cookie (set as 0) and the CSRF cookie
+will expire when the browser is closed.
+
+	$config['csrf_expire'] = INT;
+
+.. note:: If you use the session-only option, please also make sure
+you set the 'cookie_httponly' config parameter to TRUE so that an
+attacker can't attack your users using an XSS vulnerability and
+retrieve their CSRF token using JavaScript. If you can't use 'HTTPonly'
+cookies because you're also using a JavaScript framework, it is
+probably better for the user's security that you set the CSRF cookie
+to expire set duration of an hour or two.
+
 Select URIs can be whitelisted from csrf protection (for example API
 endpoints expecting externally POSTed content). You can add these URIs
 by editing the 'csrf_exclude_uris' config parameter::


narfbg:
This has nothing to do with sessions, it just enables the CSRF cookie to last until the browser is closed ...

achraf52:
@narfbg Ah sorry, I meant the Security library, post corrected.

narfbg:
Well, why don't you just set a higher expiry time?

achraf52:
Setting higher expiry time for a CSRF cookie may allow an attacker to use an XSS vulnerability (in target website) to first get the CSRF token stored by the target website with JavaScript, then use the CSRF token to submit a malicious form request as a legitimate one, without the user noticing anything.

narfbg:
I don't see how that's different to setting the expiry to 0, so a user can submit a form hours later ...

achraf52:
Setting the expiry to a "long" amount of time would mean the CSRF cookie would stay in the user's browser EVEN if they have closed the browser tab, thus, while browsing other websites or their email, they might land on a malicious page designed to XSS our website and retrieve the CSRF cookie (the attacker knows we're setting the CSRF cookie for a long enough duration) and your form is exploited.

In my approach, setting the CSRF cookie to session would mean it would get deleted once the browser tab for our website is closed, it is less likely for the user to have their CSRF token exploited if they simply make sure to close the website once they're finished using it, basic security practice.

Look, after all, I'm not forcing CodeIgniter to use session based CSRF cookies, I'm only adding the option to use it because it makes sense and I've found some "major" websites using it.

narfbg:
> In my approach, setting the CSRF cookie to session would mean it would get deleted once the browser tab for our website is closed

Closing the browser *tab* doesn't have any effect - you have to close the whole browser. Firefox acts that way, surely IE does it too and I have my doubts about Chrome being any different in that regard ... have you tested it? So, if you're basing this proposal on the assumption that closing a tab invalidates the cookie, then it is based on a false assumption ...

I understood what you're trying to do even before I asked why don't you just set a higher expiry time, but here's the thing - if you care about the security aspect, you wouldn't allow a browser to suddenly submit a CSRF token for a site that you haven't visited in days, yet this is completely possible if you have an expiry time of 0. I know my browser hasn't been closed for a week now and that's just because I've been travelling recently ... it could stay open for upwards of a month

I'm not against the idea per se, but I'd rather discourage any chance for the CSRF token to be valid for more than a fixed amount of time (that shouldn't be more than an hour). So far I'm just not convinced that the benefits outweigh the disadvantages ...

Btw, as you probably know, the word 'session' means something quite different in the PHP world and it's a bit confusing when you use it ... :)

achraf52:
I agree on the fast that it's closing the browser window that invalidates the cookie, not the tab.
The security CSRF tokens provide doesn't guarantee to you that the user had a fresh start using your website and it also doesn't guarantee to you that the token wasn't tampered with, but CSRF tokens protect your users from getting "hijacked" by another website into submitting a dangerous form request (ex: automatically sending money to the hacker through their online bank).

It is marginally better to have the CSRF cookie expire in a large enough but set amount of time, but the security benefit is negligible, because if the user visits a malicious website and your website has an XSS vulnerability, it won't matter that much whether the CSRF cookie is set to expire in a few hours or on when the browser is closed. They will still be able to retrieve the token and submit the request.

Setting to expiry time to a short amount of time is better but impacts user experience, and good luck explaining to the average guy why you have to invalidate their request if it took them a bit longer to fill the form than you've expected.

It is a battle between security and usability, but if a website is well protected and tested against XSS attacks and vulnerabilities, I think it's safe enough to use 0 as the CSRF cookie expiry value and the benefit is a better user experience.

To counter the slightly decreased security by using a 0 expiry, the developer could use the 'cookie_httponly' config parameter to protect against being "hijacked" through an XSS vulnerability (which uses JavaScript) but if that's not an option (if they're using AngularJS or something similar) then it's probably better for them to set an expiry time instead.

narfbg:
It's not uncommon (nor a significant user-experience issue IMO) that a user is forced to re-submit a form when they've submitted it *long* after it was initially opened. Plus, I'm sure you'll agree that saying that any kind of a security benefit is "neglible" is bound to raise some eyebrows ...

You have a point of course, but as I said - I'm not convinced. In fact, I don't see myself being 100% sure about either of the two possible decisions, so we should get some more opinions ...

@benedmunds @jim-parry What do you guys think?

benedmunds:
I prefer requiring an expiration time.  The dev should set a fairly small window.  It's not abnormal to have to redo a form that you let sit for awhile.

jim-parry:
Cookies should have an expiry time. What I read suggested a maximum of session expiry time plus two hours.

I also read that an expiry of "0", proposed above, effectively asks the browser to delete any such cookie immediately. This does not sound like the intent above.

achraf52:
@benedmunds We are not designing a website, CodeIgniter is a framework that should provide readily-available customization options to the developer, and they're the ones who are responsible to evaluate their requirements and "prefer" to either set an expiration time or use a 0 cookie value.
As long as adding the feature to CodeIgniter is not a security risk, I don't see why it shouldn't be added as a feature CodeIgniter can do, and the slight security risk of leaking the CSRF token can be mitigated by setting the "cookie_httponly" to TRUE which makes it practically impossible for an attacker to obtain the CSRF token via an XSS vulnerability.

@jim-parry What if the user session is set to expire when closing the browser ? Huh.
What you've read is wrong, almost all browsers I know of treat a 0 or non-existent cookie expiry value as an instruction to set a "session" cookie that expires when the browser is closed.
The only situation where a cookie expires immediately is when you set the expiry to a non-zero value lower-than or equal-to the current Unix timestamp.

achraf52:
As I said earlier, we could document in the CodeIgniter manual that an expiry og0 to the CSRF cookie should be used in conjunction with the config parameter "cookie_httponly" set to TRUE or else it exposes a small security risk of having the CSRF token stolen by an attacker via an XSS vulnerability.
If it isn't feasible to use HTTPonly cookies (due to the website using AngularJS or similar JavaScript frameworks) then it would be preferable to set an expiry time for the CSRF cookie, simple as that.

benedmunds:
@achraf52 I'm well aware that we're designing a framework here.  You'd do best to not be condescending in future arguments if you wish to bring people to your viewpoint.  

CodeIgniter gives plenty of flexibility but if there is an option that would allow developers to make a bad decision with little upside I'm not inclined to include it.  There is nothing stopping you from adding this to your personal projects.

Feel free to continue to attempt to convince us otherwise but so far I don't see this as an improvement.

achraf52:
@benedmunds No offense meant, I've just wanted to make a point.

Have you checked my 2nd reply? It explains how the CodeIgniter documentation should instruct developers to deal with this feature.

jim-parry:
@achraf52 You are correct about the "0" - I misread when I was researching.
@narfbg @benedmunds I don't have a problem with this update.
@achraf52 Will you update the comments in application/config/config.php and add one in the user guide ... security.rst?

achraf52:
Everything is alright, I have updated the comments in application/config/config.php and the security.rst user guide file. Please check the 2 changed files and let me know if there's something to fix there.

achraf52:
I think this pull request was dismissed by the project owners so I'm closing it.

PR #3784, opened 2015-04-19
Add a phone link

souri84:
As a mailto link, phone link return a link which looks like tel:0102030405

### DIFF
--- a/system/helpers/url_helper.php
+++ b/system/helpers/url_helper.php
@@ -567,3 +567,30 @@ if ( ! function_exists('redirect'))
 		exit;
 	}
 }
+
+// ------------------------------------------------------------------------
+
+if ( ! function_exists('tel'))
+{
+	/**
+	 * Tel Link
+	 *
+	 * @param	string	the phone
+	 * @param	string	the link title
+	 * @param	mixed	any attributes
+	 * @return	string
+	 */
+	function tel($phone, $title = '', $attributes = '')
+	{
+		$title = (string) $title;
+
+		if ($title === '')
+		{
+			$title = $phone;
+		}
+
+		return '<a href="tel:'.$phone.'"'._stringify_attributes($attributes).'>'.$title.'</a>';
+	}
+}
+
+// ------------------------------------------------------------------------


narfbg:
Seems a bit useless to me (as well as the `mailto()` helper in the first place ...)

@jim-parry What do you think about this?

souri84:
No problem if you think it's not important... It's just an idea.

Cheers,

Thibaut

jim-parry:
Upon digging, this actually makes sense. It is used to trigger a phone call on a mobile device. Similarly, sms: is used to trigger a text message. They are both legitimate (RFC3966 and RFC5724, respectively).

I am fine with both of those ("tel" and "sms") functions being added to the url_helper, and the user guide updated accordingly.

narfbg:
Yea, I know what it does, but is it really useful as opposed to writing the anchor tag yourself? It's not like you'd do any special parsing on it, which may be required for an `http://` link ...

jim-parry:
Good point. The mailto has the corresponding safe_mailto, which does try to do something useful. Would it make any sense to have safe_tel and safe_sms? Is that an issue?

ZombieProtectionAgency:
It might be worth it to consider that some Android phones use special codes entered in the dialer to perform system functions. The only example I know off of the top of my head is dialing *#*#2432546#*#* on a Droid Razr Maxx xt912 causes the phone to phone home to the carrier. These codes don't require the call button to be pressed either. You just type it in and your phone intercepts it.

I don't know if having them automatically populated would trigger the intercept, if so it would probably be a good idea to clean tel links. The only problem is the codes vary a lot and I don't know if there is a definite syntax.

drivingmenuts:
Seems like the appearance of a phone's special function # appearing on a web page in the browser would NOT trigger the phone to perform that special function without the user at least pressing the link. A phone number appearing on a browser page on a smartphone doesn't cause the phone to automatically dial the number - the user still has to press the link.

narfbg:
@jim-parry I don't think there's anything to make "safer" in a phone number.

PR #3785, opened 2015-04-19
404_override route is now being parsed correctly, including controller method parameters.

birdayz:
Controller method parameters have not been parsed at all when it comes to the 404_override route.
$route['404_override'] = 'error/display/400';
When using an 404_override route like above, the '400' parameter is NOT passed to the controller method. Fixed that. For the future, it could be nice if we alter the route/URI in a cleaner way, the whole 404 handling code i saw feels like a hack.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -430,9 +430,23 @@ if ( ! is_php('5.4'))
 	{
 		if ( ! empty($RTR->routes['404_override']))
 		{
-			if (sscanf($RTR->routes['404_override'], '%[^/]/%s', $error_class, $error_method) !== 2)
+			$tokens = explode('/', trim($RTR->routes['404_override'], '/'));
+
+			$error_class = $tokens[0];
+
+			if (sizeof($tokens) >= 2)
+			{
+				$error_method = $tokens[1];
+			}
+			else
 			{
 				$error_method = 'index';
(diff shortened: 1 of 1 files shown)


narfbg:
It was never supposed to parse parameters.

They are *input* parameters, incoming from client requests. You, being on the server side, don't need to regard them as variables.

birdayz:
Well these input parameters do come from my 404_override route, it's not from the client.
I think it should be possible to pass a parameter to the controller, which is called for the 404_override.

narfbg:
Well, if you think that it should be possible, why do you need to do that?

birdayz:
I have a Controller "error" with a function "display". it takes a number as parameter, and displays stuff accordingly (i.e. it gets 404 in my case by using404_override error/display/404, sorry the example with '400' made no sense)

I admit that i'm quite new to CodeIgniter and this might be a wrong way to use CI.

narfbg:
You can always just make 404 the default value:

    public function display($code = 404)

Or have a method designed specifically for 404 error handling:

    // $route['404_override'] = 'error/error_404';
    public function error_404()
    {
         return $this->display(404);
    }

birdayz:
Yeah you're right a specific method for the 404, calling the "other" method would be a good solution as well. Thanks for the feedback.

PR #3788, opened 2015-04-20
Prevent double escaping of HTML entities in form helper attribute parsin...

toxonics:
Added FALSE as the second parameter to html_escape in _parse_form_attributes to prevent HTML entities being double encoded ie. &quot; being encoded to &amp;quot;

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -907,7 +907,7 @@ if ( ! function_exists('_parse_form_attributes'))
 		{
 			if ($key === 'value')
 			{
-				$val = html_escape($val);
+				$val = html_escape($val, FALSE);
 			}
 			elseif ($key === 'name' && ! strlen($default['name']))
 			{


narfbg:
If this prevents *double* escaping, where does the first escaping occur?

toxonics:
When using the form_input helper as
`<?php echo form_input('name', set_value('name', $item->name)); ?>`
and `$item->name = "Display 14.1" WXGA+ 1440x900";` it first escapes this to `Display 14.1&quot; WXGA+ 1440x900` then on the next update it escapes it to `Display 14.1&amp;quot; WXGA+ 1440x900` then the next one is `Display 14.1&amp;amp;quot; WXGA+ 1440x900` etc. adding FALSE makes it only change the text to `Display 14.1&quot; WXGA+ 1440x900` and it stays that way.

narfbg:
`set_value()` has a parameter to disable escaping in itself, use that instead ...

toxonics:
Ok

bright-light-in-the-night:
I think this pull request should be merged into CI.

1.  your solution might not be what you want:  

> set_value() has a parameter to disable escaping in itself, use that instead ...

Isn't this enabling XSS when a form is submitted?
e.g: paste this `"><script>alert(1)</script>` into an input and submit the form.

2. why it should be merged? because when I retrieve the value from the DB, it's already escaped, thus it will double escape when echoing.

`<input value="<?php echo set_value('name', $item->name); ?>">`

This is how I save the data into the DB

`$this->db->set('name', set_value('name'));`

toxonics:
I don't use CI any more and the branch is ancient, but if you want to try and get this merged feel free

narfbg:
@bright-light-in-the-night Yes, but this issue is about preventing *double* escaping, meaning that the input was already escaped and is thus safe.

PR #3790, opened 2015-04-21
Update Xmlrpc.php

misterspelik:
Add cookie field to the Client class to be able set cookie for server.

### DIFF
--- a/system/libraries/Xmlrpc.php
+++ b/system/libraries/Xmlrpc.php
@@ -598,7 +598,7 @@ class XML_RPC_Client extends CI_Xmlrpc
 	 * @var	string
 	 */
 	public $password;
-
+	
 	/**
 	 * Proxy hostname
 	 *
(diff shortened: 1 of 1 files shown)


jim-parry:
I can see that cookie-based validation is supported for XML-RPC. Where/how would you get the token to put in the cookie?

misterspelik:
Hi! Token will be get from response headers after login. We can use Set-Cookie header and transfer it`s data to $this->xmlprc->client->cookie in controller (just as example) .

narfbg:
@misterspelik I believe @jim-parry's question was rather what the API for setting ``XML_RPC_Client::$cookie`` would be. As it stands, you're just adding a public property which is empty by default and the condition to use it if it's not empty ...

Typically, we would discourage direct assignment of values to library properties and even disallow it in most cases. Even if you believe that this is the right approach, it should at least be documented. Right now, only somebody looking at the commit messages would know about this feature's addition (eventually).

PR #3792, opened 2015-04-22
added ability to redirect() function to redirect back

azeemhassni:
Hi , 
i added ability to redirect() function to redirect to the previous page.
useful when working with forms.

Kindly merge this pull request.

### DIFF
--- a/system/helpers/url_helper.php
+++ b/system/helpers/url_helper.php
@@ -531,6 +531,15 @@ if ( ! function_exists('redirect'))
 	 */
 	function redirect($uri = '', $method = 'auto', $code = NULL)
 	{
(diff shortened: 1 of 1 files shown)


narfbg:
How's that better than just outputting a link to the previous page?

azeemhassni:
@narfbg you don't have to update the redirect function argument if you changed the method name anyways.

narfbg:
OK, to rephrase ... how's it better than outputting a link to `$_SERVER['HTTP_REFERER']`?

On another note, the 'referer' value is provided by a request header (meaning it comes from the client) and it may not exist, in which case we'd have a broken feature ... which is why I can't merge this.

PR #3793, opened 2015-04-22
User agent - set all

mkona:
set all three to make sure we know when crawlers like googlebot are testing mobile experience.

Otherwise, tools like this would not test on mobile version
https://www.google.com/webmasters/tools/mobile-friendly/


narfbg:
`_set_robot()` does that independently.

PR #3796, opened 2015-04-22
Trusted Hosts

ikarius6:
In case you want to leave the base_url configuration empty for portability, it involves a risk:
http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

To avoid security issues, you may want to configure a list of "trusted hosts".
(for dev and test environments or multi-domain sites)

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -21,6 +21,28 @@ $config['base_url'] = '';
 
 /*
 |--------------------------------------------------------------------------
+| Trusted Hosts
+|--------------------------------------------------------------------------
+|
+| In case you want to leave the base_url configuration empty for portability,
+| it involves a risk:
+|
+| http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html	
+|
+| To avoid security issues, you may want to configure a list of "trusted hosts".
+| (for dev and test environments or multi-domain sites)
+|
+| Elements of this list imply you accept any subdomain it has. 
+| For example domain.com also accepts <whatever.>domain.com
+|
+| Array:		array('localhost', 'my-development.com', 'my-production.com', 'subdomain.domain.com')
+|
+*/
+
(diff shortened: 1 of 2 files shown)


Tan5en5:
I don't think it's relevant because you just have to set base_url in your config file to avoid this. You can even have several config files by environment.

ikarius6:
Multidomain

narfbg:
There are many ways to do this without it being in CI itself.

In fact, it can be as easy as `$config['base_url'] = 'https://'.$_SERVER['SERVER_NAME'].'/';`. Or you can do a simple `in_array($_SERVER['HTTP_HOST'], $your_list_of_hosts, TRUE)` in the config file itself. Or you can set an environment variable based on the domain in use ...

Bottom line is, this is not a very common use case and I don't see a benefit in introducing yet another way to do it.

ikarius6:
The `$config['base_url'] = 'https://'.$_SERVER['SERVER_NAME'].'/';` don't allow a multidomain build, and yes, it can be resolved with `in_array` but I think than **CodeIgniter must to provide the security in his default code**, however I've made an extra for this:

https://github.com/ikarius6/codeigniter-trusted-hosts

Thank you for attention. Keep working hard for a better Internet.

narfbg:
Yes it does, you only need to point your virtual hosts to the same directory.

ikarius6:
Of course, how you can edit the vhost "easier" than a config file... just saying, in the industry of web development is better to put the available options over the table than make the life easy.

Is like the case of your "Rewrite PHP Short Tags" config option, you can even modify the php.ini but is better (for portability and installation friendly) have the option on config.

Cheers.

narfbg:
"Rewrite PHP Short Tags" is a horrible feature, and you should really stop telling me what's better and what not.

PR #3797, opened 2015-04-23
Change message to show the file name in the error

avenirer:
When doing multiple file uploads, the user won't know what file is the error related to. This way the upload library will give more information in the errors.


narfbg:
The wording feels a bit off and there's a potential security issue (similar to the one that triggered the 2.2.2 release) in simply outputting a user-submitted file name ... I'd rather not have it that way by default, you can always configure your own lang lines.

avenirer:
Actually is not a security issue, it only outputs the file name prepared by the library. Regarding the wording, it can at any time be changed by any pull request :)

narfbg:
That "prepared by the library" filename is often just the user-submitted filename as is ...

avenirer:
Of course is the user-submitted filename. That is what the user needs to know when an error appears. Anyway, the only point at which we could talk about security issues would be in the _prep_filename() method. So, if something bad happens it can only happen if it passes that function, with or without the lines I wrote in this PR.

PR #3800, opened 2015-04-23
Cache Library Clean-up

refringe:
Made a few small changes to the cache library:

- Renames the `$_adapter` variable to `$_driver` because we don't call it an adapter anywhere else.
- Renames the `$_backup_driver` variable to `$_backup` for easier initializing, and because it matches the configuration key passed in.
- Validates the primary driver **and** the backup driver against the `$valid_drivers` array.
- No longer attempts to set an undefined `$_memcached` class variable.
- Renames the `$key` variable to `$driver_type` to be more descriptive.

I've done some light testing, but it would be great if someone else could try it out as well.

### DIFF
--- a/system/libraries/Cache/Cache.php
+++ b/system/libraries/Cache/Cache.php
@@ -100,27 +100,18 @@ class CI_Cache extends CI_Driver_Library {
 	 */
 	public function __construct($config = array())
 	{
-		$default_config = array(
-			'adapter',
-			'memcached'
-		);
-
-		foreach ($default_config as $key)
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but 'adapter' to 'driver' is a change for the sake of change ... and actually breaks the code because the library uses `$this->_driver`, but still accepts 'adapter' for configuration and assigns it to `$this->_adapter` ... If better wording was your goal, then 'backup_driver' to 'backup' is also deffinately not an improvement.

The only thing I'd change is to remove that foreach loop together with the 'memcached' thing ...

PR #3804, opened 2015-04-24
Cache Library Clean-up

refringe:
Continuation of #3800. I accidentally deleted the branch and GitHub won't update the old PR anymore.

This reverts the previous variable renaming and removes the foreach loop in favour of an if condition.

### DIFF
--- a/system/libraries/Cache/Cache.php
+++ b/system/libraries/Cache/Cache.php
@@ -100,27 +100,18 @@ class CI_Cache extends CI_Driver_Library {
 	 */
 	public function __construct($config = array())
 	{
-		$default_config = array(
-			'adapter',
-			'memcached'
-		);
-
-		foreach ($default_config as $key)
+		// Loop through the available driver types and overwrite them as they're found.
+		foreach (array('adapter', 'backup') as $driver_type)
 		{
-			if (isset($config[$key]))
+			if (isset($config[$driver_type]) && in_array($config[$driver_type], $this->valid_drivers))
 			{
-				$param = '_'.$key;
-
-				$this->{$param} = $config[$key];
+				$option = '_'.$driver_type;
+				$this->{$option} = $config[$driver_type];
 			}
 		}
 
+		// Overwrite the default key prefix.
 		isset($config['key_prefix']) && $this->key_prefix = $config['key_prefix'];
-
-		if (isset($config['backup']) && in_array($config['backup'], $this->valid_drivers))
-		{
-			$this->_backup = $config['backup'];
-		}
 
 		// If the specified adapter isn't available, check the backup.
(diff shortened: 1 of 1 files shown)


system/libraries/Cache/Cache.php
```diff
@@ -100,28 +100,18 @@ class CI_Cache extends CI_Driver_Library {
 	 */
 	public function __construct($config = array())
 	{
-		$default_config = array(
-			'adapter',
-			'memcached'
-		);
-
-		foreach ($default_config as $key)
+		if (isset($config['adapter']) && in_array($config['adapter'], $this->valid_drivers))
```
narfbg: This will silently ignore any invalid driver and keep the library working with the 'dummy' one.

refringe: Good point. I've added an "error" level log_message call for invalid adapters, both primary and back-up. Let me know if you had something else in mind.

narfbg: I did ... you shouldn't have added the `in_array()` condition in the first place. Validation of the driver is done later.

refringe: It was already there for the back-up adapter, I just added it for the primary adapter as well.

> Validation of the driver is done later.

I assume by that you mean the `is_supported` function? That doesn't currently validate that the user has set a valid adapter, just that the adapter that's set is usable; is memcached actually installed?

Taking the `in_array()` checks against the `$valid_drivers` array out would allow a user to set the adapter to one that doesn't exist causing an error when the `is_supported` driver function is called on [L266](https://github.com/refringe/CodeIgniter/blob/feature/cache-defaults/system/libraries/Cache/Cache.php#L266).

What are your thoughts on scrapping the `$valid_drivers` array and rewriting the `is_supported` function to fail gracefully if the adapter doesn't exist? It's simple and arguably easier to add a custom written driver.

narfbg: The conditions that you linked in your last paragraph *are* what makes the validity checks ... the logic that you're now introducing already exists in there - that's what I meant with my previous comment.

As for the `$valid_drivers` array ... maybe, I'm not really happy with it being user-configurable.

refringe:
I'm still getting accustomed to the CI core... I can see now that the driver validation is already happening (for both primary and back-up drivers) in the [system Driver library](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Driver.php#L106).

This pull:
- Removes the for loop that attempted to set an invalid `_memcached` variable.
- Removes the extra validation on the back-up driver.

Let me know if there's anything else needed.

refringe:
@narfbg Any idea why the build is failing?

narfbg:
Unrelated ... Travis is having some issues with MySQL.

PR #3813, opened 2015-04-28
2.2 stable

DevotionGeo:


taggedboy:
I think this one is the addition of nice validation rules. Even I needed alpha with spaces months ago.

DevotionGeo:
Dear Andrey Andreev, I think the form validation rules like alpha_spaces and alpha_numeric_spaces are needed.
Every time I need to add a custom validation rule like this.

public function fullname_check($str) {
    if (! preg_match("/^([a-z0-9 ])+$/i", $str)) {
        $this->form_validation->set_message('fullname_check', 'The %s field can only be alpha numeric');
        return FALSE;
    } else {
        return TRUE;
    }


This merge will solve the problem.

ZombieProtectionAgency:
@DevotionGeo I am pretty sure narfbg has said that he wont be merging any non-security patches into CI 2 anymore

DevotionGeo:
Oh ! That's great, it means they are encouraging the use of v 3. Thank You Andrey Andreev for the great work !

PR #3814, opened 2015-04-28
Add alpha_spaces validation rule to the form_validation library.

DevotionGeo:


DevotionGeo:
The error occurred because of the change in .rst file because I tried to add the documentation, but changed the document back to the original (deleting the line I added) because I wasn't comfortable with " Sphinx " and was afraid that I'll break it.

DevotionGeo:
I closed this request myself, because it wasn't tidy, I put another request.

PR #3815, opened 2015-04-28
Add full_names rule to the form_validation library.

DevotionGeo:
rule " alpha_numeric_spaces " is already there, but I think the rule " alpha_spaces " is also important to be included in the core, so that users don't need to create custom rules for a simple thing.

Sometimes input fields which have alphabets and spaces but not numbers are required, for example "Full Name" text field where user can enter two or more names (First and Last names or First Middle and Last names) something like "Arif Hussain" or "Arif Hussain Bangash". 

Note: Apostophies and dashes are added to the rule to pass the names like "Alfered D'Mello", "Simon Von'Corrie Alt-Camero" etc. The rule is renamed to full_names to reflect its use case.
Thank you !


narfbg:
If people's names are your use case, they may also have single quotes or dashes in them ...

Also, documentation updates are required, you don't even have to know that it is based on sphinx ... the syntax is super-easy and you could really just take another validation rule as an example.

DevotionGeo:
Yeah, you're right, I must consider single quotes and dashes in them. I'll get back to it and will recreate this rule + documentation.

DevotionGeo:
Dear Andrey Andreev, What if I name this rule as full_names instead of alpha_spaces ?

It will allow strings like "Andrey Andreev", "Arif Hussain Bangash", "Alfered D'Mello", "Simon Von'Corrie Alt-Camero" etc.

Tpojka:
This rule should contain dots too. I.e. `J. K. Rowling`.

DevotionGeo:
Dear @narfbg what do you think about @Tpojka 's suggestion ?

Tpojka:
I've just meant that is nut unusual people use middle name initial etc. And also it is meant just regarding full name expectation.

mwhitneysdsu:
I think once the name of the rule is full_names, you're pretty much out of luck on validation. About the only thing you can do to validate a full name is to make sure the content is printable and isn't going to break your site.

DevotionGeo:
Yes @mwhitneysdsu , that's also a good point !

narfbg:
So now we're getting at why no such rule exists so far in the first place - too many possible conditions ... What if I wanted to use non-Latin letters?

Sorry, not happening.

DevotionGeo:
Dear @narfbg , I think non-Latin letters and all those other signs are sometimes used in Usernames, not in Full Names, in full names people only use apostrophes and dashes at the most !

mwhitneysdsu:
Most of the people on this planet do not use English when naming their children, which makes it much more likely that non-Latin letters come into play. While many people will have some version of their name which they will use for websites with strict limitations on the allowed characters, most will not be happy if they can't use accents, umlauts, circumflex, etc. (and that's just the western Europeans). We'll not even go into the people who name themselves.

In my opinion, there is far more room to limit usernames than full names, but I don't really think the framework is the place to determine that, especially when it doesn't include authentication.

PR #3817, opened 2015-04-28
Fix sqlite list_fields [pdo_sqlite_driver]

morkid:
I tried to migrate my CodeIgniter version to 3.0 with SQLite database. But some code for example 
`$this->db->list_fields("tablename")` returns the output like this 
```
This feature is not available for the database you are using.
...
```
I found `_list_columns` method return false only, which means not supported. I added a bit of code on `_list_columns` method of `CI_DB_pdo_sqlite_driver` class and it worked. 
```php
// Filename : system/database/drivers/pdo/subdrivers/pdo_sqlite_driver.php

protected function _list_columns($table = '')
{
		if ($table === "")
		{
			return FALSE;
		}

		$field_data = $this->field_data($table);
		
		$fields = array();
		foreach($field_data as $id => $col)
		{
			$fields[] = 'SELECT '.$id.' AS "COLUMN_ID", \''.$col->name.'\' AS "COLUMN_NAME"';
		}

		if (!empty($fields))
		{
			return 'SELECT "COLUMN_NAME" FROM ('.implode(' UNION ',$fields).') AS "LIST_COLUMNS" ORDER BY "COLUMN_ID" ASC';
		}
		// Not supported
		return FALSE;
}
```

### DIFF
--- a/system/database/drivers/pdo/subdrivers/pdo_sqlite_driver.php
+++ b/system/database/drivers/pdo/subdrivers/pdo_sqlite_driver.php
@@ -130,6 +130,23 @@ class CI_DB_pdo_sqlite_driver extends CI_DB_pdo_driver {
 	 */
 	protected function _list_columns($table = '')
 	{
+		if ($table === "")
+		{
+			return FALSE;
+		}
+		
+		$field_data = $this->field_data($table);
+		
+		$fields = array();
+		foreach($field_data as $id => $col)
+		{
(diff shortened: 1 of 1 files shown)


narfbg:
I wasn't very keen on running a query (via `field_data()`) in order to generate another query, so I came up with another solution: 0f8509025d2338dc069ddf42b6ade974a5da0ca3

Thanks for the effort though.

PR #3820, opened 2015-04-29
update Migration Class Reference 

ftwbzhao:
http://www.codeigniter.com/userguide3/libraries/migration.html#class-reference

### DIFF
--- a/system/libraries/Migration.php
+++ b/system/libraries/Migration.php
@@ -191,7 +191,7 @@ class CI_Migration {
 	 * choice
 	 *
 	 * @param	string	$target_version	Target schema version
-	 * @return	mixed	TRUE if already latest, FALSE if failed, string if upgraded
+	 * @return	mixed	TRUE if no migrations are found, current version string on success, FALSE on failure
 	 */
 	public function version($target_version)
 	{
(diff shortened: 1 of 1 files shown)


ftwbzhao:
When ``$target_version`` == ``$current_version``

Run
 ```php
var_dump($this->migration->version(5));
```

Expected

```php
string(3) "005"
```

Actual
```php
int(5)
```

ftwbzhao:
Class reference is ambiguity.

### userguide

---

http://www.codeigniter.com/userguide3/libraries/migration.html#class-reference

### code

---

```php
 * @return	mixed	TRUE if already latest, FALSE if failed, string if upgraded
```

narfbg:
Sorry, but your comments are way more ambiguous ... I understand that you barely know English, but it doesn't seem like you're even trying. Even if there is an issue here, I'd have to go through all of it on my own to figure out what it is.

ftwbzhao:
just change 'Class reference'

narfbg:
... that's the kind of comment that says literally nothing.

ftwbzhao:
Sorry, I thought you knew all about.

now, You just gotta tell me

function ``version`` return type is what? 

### In UserGuide Says
---
 TRUE if no migrations are found, current version string on success, FALSE on failure


### In Code Says 
---
 TRUE if already latest, FALSE if failed, string if upgraded

ftwbzhao:
still ambiguous? If you do not understand, please tell me.

:rose

PR #3824, opened 2015-04-30
Fix typos in user guide.

jim-parry:
Corrected "csrf_regenerate" in libraries/security.rst (it read "csrf_regeneration", incorrect per source code).
Corrected "blog" as default controller in general/controllers.rst (it read "Blog", not per convention).


narfbg:
Just FYI, `$route['default_controller'] = 'blog';` works fine ... the convention is about file naming, `ucfirst()` is applied to code/strings automatically.

jim-parry:
I am not suggesting otherwise, just that the capitalization wasn't following the convention, and this was pointed out to me  :-/

narfbg:
Well, it doesn't *have* to follow it ... :)

PR #3828, opened 2015-05-03
Bug Fix manually delete caching method and improving Cache Query String behaviour

w0den:
According to documentation, to manually delete cache for page `/foo/bar` we should run `$this->output->delete_cache('/foo/bar')`, but in this case MD5 hash will be calculated for `http://site.com//foo/bar` and this is why, we should pass `$uri` without beginning slash (ie, `foo/bar`). But the problem is that there is no way to delete cache for home page because:
1) `$this->output->delete_cache('/')` — MD5 hash will be calculated for `http://site.com//` and cache file will not be deleted.
2) `$this->output->delete_cache('')` — MD5 hash will be calculated for `http://site.com/%CURRENT_PAGE%` and again, cache file will not be deleted.

Trimming the beginning slash, we enable ability to delete cache for home page by calling `$this->output->delete_cache('/')`. Also, this method will work as specified in the documentation.

### DIFF
--- a/system/core/Output.php
+++ b/system/core/Output.php
@@ -556,9 +556,16 @@ class CI_Output {
 			.$CI->config->item('index_page')
 			.$CI->uri->uri_string();
 
-		if ($CI->config->item('cache_query_string') && ! empty($_SERVER['QUERY_STRING']))
+		if (($cache_query_string = $CI->config->item('cache_query_string')) && !empty($_SERVER['QUERY_STRING']))
 		{
-			$uri .= '?'.$_SERVER['QUERY_STRING'];
+			if (is_array($cache_query_string))
+			{
+				$uri .= '?'.http_build_query(array_intersect_key($_GET, array_flip($cache_query_string)));
+			}
+			else
+			{
+				$uri .= '?'.$_SERVER['QUERY_STRING'];
+			}
 		}
 
 		$cache_path .= md5($uri);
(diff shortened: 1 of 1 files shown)


lonnieezell:
+1

narfbg:
What if 'index_page' is not empty?

w0den:
For example, if `index_page = index.php` and we want to delete cache for home page using `$this->output->delete_cache('/')`, at this moment MD5 hash will be calculated for `http://site.com/index.php/` (which is wrong). By adding `ltrim($uri, '/')` we delete beginning slash for `$uri` and will pass `http://site.com/index.php` to `md5()` which will return the right hash.

narfbg:
I don't see how '/index.php/' is wrong, but that wasn't my point - `delete_cache('/foo')` with your changes would result in a calculation for '/index.phpfoo', which I'm sure you'll agree is wrong.

w0den:
Look at method `_write_cache()`, line 555:

```php
$uri = $CI->config->item('base_url')
	.$CI->config->item('index_page')
	.$CI->uri->uri_string();
```

if `index_page = index.php` accessing `http://site.com/index.php/foo` cache filename will be calculated for `md5(http://site.com/index.phpfoo)`. If we want to remove cache for this page, at this moment, we should call `$this->output->delete_cache('foo')`.

About `/index.php/` — this is wrong, because when access `http://site.com/index.php` or `http://site.com/` cache filename will be calculated for `md5(http://site.com/index.php)`. But when we want to remove cache for home page, this method will always fail, because:
* If we call `$this->output->delete_cache('/')`, hash will be calculated as `md5(http://site.com/index.php/)` and will fail.

* If we call `$this->output->delete_cache('')`, at line 555 exists `if (empty($uri))` and hash will be calculated as `md5(http://site.com/index.php/{$CI->uri->uri_string()})` (ie, for current page, but not for home page) and again will fail.

So, as I said, there is no way to remove cache for home page.

Of course, we can also modify entire method, by using:
```php
// Instead of `$uri = ''` we should use `$uri = false`
public function delete_cache($uri = false) {
	// and instead of if (empty($uri)) at line 728 use:
	if ($uri === false) {

```

But in this case also, we should modify documentation because, to delete cache, we always should call this method without beginning slash (ex. `$this->output->delete_cache('foo/bar')` for `/foo/bar` and `$this->output->delete_cache('')` for home page).

narfbg:
I've added the necessary inline documentation and changelog entries on top of your patch.

However, for future reference, please don't add features on top of bug fixes (or 2+ features, or 2+ bug fixes, etc.) in the same pull request. It is very rarely the case like this one when we actually like both of them and usually we'd just make you revert one of the changesets ...

PR #3831, opened 2015-05-04
GitAttributes file to clean up composer installs

lonnieezell:
This basically just removes a number of development-only files (like the user guide source and our tests) from any installs made through composer with the --prefer-dist flag set. 

Helps create a faster install with less unnecessary files to new projects.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -18,3 +18,5 @@ readme.rst
 tests/codeigniter/ export-ignore
 tests/travis/ export-ignore
 
+# User Guide Source Files
+user_guide_src
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,20 @@
+# This file tells which files and folders should be ignored and
+# NOT downloaded when using composer to pull down a project with
+# the --prefer-dist option selected. Used to remove development
+# specific files so user has a clean download.
+
+# git files
+.gitattributes export-ignore
+# .gitignore
+
+# helper config files
+.travis.yml export-ignore
+phpdoc.dist.xml export-ignore
+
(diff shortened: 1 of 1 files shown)


narfbg:
Not sure about readme.rst and tests/codeigniter/ ... the whole tests/ directory is kinda useless without the latter.

Speaking of directories ... can we not use the word "folders" please? :)

lonnieezell:
My main reason for removing the /codeigniter folder and leaving the rest is that I thought the mocks **directory** :) would be helpful for people developing their applications to have, but they didn't necessarily need all of our tests in their application. 

And I don't have a strong opinion one way or the other about the readme file so I'm fine keeping that. 

Let me know final consensus on the tests (which I've just been reading through comments when Symfony went through this a while back and the opinions are definitely mixed) and I'll fix it up.

narfbg:
Oh, I thought the readme files was also ignored ... should look more carefully.

Still unsure about the tests though.

lonnieezell:
As for tests - I guess it comes down to use-case. When we pull down a copy of the repo to start a new project with, do we need CI's tests? On one hand, it can be helpful to look at the existing tests for a reference, but you're not typically going to want to run CI's test suite whenever you run your own tests. 

If people want to ensure they get the tests, which I think would be in the minority, then they can always do a git clone, or use `--prefer-source`  with a composer install.

Obviously, I don't think they need to be in there but don't have a strong opinion either way. I can always add some docs that specify how to do both? 

@jim-parry @benedmunds What do you guys think on this one?

benedmunds:
I'd vote for not including the tests.

PR #3833, opened 2015-05-06
remove references

ftwbzhao:
return ``$result`` not ``$array``, 
so  ``foreach ($array as &$a)`` remove references


narfbg:
Don't submit pull requests for the sake of it ...

PR #3836, opened 2015-05-06
fix/cache/redis

ftwbzhao:
http://redis.io/commands/decr
http://redis.io/commands/decrby


narfbg:
https://github.com/phpredis/phpredis#incr-incrby
https://github.com/phpredis/phpredis#decr-decrby

ftwbzhao:
Can describe in detail？
Why not use ``incrBy`` and ``decrBy``

### incr, incrBy
---

_**Description**_: Increment the number stored at key by one. If the second argument is filled, it will be used as the integer value of the increment.

Description: Increment the number stored at key by one. If the second argument is filled, it will be used as the integer value of the increment.

##### *Parameters*

*key*  
*value*: value that will be added to key (only for incrBy)


 ---
function ``incr`` is support for parameters ``offset`` ?

narfbg:
Good point, `incr()` doesn't seem to support the offset parameter. However, it's also unclear what happens when you use `incrBy()` on a non-existing key.

ftwbzhao:
on a non-existing key ?

key means
``$id`` or ``$offset`` ?

Can describe in detail？

PR #3840, opened 2015-05-07
KISS principle

ftwbzhao:
This part of the code could be nicer. But your ``changelog`` still obtains. 

meet sb half way. :+1:


narfbg:
You know, I would've guessed that `incr()` is atomic, while `incrBy()` isn't, but when you're that insistent and when you start using principle names as your arguments, I get really annoyed. So I did what *you* should've done in the first place and looked at what phpredis actually does.

Guess what? `incr()` and `incrBy()` do exactly the same thing, only the offset/value parameter is optional for `incr()`.

https://github.com/phpredis/phpredis/blob/develop/redis.c#L1371

So nothing should've been changed in the first place and you did nothing but to waste time for both of us. For the last time - stop submitting pull requests if you don't know what you're doing.

ftwbzhao:
sorry. No next time.

ftwbzhao:
your actions inspire me.  be strict in submitting pull requests.

Thanks.

PR #3841, opened 2015-05-07
Remove spaces...

vkeranov:


narfbg:
Most of these are irrelevant, change for the sake of change; and spaces are the better choice in some of them ...

Please limit such patches to files under system/ only.

PR #3847, opened 2015-05-10
Pull request for the system/libraries/Profiler.php

paralogizing:
In this pull request I have separated the Profiler presentation from the logic by creating a profiler_template.php View for the presentation, and the Profiler Library just creates and passes along the variables to it.  The only thing I have added is a protected $_profiler_view property to change the template if desired.  Otherwise, it looks and behaves exactly the same.  This code was highly inspired by Lonnie Ezell, and the work he has done at Bonfire.  It is genius.  I'm surprised he hasn't submitted this himself.


narfbg:
And what's the point really? The resulting view isn't any nicer to work with - it's spaghetti code without context, which is at least the one thing you have when it is inside the library.

paralogizing:
The point is to be able to get rid of all the spaghetti code, and allow others to improve on the looks and functionality of it.  As it stands, it is impossible to work on without hacking the core.  This pull request makes the information accessible so that it can be presented however one pleases by either modifying the default template, or overriding it with their own.

ZombieProtectionAgency:
Since there are specified sections and a lot of checks wouldnt it make sense to just add a config file for this and set the html that wraps each section? If you want to make it a template file like that, wouldnt it make more sense to use phps alternative syntax for control structures and just close/reopen php tags where needed instead of concatenating the entire string and echoing it at the end? It still looks like crud but I think it is a lot nicer looking then a bunch of concatenated html

paralogizing:
In a config file you wouldn't be able to loop through arrays would you?  As for the alternative syntax, that would work too.  But in the library it was all placed into strings and concatenated together at the end, so I was just trying to stay as true to the original as possible to make a one to one comparison more feasible.  I've got no bones about it looking like crud, but at least now the meat is separated from the carcass.

ZombieProtectionAgency:
You wouldnt include the loops in the config file. You just include the HTML Wrappers.

I used something like this for a nav control in VB once...

I set `MenuItemTemplate` to `<a>{{text}}</a>` and `MenuTemplate` to `<div>{{menu_items}}</div>`.

Then I create a string called `GeneratedMenuItems`, loop through all menu items, replace `{{text}}` in `MenuItemTemplate` with the anchor text I wanted, and append the new string to `GeneratedMenuItems`. Then I replaced `{{menu_items}}` in `MenuTemplate` with `GeneratedMenuItems`.

paralogizing:
Well, I guess that's another way to do it then.

ZombieProtectionAgency:
@paralogizing I updated my example to use code from this library so it actually makes a little sense, sorry about babbling a little

ZombieProtectionAgency:
What about something like this? Obviously it would need a lot of work but, I couldnt spend too much time on it, but it works.

My config file
http://pastebin.com/eC5tJhT7

Stuff I added/updated in system/libraries/Profiler.php
http://pastebin.com/TQfeEbMG

The `_compile_benchmarks` function returns 

    array(
    	'title' => 'BENCHMARKS',
    	'items' => array(
    		array(
    			'key' => 'Loading Time: Base Classes',
    			'value' => '0.0850',
    		),
    		array(
    			'key' => 'Controller Execution Time ( Welcome / Index )',
    			'value' => '0.0090',
    		),
    		array(
    			'key' => 'Total Execution Time',
    			'value' => '0.0940',
    		)
    	)
    );

Using the above data as an example... 

The `run` function loops over that data. 
The first key is a string so it grabs this sections base template and replaces the matching key, `{{title}}`, with the value "BENCHMARKS". 
Next it finds items which is an array. it finds the items key in this sections template array, loops over each key value pair, substitutes those values in the sections "items" template, and appends the new section to `$sections_output`. 
After all sections have been processed it replaces `{{profiler_data}}` in `$config['profiler_template']` with `$sections_output`

paralogizing:
Does this mean you have something against views?  That's fine, but this could be better implemented by creating a MY_Profiler.php and rewriting the run method to do whatever you like (in my opinion).  If this is how the CodeIgniter gurus-that-be what to approach it then great.  All I am saying is that the presentation (html) should be separated from the logic (arrays).

ZombieProtectionAgency:
Oh no, sorry! I am definitely NOT a CI guru lol

I was just trying to offer my input because I really want to help around here. I thought it offered a nice level of separation from view and code like narf would like and it changed the output of the _compile_ functions from html to objects like you said you wanted.

paralogizing:
Yes, it looks awesome!  You did a really good job.  I would happily swap views with you.  What do you think narf?  I guess until he speaks we'll never know.

narfbg:
@ZombieProtectionAgency's pastebin code is certainly an improvement over what you're suggesting.

However, this is a major BC break for anybody using the Profiler, and for that alone I just cannot merge it in the current version tree, sorry.

paralogizing:
How is it a major BC break?  It works exactly the same without anyone changing anything.  The only thing it introduces is a new view.  If anyone has extended it, they would have had to of done so drastically, and I doubt even they would be affected.

narfbg:
If 3.0.0 requires nothing for this to work, and then you change it in 3.0.1 to *require* a view, how is it not a BC break?

paralogizing:
Well it includes the view automatically, but okay.  When would be a better time for me to re-submit?

ZombieProtectionAgency:
@paralogizing 

You can take that view if you would like. Also, I think that generally the upgrade process is "Replace your system folder". The view file would be there for anybody doing a fresh install but for people who just update it would break. If you want to re-submit this then I think you would have to wait for CI 4 at the very least and I think that even then, they prefer to avoid BC breaks at all costs so it might not be OK still.

@narfbg Would it still count as a BC break if the Profiler library were edited to output the values in a simple table if the view was not present? It would function fine, It just wouldn't be pretty unless you download the view file when you update.

narfbg:
@paralogizing 2 months ago ... The next chance for that would be 3.1 or 4.0, both of which are way too far away from now.

@ZombieProtectionAgency Perhaps, but there would still be a lot of questions about that ... I closed this because I simply can't accept a BC break for 3.0.1, but that doesn't mean it's the only thing to have in mind.

If you'd still be outputting directly from the class, why do you care about an optional view?
If that's for better styling, why do you care for styling at all? It's a profiler, not something that you display to your website's visitors.
And that last one is double sided ... even if this wasn't a BC break and it still *required* a view, what would be the point in that, given that it's a development tool only?

paralogizing:
Okay, thank you narf.  The reason I have for styling it is for the administrative backend at BootPress.  I want it to look good for my clients.  Lonnie did the same thing at Bonfire.  I guess we just want our applications to look professional, and for the profiler to be a little more useful than the default is.

narfbg:
Well, that makes sense I guess ... but it has to happen in a way that doesn't *require* a view or additional configuration.

paralogizing:
Well you've got me there.  I'll have to think about that and try again come 3.1.  Thanks again.

ZombieProtectionAgency:
@paralogizing You can always override it and use the view I posted if you think it would make your life a little easier, if it would anyways. I doubt this library will see too many updates lol

@narfbg Thank you for the input. I like paralogizings reasoning and I think it would be nice to have a little more control, by default. I did consider the last point. That was why my original suggestion was to use config options to style it. What is your opinion on using the built in template engine with config values containing the HTML for each section? If the values weren't set it would just print the data in a simple table like I suggested before. 

I think a config layout like I posted before would work well (maybe leaving out the heredoc declarations). All templates for each section would be in an array in the config['templates'] array.

narfbg:
I'm an anti-fan of template engines, and here you wouldn't even be working with a template, but with some config variables ...

ZombieProtectionAgency:
Hmm, I liked the idea of letting the end user completely control the layout. Another option would be to have a list of pre and post html for each variable/area. That would be much simpler to work with in the Profiler library too... Do you think that sounds reasonable? 

Something like...

    $config['layouts'] = array(
    	'benchmarks' => array(
    		'section_prefix' = '',
    		'title_prefix' = '',
    		'title_suffix' = '',
    		'item_prefix' = '',
    		'item_suffix' = '',
    		'section_suffix' = ''
    	),
    	'get' => array(
    		'section_prefix' = '',
    		'title_prefix' = '',
    		'title_suffix' = '',
    		'item_prefix' = '',
    		'item_suffix' = '',
    		'section_suffix' = ''
    	)
    );

PR #3853, opened 2015-05-13
[Cache]minor adjustments for save function

ftwbzhao:
https://github.com/phpredis/phpredis#set


narfbg:
And what if you passed a TTL of 0 in case you don't want the value to expire? Have you verified if/how that works?

ftwbzhao:
https://github.com/phpredis/phpredis/blob/develop/redis_commands.c#L1258

PR #3854, opened 2015-05-13
Fixed bug - using field_data() on Oracle databases

leandrowkz:
When you're using oracle databases and want to retrieve column information through the function field_data($table) you get the following notice: 

- Notice: Undefined property: stdClass::$COLUMN_DEFAULT in system/database/drivers/oci8/oci8_driver.php on line 576;

This happens because the oci8 driver tries to access a property that does not exist on query used to get field information. Checking the code we see a small validation to set default value, but the variable $default is not used. So we fix this bug by simply changing:

$retval[$i]->default = $query[$i]->COLUMN_DEFAULT;
to 
$retval[$i]->default = $default;

Bug fixed. No more notices and the properly value is set.

### DIFF
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -573,7 +573,7 @@ class CI_DB_oci8_driver extends CI_DB {
 			{
 				$default = '';
 			}
-			$retval[$i]->default		= $query[$i]->COLUMN_DEFAULT;
+			$retval[$i]->default = $default;
 		}
 
 		return $retval;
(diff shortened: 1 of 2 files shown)


narfbg:
Great. However, a changelog entry would also be required for this ... can you do that?

leandrowkz:
Ok, done.

narfbg:
Uh, sorry, but "Using field_data() on Oracle databases" is not a bugfix description ... Please replace it with this:

    -  Fixed a bug (#3854) - `field_data()` didn't work properly with the Oracle (OCI8) database driver.

Also, put it *after* the other messages, not above them.

leandrowkz:
Ok. I think it's right now.

PR #3856, opened 2015-05-13
Update postgre_driver.php

russ-ab:
Updated function _execute() - The previews function return a resource or throw a PHP error if Postgres returned an error. Now its do a pg_send_query and get back the result from pg_get_result, with pg_result_error_field we can verify if Postgres returned an error, if yes, send back FALSE and the db_driver class will do the rest like all in others drivers (mysql, oracle, etc) not halting the execution with a PHP error.

Updated function error() - Return the correct CODE and MESSAGE result resource.

### DIFF
--- a/system/database/drivers/postgre/postgre_driver.php
+++ b/system/database/drivers/postgre/postgre_driver.php
@@ -239,7 +239,22 @@ class CI_DB_postgre_driver extends CI_DB {
 	 */
 	protected function _execute($sql)
 	{
-		return pg_query($this->conn_id, $sql);
+            /** Send query */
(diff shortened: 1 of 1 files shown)


narfbg:
This changes the behavior way more than I think you intend it to do ...

For one it allows multiple queries to be executed at once, it makes the method non-blocking and probably introduces an issue where you're not exactly sure which result you're getting back.  
Also, your patch uses an undeclared class property (we don't do that) and doesn't conform to our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html).

You'd do better with trying to explain what is it exactly that you're trying to do ... the description is a bit unclear.

PR #3858, opened 2015-05-16
Add functionality to add output in <head> tag for Profiler

kenjis:
If you extends Profiler or use thrid-party profiler, you may want to add HTML for profiler in <head> tag.

This PR provides such functionality.

### DIFF
--- a/system/core/Output.php
+++ b/system/core/Output.php
@@ -508,9 +508,17 @@ class CI_Output {
 				$CI->profiler->set_sections($this->_profiler_sections);
 			}
 
+			$profiler = $CI->profiler->run();
+
+			if (is_array($profiler))
+			{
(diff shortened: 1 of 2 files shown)


narfbg:
Well, if you extend the profiler, you might as well extend the Output class too ...

Sorry, I just don't see this as something that would be useful to many people, and it makes the coupling between CI_Profiler and CI_Output even stronger, which we should be avoiding.

kenjis:
I want to add `<link>` tag for CSS in `<head>`. I think it is useful to many people, because we could separate CSS for profiling to css files.

But I understood the coupling we avoid.

Do you think we can do something to add `<link>` tag in `<head>`, without making the coupling between CI_Profiler and CI_Output stronger?

To loose coupling between CI_Profiler and CI_Output, It seems we have to change the API.

narfbg:
Read through the discussion on #3847 ... it's all in there.

PR #3859, opened 2015-05-16
Session file driver use system temporary directory

mul14:
If `$config['sess_save_path']` and `session.save_path`
in `php.ini` not set, then use the system temporary directory.

Signed-off-by: Mulia Arifandi Nasution <mul14.net@gmail.com>


feryardiant:
:+1:

narfbg:
No ... If the system's tempdir was an appropriate place to store sessions, it would've been the default.

PR #3864, opened 2015-05-18
Backwards compatibility for uppercase model/controller file names

DammitDave:
Added backwards compatibility config item - for all my controller / models I don't have the file names upper case first - presumably others do too. Added the config item, implemented in model loader & controller route loading (and error class loading), updated the changelog hopefully correctly.


narfbg:
If we wanted to support these cases, we wouldn't make Ucfirst a requirement ...

DammitDave:
May I ask, why don't you want to support these cases?

avenirer:
Because the idea was to get rid of all these "elseif"s.

PR #3866, opened 2015-05-19
should have news prefix

gavinr:
The sample code should have /news/ prefix so it works correctly.

### DIFF
--- a/user_guide_src/source/tutorial/news_section.rst
+++ b/user_guide_src/source/tutorial/news_section.rst
@@ -151,7 +151,7 @@ and add the next piece of code.
 		<div class="main">
 			<?php echo $news_item['text']; ?>
 		</div>
-		<p><a href="<?php echo $news_item['slug']; ?>">View article</a></p>
+		<p><a href="news/<?php echo $news_item['slug']; ?>">View article</a></p>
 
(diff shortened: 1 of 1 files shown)


narfbg:
1701ad24e7b52df9e0dd51ef7a9a4bb9a99f28bf

vlakoff:
Relative links break so easily when changing site structure, reusing sections...

We should encourage good practices and advise this instead:

    <?php echo site_url('news/'.$news_item['slug']); ?>

(which I'd personally shorten to:)

    <?= url('news/'.$news_item->slug) ?>
    <?= $news_item->url() ?>

gavinr:
@narfbg I see where it was removed, but all I'm saying is that I'm (re-)learning CI and that link does not work for me without `news/` ... maybe I'm missing something.

@vlakoff thanks for that input. Maybe that should be the change instead?

narfbg:
@gavinr More info on the topic: http://forum.codeigniter.com/thread-1217-post-5641.html#pid5641

@vlakoff `site_url()` may do it indeed, good point ... will have to look into it, see if it makes sense within the tutorial (e.g. helpers are not yet explained at that point ...)

PR #3867, opened 2015-05-19
Check for empty values when setting method in Router in case of regex that uses wildcards

lonnieezell:
For example, if I want to pass items to a product lookup method that may or may not have the value, and simply have the product lookup method deal with the error itself, I would use a regex like: 

`$route['product/?([^/]*)'] = 'catalog/product_lookup/$1';`

This results in an empty string when it gets to the routers `set_method()` function.

This fix simply checks that before setting it.

### DIFF
--- a/system/core/Router.php
+++ b/system/core/Router.php
@@ -471,6 +471,8 @@ class CI_Router {
 	 */
 	public function set_method($method)
 	{
+		if (empty($method)) return;
+
 		$this->method = $method;
 	}
 


narfbg:
I don't get it ... can you elaborate?

lonnieezell:
It simply doesn't allow empty strings to be set as the method name during routing. 

The only time I've run into this is when using a regex with an optional segment. Here's an example: 

Say I've got a routes for `/products/category/{optional}`. If I have a parameter in the optional spot it works fine. If I don't, it will currently end up with an empty string as one of the segments. During the routing process it sees that an another segment exists and thinks it should map it to the method. However, doing so ends up with the method being an empty string instead of the default `index`. This happens in the `_set_request()` method around line 270. 

This pull request just makes it so that `index` can never be overwritten by an empty string, which only happens with optional parameters as far as I've ever seen.

Has no effect on the current process and just catches an edge case.

narfbg:
Well, unless `products` in your example is a directory, then your optional argument has nothing to do with `set_method()` ... If it is, then why should we consider arbitrary routing to an unknown method to be a valid case?

lonnieezell:
Products would be the Controller, Category the method. This allows for the category name to be optional. In this example, it might provide a list of categories in the system. 

This example is a single way of setting 2 different routes: 

```
// Replace this: 
$route['products/category/(:any)'] = 'products/category/$1';
$route['products/category'] = 'products/category';

// With: 
$route['products/category/?([^/]*)'] = 'products/category$1';

// where the method is: 
public function category ($slug=null) { . . . }
```

Like I said, an edge case, but we support regex so I thought we should support this portion also.

It has to do with `set_method` because, like I said, it is an empty value in an array that gets created in `parse_request` at line 428, passed through `set_request` and on into `set_method` because the item exists. There is no check along the way for an empty string, which is something that we'd never want to be set as a method name anyway, right?

narfbg:
Yes, but given that `category` is your method, whatever the optional value is - it will never be passed to `set_method()`. It will be passed as your `$slug`, but that you have to validate anyways.

lonnieezell:
I've followed it through and this was the fix that was required. Watch in XDebug as it goes through with an optional tag and you'll see it does get through. And this simple change was the fix. 

If you don't want to include it, that's fine, but at least take a couple of minutes to walk through the process before deciding, please.

lonnieezell:
I guess it boils down to this, really: Is there any time that you would want to set an empty string as the method name? If so, then, cool, I've missed that. If not, then what's the harm?

narfbg:
The harm could be twofold - enabling you to declare routes that you shouldn't declare (which I already raised as a concern), or masking another bug (an empty string should never be *passed* to `set_method()` in the first place).

If what you say is correct, I'm more concerned with the latter, but indeed - I have to test first.

lonnieezell:
I'm pretty sure the only reason it happens here is that the optional parameter shows up as an empty match in the regular expression initially. I don't believe it can happen at other times but I'm not near my computer currently.

narfbg:
Can't reproduce it ... Diff is against a fresh clone from the 3.0.0 tag.

<pre>
$ git diff HEAD
diff --git a/application/config/routes.php b/application/config/routes.php
index a98c6d1..0256c0b 100644
--- a/application/config/routes.php
+++ b/application/config/routes.php
@@ -50,5 +50,6 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 |              my-controller/my-method -> my_controller/my_method
 */
 $route['default_controller'] = 'welcome';
+$route['product/?([^/]*)'] = 'products/category/$1';
 $route['404_override'] = '';
 $route['translate_uri_dashes'] = FALSE;
diff --git a/application/controllers/Products.php b/application/controllers/Products.php
new file mode 100644
index 0000000..08481d7
--- /dev/null
+++ b/application/controllers/Products.php
@@ -0,0 +1,7 @@
+<?php
+class Products {
+
+       public function index() {}
+
+       public function category($foo = null) {}
+}
diff --git a/system/core/Router.php b/system/core/Router.php
index eb3da22..790daf9 100644
--- a/system/core/Router.php
+++ b/system/core/Router.php
@@ -471,6 +471,7 @@ class CI_Router {
         */
        public function set_method($method)
        {
+               var_dump($method);
                $this->method = $method;
        }
 
$ php index.php products
string(8) "category"
$ php index.php products/
string(8) "category"
</pre>

lonnieezell:
Hmmm. Was working off of current develop branch I believe. Out of town for the weekend here. Will take a look early next week.

lonnieezell:
I'm not sure what the heck was going on last week but I've tested against both the current dev branch and the master branch and both work just fine with the optional tag. 

Frustrated - and closing. :)

PR #3868, opened 2015-05-20
optimize migrations class reference

ftwbzhao:
[Code](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Migration.php#L194) and [User Guide](http://www.codeigniter.com/user_guide/libraries/migration.html#CI_Migration::version) keep the consistency.

### DIFF
--- a/user_guide_src/source/libraries/migration.rst
+++ b/user_guide_src/source/libraries/migration.rst
@@ -164,7 +164,7 @@ Class Reference
 
 	.. php:method:: latest()
 
-		:returns:	TRUE if no migrations are found, current version string on success, FALSE on failure
+		:returns:	current version string on success, FALSE on failure
 		:rtype:	mixed
 
 		This works much the same way as ``current()`` but instead of looking for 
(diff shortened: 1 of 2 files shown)


user_guide_src/source/libraries/migration.rst
```diff
@@ -164,7 +164,7 @@ Class Reference
 
 	.. php:method:: latest()
 
-		:returns:	TRUE if no migrations are found, current version string on success, FALSE on failure
```
ftwbzhao: https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Migration.php#L303

narfbg: 'current' -> 'Current'

user_guide_src/source/libraries/migration.rst
```diff
@@ -164,7 +164,7 @@ Class Reference
...
+		:returns:	current version string on success, FALSE on failure
```
narfbg: Same here.

PR #3870, opened 2015-05-21
Update Input.php

eecjimmy:
where $name is array and  not include 'expire'

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -353,7 +353,7 @@ class CI_Input {
 	 * @param	bool		$httponly	Whether to only makes the cookie accessible via HTTP (no javascript)
 	 * @return	void
 	 */
-	public function set_cookie($name, $value = '', $expire = '', $domain = '', $path = '/', $prefix = '', $secure = FALSE, $httponly = FALSE)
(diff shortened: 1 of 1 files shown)


narfbg:
Explain please ... this doesn't seem to make any sense.

eecjimmy:
oh, my english is so poor ^_^

eecjimmy:
if $name is an array but not explain 'expire', and when function argument['expire'] is not set, will can't setcookie success.

narfbg:
Yes it [can](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Input.php#L395).

eecjimmy:
my PHP version is 5.3.29.
when code like below, can set cookie success.
```
setcookie('test','test');
```
but when code like this, set cookie failed.
```
$this->input->set_cookie('test2','test2');
```

narfbg:
It doesn't fail, it deletes the cookie instead.

That's how this method has always worked ... it's not just an alias for `setcookie()`.

eecjimmy:
This may make people for their mistakes, perhaps, we should follow usage.

eecjimmy:
however , tks.

PR #3872, opened 2015-05-21
Fix wordwrap bug

w0den:
Fixed a bug due to which some email preferences (`wordwrap`, `mailtype`, `wrapchars`) are ignored, and as a result it breaks message body markup.

### DIFF
--- a/system/libraries/Email.php
+++ b/system/libraries/Email.php
@@ -1526,7 +1526,7 @@ class CI_Email {
 
 				// If we're at the character limit, add the line to the output,
 				// reset our temp variable, and keep on chuggin'
-				if ((strlen($temp) + strlen($char)) >= 76)
+				if ($this->wordwrap === TRUE && $this->mailtype !== 'html' && (strlen($temp) + strlen($char)) >= $this->wrapchars)
 				{
 					$output .= $temp.$escape.$this->crlf;
 					$temp = '';


narfbg:
The method that you're modifying deals with message encoding and the way it works is defined by an [IETF standard](http://www.ietf.org/rfc/rfc2045.txt). It has nothing to do with word-wrapping.

w0den:
You generally tried to send HTML emails using this class? I spent a couple of hours to figure out why HTML-markup is break and I came to the conclusion, that when developer pointed out that the `wordwrap == FALSE` it should be `FALSE`. Same thing with `wrapchars`. If I set up that class it should not cut lines, it should not do it.

At least you need to specify in the documentation — "hey, guys, do not try to change these settings, they will still be ignored".

mwhitneysdsu:
@w0den the standard referenced by narfbg includes this:

```
(Soft Line Breaks) The Quoted-Printable encoding
          REQUIRES that encoded lines be no more than 76
          characters long.
```

So, the modification you've made/requested violates the requirements of the standard for Quoted-Printable Content-Transfer-Encoding in the IETF standard.

There are many reasons that HTML email is a pain, but line breaks are generally not a big concern for HTML (except for one or two elements, like pre).

w0den:
Oh my God! Andrew, excuse me please — you're absolutely right. I really did not notice which standard I tried to violate. It seems I need to sleep more.

Matt, I am very grateful to you for quoted text. You saved my life :)

PR #3874, opened 2015-05-22
Add functionality to load namespaced libraries

kenjis:
This PR enables to load namespaced libraries using (Composer) auto-loading.

```php
$this->load->library('Vendor\Project\Foo');
```

You can use the object as `$this->foo`.
And it uses config file `config/vendor_project_foo.php`, if it exists.

How do you guys think?


narfbg:
If it's auto-loading, you don't need to call `$this->load->library()`, just instantiate it as a regular object.

kenjis:
Thank you for your comment. I got it.

PR #3881, opened 2015-05-26
Update url_helper.php

orhanbhr:
redirect refresh added to time


narfbg:
The 'Refresh' header is a non-standard one and while we may keep it for backwards-compatibility, we won't be enhancing its support.

Issue #3880, opened 2015-05-26
PHP Coding style


FDiskas:
Are you sure that TRUE and FALSE must be in uppercase?
https://github.com/bcit-ci/CodeIgniter/blob/develop/user_guide_src/source/general/styleguide.rst#true-false-and-null
I suppose that only real constants should be in uppercase. Lets see what we got here: http://3v4l.org/4Ze6h
I know that uppercase is more readable, but it's not true in meaning of constant. http://php.net/manual/en/language.constants.php "By convention, constant identifiers are always uppercase."

FDiskas:
And one more example that TRUE and FALSE can be changed http://3v4l.org/S2AFZ
http://php.net/manual/en/language.constants.php "value cannot change during the execution of the script"

FDiskas:
I like this coding standard "psr-2" http://www.php-fig.org/psr/psr-2/#2.5.-keywords-and-true/false/null

jim-parry:
This is a CodeIgniter convention.
If you have comments on this, raise them on forum.codeigniter.com, not here ... github is for development, not support.

FDiskas:
I'm not asking for support :D

ZombieProtectionAgency:
@FDiskas The forum can also be used to ask about conventions or changing things that arent broken. The github issue tracker is generally only for bug reports.

PR #3882, opened 2015-05-26
Update styleguide.rst

FDiskas:
https://github.com/bcit-ci/CodeIgniter/issues/3880


narfbg:
You would've done better to comment about this on the forums first, as it was suggested.

Yes, we're very aware that you can `define('TRUE', FALSE)`, but that won't change the styleguide for CI3.

PR #3883, opened 2015-05-26
URL helper - Change redirect() signature

vlakoff:
Just a proposal to have your opinion on this.
Refinements yet to be done (documentation, etc.)

This let us do `redirect('foobar', 301)` instead of `redirect('foobar', 'auto', 301)`, the "auto" parameter being quite superfluous.


narfbg:
I agree that it would be a nicer API, but it's also a BC break and therefore can't happen in a minor version release.

CI4 is currently in planning and this will surely be changed in there, but can't consider it for 3.x.

PR #3884, opened 2015-05-27
Make sure not to show Error on production environment

kenjis:
I got the error display on production environment.

```
A Database Error Occurred

Error Number: HY000/1
no such table: project
SELECT * FROM "project" ORDER BY "score" DESC, "url" ASC LIMIT 100
Filename: models/ProjectRepository.php
Line Number: 70
```

### DIFF
--- a/system/core/Exceptions.php
+++ b/system/core/Exceptions.php
@@ -182,7 +182,12 @@ class CI_Exceptions {
 		include($templates_path.$template.'.php');
 		$buffer = ob_get_contents();
 		ob_end_clean();
-		return $buffer;
+
+		if (str_ireplace(array('off', 'none', 'no', 'false', 'null'), '', ini_get('display_errors')))
+		{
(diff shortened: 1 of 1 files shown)


narfbg:
This is controlled by configuration.

kenjis:
@narfbg What do you mean? How do I configure not to show error?

narfbg:
For a database error? By setting 'db_debug' to FALSE.

kenjis:
@narfbg Thank you!

User Guide:
> Conversely, setting the constant to ‘production’ will disable all error output. 

The description is not a bit correct.  `production` and `db_debug` to `FALSE` will disable *all* error output. Right?

narfbg:
Well, it's not incorrect, but rather 'db_debug' is overriding this ...

I see how it can be tricky, I'll think about the best way to solve that.

PR #3888, opened 2015-05-29
Clean up. Variable block indentation

itrea:
This has been a thorn in my eye for the past 5 years. I could not find anything in the styleguide about variable block indentation, so I thought I would give it a shot. If this one gets merged into base, then I will go through more of the code a start indenting.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -181,11 +181,11 @@ $config['permitted_uri_chars'] = 'a-z 0-9~%.:_\-';
 | use segment based URLs.
 |
 */
-$config['allow_get_array'] = TRUE;
+$config['allow_get_array']      = TRUE;
 $config['enable_query_strings'] = FALSE;
-$config['controller_trigger'] = 'c';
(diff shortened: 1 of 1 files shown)


narfbg:
That's alignment, not indentation ... I know it may seem the same, but there's a slight difference.

There's nothing in the styleguide about it because it's appropriate in some cases and annoying in others - it can't possibly be a hard rule. We don't consider it important and thus, patches like the one you're proposing fall in the "change for the sake of change" category and are not accepted. Sorry.

itrea:
Thats ok. Thanks for the explanation.
I just noticed that the block that is already aligned, has been done using tabs, thats why it looks weird for me in vi.

narfbg:
That ... yea, because alignment is often confused with indentation and we do require tabs for indentation, some of that exists (I'm guilty of it as well). Of course, it became annoying so we now accept only "no alignment" or "align with spaces", kind of unofficially, but current code remains as is unless you need to change it for another reason. :)

PR #3893, opened 2015-06-03
made all form helpers consistent regarding extra attributes

darkhouse:
Extra attributes in form methods can now all be passed as arrays or strings, form_dropdown and form_open already behaved that way.

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -483,8 +483,6 @@ if ( ! function_exists('form_radio'))
 		is_array($data) OR $data = array('name' => $data);
 		$data['type'] = 'radio';
 
-		$extra = _attributes_to_string($extra);
-
 		return form_checkbox($data, $value, $checked, $extra);
 	}
 }
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -197,7 +197,7 @@ if ( ! function_exists('form_input'))
 	 *
 	 * @param	mixed
 	 * @param	string
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_input($data = '', $value = '', $extra = '')
@@ -207,6 +207,8 @@ if ( ! function_exists('form_input'))
 			'name' => is_array($data) ? '' : $data,
 			'value' => $value
 		);
+
+		$extra = _attributes_to_string($extra);
 
 		return '<input '._parse_form_attributes($data, $defaults).$extra." />\n";
 	}
@@ -223,7 +225,7 @@ if ( ! function_exists('form_password'))
 	 *
 	 * @param	mixed
 	 * @param	string
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_password($data = '', $value = '', $extra = '')
@@ -245,7 +247,7 @@ if ( ! function_exists('form_upload'))
 	 *
 	 * @param	mixed
 	 * @param	string
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_upload($data = '', $value = '', $extra = '')
@@ -253,6 +255,9 @@ if ( ! function_exists('form_upload'))
 		$defaults = array('type' => 'file', 'name' => '');
 		is_array($data) OR $data = array('name' => $data);
 		$data['type'] = 'file';
+
+		$extra = _attributes_to_string($extra);
+		
 		return '<input '._parse_form_attributes($data, $defaults).$extra." />\n";
 	}
 }
@@ -266,7 +271,7 @@ if ( ! function_exists('form_textarea'))
 	 *
 	 * @param	mixed	$data
 	 * @param	string	$value
-	 * @param	string	$extra
+	 * @param	mixed	$extra
 	 * @return	string
 	 */
 	function form_textarea($data = '', $value = '', $extra = '')
@@ -287,6 +292,8 @@ if ( ! function_exists('form_textarea'))
 			unset($data['value']); // textareas don't use the value attribute
 		}
 
+		$extra = _attributes_to_string($extra);
+
 		return '<textarea '._parse_form_attributes($data, $defaults).$extra.'>'.html_escape($val)."</textarea>\n";
 	}
 }
@@ -301,11 +308,13 @@ if ( ! function_exists('form_multiselect'))
 	 * @param	string
 	 * @param	array
 	 * @param	mixed
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_multiselect($name = '', $options = array(), $selected = array(), $extra = '')
 	{
+		$extra = _attributes_to_string($extra);
+
 		if ( ! strpos($extra, 'multiple'))
 		{
 			$extra .= ' multiple="multiple"';
@@ -420,7 +429,7 @@ if ( ! function_exists('form_checkbox'))
 	 * @param	mixed
 	 * @param	string
 	 * @param	bool
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_checkbox($data = '', $value = '', $checked = FALSE, $extra = '')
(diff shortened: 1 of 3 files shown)


narfbg:
If you want to pass attributes as an array, you can just use the first parameter as an array for that ...

Adding the "extra" parameter was a mistake on our part ... adding more of it for "consistency" would be a bigger mistake.

darkhouse:
I think that passing a string for the $extra param is a mistake, it's messy, but I disagree that having it at all is a mistake. I don't normally pass my field config as an array. It's easier to just to form_input('name', set_value('name')) than it is to pass all of that as an array. I don't like the idea of having to send all params in an array just to add a class to my input, that's silly. Plus, I don't think the dropdown helper allows you to send an array, so it is still inconsistent.

Is the code ever likely to change to force us to send everything as an array? Will the $extra field be deprecated in the future? I think that would hurt more than just allowing us to pass that param as an array.

narfbg:
Well, 'name' itself is an attribute and I don't see anything silly about passing all attributes at once ... quite the opposite - passing one attribute separately from the others is what I would call "silly". :)

darkhouse:
form_input('name', set_value('name')) // short and sweet
form_input(array('name' => 'name', 'value' => set_value('name')) // almost double the code

I think the ability to pass everything as an array is fine, but you're saying if you want to set a class on a field it's better to send everything as an array, which again, won't work on dropdowns, yet dropdown is the only input helper that already allows sending the $extra param as an array. 

What if I've already got a giant form in place using the "short and sweet" system, and now for some reason I have to add a class to a bunch of fields? I'd have to change the code for all those fields and move everything to an array rather than just adding the extra class on the end of them.

I don't think having the $extra param is silly at all, I don't think sending it as an array is silly. The name and value params are mandatory, so they can stay as separate params. There's nothing wrong with tacking on an array of extra attributes for class, style, data-*, placeholder, etc. The idea here is that I think everything should be consistent. Dropdown already allows an array, form_open allows an array. Arrays are certainly cleaner code than messy strings, they should all allow arrays, or strings so you're not breaking existing forms.

Of course, I can just use my own form helpers and not worry about adding this into the core but I really think this should be part of the core, and would be beneficial to a lot of people. I don't believe having the $extra param is a mistake, and I don't think passing it as an array is a bigger mistake. I think the idea of having to pass everything as an array just to add a class is silly.

narfbg:
If you ask me, `form_dropdown()` is the only *useful* "form helper", as it does make a lot of heavy lifting for you, while with the rest of them you just write PHP code instead of HTML, but with no functional benefit ...

But anyway, you're arguing in favor of the "extra" parameter, yet you're giving examples without it ...

    form_input('name', set_value('name')) // short and sweet
    form_input(array('name' => 'name', 'value' => set_value('name')) // almost double the code

Now, this looks a lot more different:

    form_input('name', set_value('name'), array('class' => 'foo', 'id' => 'bar');
    form_input(array('name' => 'name', 'value' => set_value('name'), 'class' => 'foo', 'id' => 'bar'));

You can argue that the first one is shorter, but you're also using "consistency" as an argument and everything in one array is way more consistent than three parameters working in many different ways under multiple conditions. Not to mention the maintenance cost that comes with this.

Nobody is saying that you wouldn't still be able to use `form_open('name', 'value')`, but if you want to pass an array of attributes ... I think I've already made my point about that.

darkhouse:
I'm arguing that most fields I don't need the extra attributes, so it would be inconsistent for me to pass everything as an array only for the ones I do need extra attributes. And it would be more code to always pass a single array for all fields (I know you're not suggesting that, but for consistency's sake that's how it would be). I think your idea of consistency and mine don't match. Clearly we have different development styles and preferences. I think my solution is an improvement and cleans things up a bit. Your response is basically saying I should keep my code inconsistent, because even if I wanted to pass everything as an array, I couldn't with dropdowns. Inconsistency is consistent. I guess custom form helpers it is.

benedmunds:
+1 to this PR

Just because we might want to move away from this in the future doesn't mean that we shouldn't improve the framework and make it more consistent as is.  I also prefer passing in an array as the extra param, building the entire element as an array is cumbersome, especially in a view.

narfbg:
@lonnieezell @jim-parry What do you guys think?

jim-parry:
+1
I don't see a downside to this.

lonnieezell:
+1
Consistency is a good thing, even if PHP won't give it to us natively. :)

narfbg:
I guess the consistency obsession wins then ...

system/helpers/form_helper.php
```diff
@@ -301,11 +308,13 @@ function form_textarea($data = '', $value = '', $extra = '')
 	 * @param	string
 	 * @param	array
 	 * @param	mixed
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_multiselect($name = '', $options = array(), $selected = array(), $extra = '')
 	{
+		$extra = _attributes_to_string($extra);
```
narfbg: This is unnecessary, because `form_multiselect()` calls `form_dropdown()`, which will handle this.

system/helpers/form_helper.php
```diff
@@ -464,13 +475,16 @@ function form_checkbox($data = '', $value = '', $checked = FALSE, $extra = '')
...
-	 * @param	string
+	 * @param	mixed
 	 * @return	string
 	 */
 	function form_radio($data = '', $value = '', $checked = FALSE, $extra = '')
 	{
 		is_array($data) OR $data = array('name' => $data);
 		$data['type'] = 'radio';
+
+		$extra = _attributes_to_string($extra);
```
narfbg: Also unnecessary, because `form_checkbox()` is called afterwards.

user_guide_src/source/helpers/form_helper.rst
```diff
@@ -191,7 +191,7 @@ The following functions are available:
 
 	:param	array	$data: Field attributes data
 	:param	string	$value: Field value
-	:param	string	$extra: Extra attributes to be added to the tag *as is*
+	:param	mixed	$extra: Extra attributes to be added to the tag *as is* either as array or string
```
narfbg: If it is "array or string", it's no longer "as is" ...

narfbg:
In addition to the inline comments that I left, a changelog entry is also necessary.

system/helpers/form_helper.php
darkhouse: I added it here because it appends $extra .= ' multiple="multiple"'; there. We have to make sure it's a string first, and then we pass it all on to form_dropdown

narfbg: I see ... OK, that makes sense. :+1:

narfbg:
Sorry ... there are no notifications for PR commits and I just now saw you pushed the changes. Unfortunately my latest changes to changelog.rst created a conflict with this patch and you'll have to pull my commits before it becomes mergeable.

I'll drop some more nitpicky comments on top of that. :)

user_guide_src/source/changelog.rst
```diff
@@ -8,13 +8,19 @@ Version 3.0.1
 Release Date: Not Released
 
 -  Core
-
+
```
narfbg: You're adding some whitespace here (and after "Database" too).

user_guide_src/source/changelog.rst
```diff
@@ -8,13 +8,19 @@ Version 3.0.1
...
-
+-  Database
+   
    -  Added ``list_fields()`` support for SQLite ('sqlite3' and 'pdo_sqlite' drivers).
 
+-  Helpers
+   
+   -  :doc:`Form Helper <helpers/form_helper>` changes include:
+
+      -  Made all form helpers consistent by allowing an array to be passed for extra attributes.
```
narfbg: Remove this line and just change the parent one to this:

     -  :doc:`Form Helper <helpers/form_helper>` functions now all accept an array for ``$extra`` attributes.

user_guide_src/source/helpers/form_helper.rst
```diff
@@ -191,7 +191,7 @@ The following functions are available:
 
 	:param	array	$data: Field attributes data
 	:param	string	$value: Field value
-	:param	string	$extra: Extra attributes to be added to the tag *as is*
+	:param	mixed	$extra: Extra attributes to be added to the tag either as array or string
```
narfbg: 'either as *an* array or *a literal* string'

darkhouse:
Thanks for merging and making those changes. I'm sorry I wasn't able to do this sooner, been too busy to get back to it.

narfbg:
No worries.

PR #3896, opened 2015-06-05
Updated the MySQLi Driver to Support SSL Connections

### LINKED ISSUE #3781: Add SSL Support to MySQLi Driver
There has been a hack out there for years to enable this. Really should be put into core. Given the security issues that have risen over the past couple of years there really is no excuse for not using encrypted communications between systems internally. I will fork and submit a proper pull request for this feature once I get it working and tested. You can find a snippet of that core hack on this pastbin (http://pastebin.com/ekVJMdRG). I'm working on something a bit more complete.

timnolte:
Uses the DB_driver class encrypt option as the flag for turning on encryption. Also added SSL connection validation with error logging in order to provide users a way to know if they are actually connecting via SSL. This commit fixes #3781

Signed-off-by: Tim Nolte <noltet@sekisui-spi.com>


timnolte:
OK, so I added some additional parameters to the MySQLi driver that I realize that I didn't add to the database documentation. The options I added are not actually critical to SSL work depending on the users' server setup. If I need to I can add those additional parameters to the documentation and submit a new pull request.

timnolte:
OK, went back and made some additional updates to make sure I covered documentation/example requirements. This PR should be all set barring any disagreement with the implementation.

narfbg:
It's a great feature, but adding 6 config options that work only with a single database driver ... I'm really not sure about that.

timnolte:
@narfbg OK, well 5 of those are for SSL specifically. We could make that 1 config option as an array (ssl_options). I guess I didn't see a problem since the MySQLi driver already pushes 3 unique config options (compress/delete_hack/stricton). I want to make sure that the SSL implementation is complete and won't result in some users having to hack core in order to get their SSL setup working. What would you rather see for this? I'm open to whatever alternatives better fit with the vision of the project. Thanks!

narfbg:
Well, we try to only have features that are available to at least few drivers. 'delete_hack' and 'stricton' are MySQL-only, but not limited to the 'mysqli' driver (and to be honest, they are quite old too; the former probably wouldn't be accepted today),

'compress' being an option (not an additional QB method, for example) made it acceptable, but it's a *single* option ... Squeezing these 5 into an array might work, but please remove the 'db_options' one.

timnolte:
OK, well I guess I disagree about removing 'db_options' as it's generic enough that it could be used in other drivers to set database type specific things (i.e. for Oracle a way to call oci_set_client_identifier or set sqlsrv_configure option with SQL Server connections). Is it acceptable to at least add an 'ssl_verify_cert' option to the new 'ssl_options' array and then use that to set the MySQLi option 'MYSQLI_OPT_SSL_VERIFY_SERVER_CERT'?

narfbg:
No ... that would belong in a "db_options"-like setting, but you shouldn't introduce 2 features with 1 pull request. We can think of a solution for that later.

narfbg:
I see you've updated the patch, please drop a comment next time (can't get notifications from commits).

Anyway, I'm still unhappy with it and it would take a lot of back-and-forth comments until we reach a point where it's acceptable for merge ... nothing wrong with it per se, just doesn't meet my "quality standards" if I can put it that way.  
Mind if I refactor it and see where we can get it from there?

timnolte:
@narfbg refactor away. I guess I don't care how the SSL support gets in there, I just figured I'd be apart of the solution and help. I'll just take a look at your code so that I'm aware of what the coding standard is going forward.

narfbg:
Well, I'll pull your patch and work on top of it - that way you'll still get attribution. :)

timnolte:
OK, thanks. Sounds good.

narfbg:
@timnolte Please see the referenced commit above and tell me if I've missed anything or if you've got any concerns about it.

Also, you mentioned an MYSQLI_OPT_SSL_VERIFY_SERVER_CERT option, but I can't find any documentation on that ... How/what do you know about it?

TODO: Add the same stuff to 'pdo/mysql'.

narfbg:
Added PDO support as well ... it's a few more commits now, so you might want to look at this diff instead: https://github.com/bcit-ci/CodeIgniter/compare/develop...feature/mysqli_ssl

timnolte:
I'm concerned about your reuse of the $ssl array for 2 different use cases.(setting up the SSL parameters and checking it in the SSL connection verification) This seem like that opens this code up for more opportunities for improper validation. I also don't like the use of empty with array keys as this has the potential to break on PHP older than 5.4, like 5.3 specifically. I'm fine with that so long as the CI README is update to reflect that anything below PHP 5.4 should now explicitly not be used. The README eludes to that currently but this sort of use of empty() to me should dictate that anything below 5.4 should no longer be supported.

So that MYSQLI_OPT_SSL_VERIFY_SERVER_CERT constant is an as yet undocumented option. It controls whether the server certificate is verified. You can find it mentioned in the following places:
http://svn.php.net/viewvc?view=revision&revision=302897
https://bugs.php.net/bug.php?id=54962

narfbg:
I don't understand your concern about `$ssl`, at all. If it's some sort of "code quality" criteria on your part, then it directly contradicts with mine ... :)

`empty()` doesn't break anything with array keys: http://3v4l.org/OOjWq

Thanks for the clarification on MYSQLI_OPT_SSL_VERIFY_SERVER_CERT. I've added it here: cfc9e77c89ee5377b25e411ef3d8ab43c8900b7e

timnolte:
Actually, if you read the PHP manual `empty()` does have a problem if `['encrypt']` is set to an empty string. http://3v4l.org/rFP6O
http://php.net/manual/en/function.empty.php (See Example #2 empty() on String Offsets).

timnolte:
Also, as for the reuse of $ssl, it comes down to clear readability. You have an $ssl variable that is used one way, as parameter storage, in one place then as something completely different (SSL result status) in another place with the difference being you are checking a second numerical index only which has no readable meaning. A numerical index could easily be mistakenly overwritten and therefore break validation.

narfbg:
That's not a problem just for `empty()` ... you'll get that with anything, including `array_key_exists()`. That's why there's an `is_array()` check wrapping the whole thing. :)

On the `$ssl` thing - that numerical index will not easily be mistakenly overwritten ... it doesn't exist in the first place and even if it did - it's a good thing that it would be overwritten and that's very much intentional. 
But anyway, this is just chit-chat as I got that numerical index check from [your patch](https://github.com/ndigitals/CodeIgniter/commit/a0f1872e4978304a3b096ce90ee539c5e481b4f2#diff-1081d7a1c04389eb6a0cc1eb6d68bb67R206) (reusing the variable is my part of course) and subsequently replaced it in my third commit after pulling your changes, here: 9194b492f900b05acd204cb1b4a524149402be75

PR #3897, opened 2015-06-06
Add namespaced controller

kenjis:
How about this?

If we use sub folder, we sometimes create the same name controller classes.
If we could use namespaced controllers, we can avoid name conflicts.

For example:
```
controllers/Welcome.php     -> App\Controller\Welcome
controllers/foo/Welcome.php -> App\Controller\Foo\Welcome
```

If there are controllers with the same class name, we can't do phpunit testing.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -496,3 +496,14 @@ $config['rewrite_short_tags'] = FALSE;
 | Array:		array('10.0.1.200', '192.168.5.0/24')
 */
 $config['proxy_ips'] = '';
+
+/*
+|--------------------------------------------------------------------------
+| Controller Namespace
+|--------------------------------------------------------------------------
+|
+| This item allows you to set namespace to controllers.
+|
+|	$config['controller_namespace'] = 'App\\Controller';
+*/
+$config['controller_namespace'] = '';
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -397,6 +397,22 @@ if ( ! is_php('5.4'))
 
 	$e404 = FALSE;
 	$class = ucfirst($RTR->class);
+	$namespace = config_item('controller_namespace');
+	if ($namespace)
+	{
+		if ($RTR->directory)
+		{
+			$classname = $namespace.'\\'.$RTR->directory.'\\'.$class;
+		}
+		else
+		{
+			$classname = $namespace.'\\'.$class;
+		}
+	}
+	else
+	{
+		$classname = $class;
+	}
 	$method = $RTR->method;
 
 	if (empty($class) OR ! file_exists(APPPATH.'controllers/'.$RTR->directory.$class.'.php'))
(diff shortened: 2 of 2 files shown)


lonnieezell:
If you're using Composer as your autoloader, there's no need for anything like this. Simply add the folders you want your controllers to be found in as their namespace. 

Something like this in composer.json:

```
"autoload": {
  "psr-4": {
    "App\\Controllers\\": "application/controllers",
    "App\\": ["application/libraries", "application/models"]
  }
}
```

This example also allows the main App namespace to find files in the libraries or models folder.

kenjis:
I did as your way, but I got "404 Page Not Found".

~~~diff
diff --git a/application/config/config.php b/application/config/config.php
index 94e5d28..7ca76a3 100644
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -131,7 +131,7 @@ $config['subclass_prefix'] = 'MY_';
 | Note: This will NOT disable or override the CodeIgniter-specific
 |  autoloading (application/config/autoload.php)
 */
-$config['composer_autoload'] = FALSE;
+$config['composer_autoload'] = TRUE;
 
 /*
 |--------------------------------------------------------------------------
diff --git a/application/controllers/Welcome.php b/application/controllers/Welcome.php
index 34535ef..64e1e10 100644
--- a/application/controllers/Welcome.php
+++ b/application/controllers/Welcome.php
@@ -1,7 +1,9 @@
 <?php
+namespace App\Controller;
+
 defined('BASEPATH') OR exit('No direct script access allowed');
 
-class Welcome extends CI_Controller {
+class Welcome extends \CI_Controller {
 
    /**
     * Index Page for this controller.
diff --git a/composer.json b/composer.json
index 0653a78..2977490 100644
--- a/composer.json
+++ b/composer.json
@@ -15,5 +15,10 @@
    },
    "require-dev": {
        "mikey179/vfsStream": "1.1.*"
+ },
+ "autoload": {
+     "psr-4": {
+         "App\\Controller\\": "application/controllers"
+     }
    }
 }
\ No newline at end of file
~~~

~~~
$ composer update
~~~

Am I something missing?

lonnieezell:
I would do a `composer dump-autoload`. That has a tendency to fix things when I've added new namespaces into the mix. 

Also - have you run `composer install` in that project before to get composer setup correctly?

kenjis:
I run `composer update`. It generates autoload files.

> Also - have you run composer install in that project before to get composer setup correctly? 

I don't remember.

How about this?
~~~
$ rm -rf vendor/
$ composer update
~~~

I still get "404 Page Not Found".

I don't understand why your way could run controller method.
If I access `/`, then controller name in CI is `Welcome` and not `App\Controller\Welcome`.

narfbg:
No - it's a BC break. And I don't see the point really; even if we supported namespaces, controllers would be the least useful place to use them.

Judging by your previous PR as well, you seem very eager to get namespaces into CI ... that's all fine, but it's not going to happen with tiny patches and deffinately not in the 3.0.x version tree.

kenjis:
Where is BC break?

The problem is how to test the same name controllers.

For example,

~~~
controllers/Welcome.php
controllers/foo/Welcome.php
~~~

They have the same classname `Welcome`. We can't declare the same name classes.
We don't use the same classname if we want to test them.

narfbg:
Hmm ... I missed the `if ($namespace)` condition - it's not a BC break indeed.

Still, adding a config item for namespaces for the sole purpose of easier testing isn't something I'd be happy with. PHPUnit (which I assume you use for testing) supports process isolation - every testcase can be ran in a different process and you wouldn't have name collisions.

kenjis:
Yes, I use PHPUnit, but process isolation does not help.
Because process isolation automatically includes all previously included files.
After all, I get `Cannot redeclare class Welcome...`

narfbg:
That behavior can also be modified ... https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.preserveGlobalState

kenjis:
Oh, you're right.

I could write test for controllers with name collisions:
https://github.com/kenjis/ci-app-for-ci-phpunit-test/blob/master/application/tests/controllers/sub/Welcome_test.php

Thanks for your help. I could probably write tests with 100% coverage for CodeIgniter applications.

PR #3900, opened 2015-06-08
Precount for loop

acidvertigo:
This just to avoid to have the count function executed on every for loop


narfbg:
It's not executed on every loop ... take a careful look at it.

PR #3901, opened 2015-06-09
$this->db->replace_string() for creating "REPLACE INTO" SQL queries

vlasky:
This is based on insert_string(), but modified to generate "REPLACE INTO" instead of "INSERT_INTO".

As you know, REPLACE INTO behaves differently to INSERT INTO in that if there is an existing row with the same primary key or unique key, the existing row will be replaced with the new data, otherwise it will be inserted.

Also, Some MySQL storage backends like TokuDB perform faster when using REPLACE INTO instead of INSERT INTO.

For these reasons, this is a useful helper.

Without this function, a "REPLACE INTO" query can only be generated in CodeIgniter by performing a string replace on the output of insert_string(), which is messy.

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -1395,6 +1395,45 @@ abstract class CI_DB_driver {
 	// --------------------------------------------------------------------
 
 	/**
+	 * Generate a replace string
+	 *
+	 * @param	string	the table upon which the query will be performed
+	 * @param	array	an associative array data of key/values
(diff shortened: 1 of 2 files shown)


narfbg:
This may complement `insert_string()` and `update_string()`, but those methods are fundamentally broken (they don't work without Query Builder enabled) and only exist for BC purposes. We won't be accepting any features that extend them.

Just use `replace()`.

PR #3913, opened 2015-06-17
Bug Fix

plastiik:
This actually does not work (as told in documentation)

##################################################

All of the methods listed above can be accessed without passing a specific adapter to the driver loader as follows:

$this->load->driver('cache');
$this->cache->memcached->save('foo', 'bar', 10);

##################################################

We need to force setup in order to assign $_memcached, else caching will not work, and all function associated to it. Please do your testing, you will see I'm right.


narfbg:
I did a some larger changes than the one you suggested (although yours should work too). Thanks for the effort though and thanks for reporting it.

PR #3916, opened 2015-06-18
multiple index in update_batch()  ( Active Record )

parthviroja:
i solved issue of assign multiple index not allow in update_batch().

just few line of code we have to modify and , its working fine.


parthviroja:
https://github.com/parthviroja/codeigniter/

i solved this issue

PR #3917, opened 2015-06-18
DB_active_rec.php

parthviroja:
modify only 

update_batch() and set_update_batch() function 

now we can use multiple fileld for where condition

Example:

$data = array(
   array(
      'title' => 'My title' ,
      'name' => 'My Name 2' ,
      'date' => 'My date 2'
   ),
   array(
      'title' => 'Another title' ,
      'name' => 'Another Name 2' ,
      'date' => 'Another date 2'
   )
);

$this->db->update_batch('mytable', $data,array('title','date'));

### DIFF
--- a/system/database/DB_active_rec.php
+++ b/system/database/DB_active_rec.php
@@ -1323,11 +1323,12 @@ class CI_DB_active_record extends CI_DB_driver {
 		// Combine any cached components with the current statements
 		$this->_merge_cache();
 
-		if (is_null($index))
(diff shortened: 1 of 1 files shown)


narfbg:
There are multiple problems with this patch ... There's a syntax error, multiple BC breaks, lack of documentation updates ...

Also, CI2 will not receive any new features and is nearing its end-of-life, so it only gets security fixes.

PR #3924, opened 2015-06-22
Added .ico mime type

EdersonTS:
Just added ico to mime list


yinlianwei:
`.ico` file have different mimes linke: 'image/x-ico','image/x-icon','image/vnd.microsoft.icon'
so you should add
`ico'   =>  array('image/x-ico','image/x-icon','image/vnd.microsoft.icon'),`

narfbg:
7013cd2514b98c168f8edb402792b4d55c0323d6

PR #3935, opened 2015-06-30
Try to solve two problems when bind params to sql

diaosj:
1.In the old version, when the bind param is a scalar 0, then no param will be passed to sql. 
2.When the quotation marker is ", then the old compile_binds() function will replace a chunk inside a string that happens to match the bind marker.


narfbg:
1. You shouldn't be passing a scalar at all.
2. If your bind marker is `"`, then you're obviously trying to break the whole thing ... this can't happen by accident.
3. You're changing every line in that method (and violating our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html) with that), making the actual code changes indistinguishable, so even if there's something valuable in your patch - I can't spot it.

PR #3937, opened 2015-06-30
Fix csv_from_result if contain newline

thormug:
If $item is a column with a newline inside, the old version create a newline in a result file, breaking the csv.
This fix replace any newline with a new parameter : $newline_delim

### DIFF
--- a/system/database/DB_utility.php
+++ b/system/database/DB_utility.php
@@ -235,7 +235,7 @@ abstract class CI_DB_utility {
 	 * @param	string	$enclosure	Enclosure (default: ")
 	 * @return	string
 	 */
-	public function csv_from_result($query, $delim = ',', $newline = "\n", $enclosure = '"')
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but I don't see how the issue that you're describing could happen. Can you give an example?

thormug:
This function take a sql query in parameter, read it and parse it in CSV.
If in a table there is a line (for example a varchar() or text()) containing a newline, this newline is printed into the result file and break the csv.


Example : 

table db.item :
| id | name | description
---------------------------------
| 1 | example | This is a description with a newline \n this is the second line
| 2 | another | another description


This is the result : 

"id";"name";"description"
"1";"example";"This is a description with a newline
this is the second line";
"2";"another";"another description";

Please note there is 4 lines instead of 3 (2 + head)
The csv is break

narfbg:
That's perfectly valid CSV ... of course a newline character would be rendered as a new line by a text editor, that doesn't mean it's broken.

PR #3938, opened 2015-07-01
Add custom language support for pagination lib

ragamo:
Added ‘language’ param in constructor for translate pagination to a custom language.


mwhitneysdsu:
This is completely unnecessary if you set the language on the current request (before loading the pagination library) via `$this->config->set_item('language', 'language_idiom_here');`

narfbg:
@mwhitneysdsu is right, it doesn't make any sense to configure each library's language separately.

PR #3942, opened 2015-07-02
Eliminated dead code

ftwbzhao:
```php
		$query = $this->_drop_table($this->db->dbprefix.$table_name, $if_exists);

		if ($query === FALSE)
		{
			return ($this->db->db_debug) ? $this->db->display_error('db_unsupported_feature') : FALSE;
		}
		elseif ($query === TRUE)
		{
			return TRUE;
		}
```

> Use `$this->_drop_table`, return type **never equals FALSE**.

So 

```php
		if ($query === FALSE)
		{
			return ($this->db->db_debug) ? $this->db->display_error('db_unsupported_feature') : FALSE;
		}
```
 Eliminated  it.

### DIFF
--- a/system/database/DB_forge.php
+++ b/system/database/DB_forge.php
@@ -454,11 +454,8 @@ abstract class CI_DB_forge {
 		}
 
 		$query = $this->_drop_table($this->db->dbprefix.$table_name, $if_exists);
-		if ($query === FALSE)
-		{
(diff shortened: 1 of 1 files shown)


narfbg:
You're not eliminating dead code with this ... you're just altering my patch that did.

ftwbzhao:
err.. after PR, see your code.

sorry.

ftwbzhao:
Thanks for your patch. :rose:

PR #3945, opened 2015-07-02
General .htaccess file CI

vrkansagara:
this is a general purpose file for CI.


narfbg:
We are intentionally not providing a default .htaccess file. How to create one is explained in the documentation instead.

Also, please don't include extra code in your PRs - your additions to index.php are irrelevant to .htaccess and even if we were to merge the .htaccess addition, I'd still make you revert them first.

PR #3946, opened 2015-07-02
Disable csrf for specific controller or method, when it global enabled in config

dimonneon:
Disable csrf for specific controller or method, when it global enabled in config.
Just use in Controller constructor:
```php
public function __construct(){
  parent::__construct();
  $this->security->disable_csrf = TRUE;
}
```
Or in Controller method:
```php
public function create(){
  $this->security->disable_csrf = TRUE;
  /*some code*/
}
```

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -82,6 +82,15 @@ class CI_Security {
 	public $charset = 'UTF-8';
 
 	/**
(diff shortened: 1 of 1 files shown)


narfbg:
No ... csrf_exclude_uris is enough.

PR #3949, opened 2015-07-03
minor adjustments to DELETE hack flag

ftwbzhao:
http://php.net/manual/en/function.mysql-affected-rows.php#refsect1-function.mysql-affected-rows-returnvalues


narfbg:
No ... this only adds overhead with no benefit.

ftwbzhao:
`DELETE hack flag` may outdated.

mysql api has been supported for many years.

Do you need to remove the hook？

narfbg:
No, outdated or not - removing it would be a BC break.

Please, stop looking for things to change ... don't try to fix something that's not broken.

PR #3950, opened 2015-07-03
PRI_KEY_FLAG is set return 1

ftwbzhao:
http://php.net/manual/en/mysqli-result.fetch-fields.php#101828

```php
$field_data[$i]->flags & 2 //return 2
```

return 1, easier to understand.


narfbg:
Just less than 2 hours ago I told you [not to fix what's not broken](https://github.com/bcit-ci/CodeIgniter/pull/3949#issuecomment-118277079).

PR #3951, opened 2015-07-03
Fix migration_type default value

giancarlorosa:
The default value for migration_type was set as timestamp, but all documentation says that is sequential. So, there was some people that was unaccustomed with this problem than they not find where is wrong.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -29,7 +29,7 @@ $config['migration_enabled'] = FALSE;
 | to 'sequential' for backward compatibility.
 |
 */
-$config['migration_type'] = 'timestamp';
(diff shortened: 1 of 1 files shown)


narfbg:
I don't see the documentation saying that sequential is the default ...

giancarlorosa:
I'm so sorry. You're right! Mabe I thought that because CI2, I dont'n know. Then I just read first part of migration_type comments and assumed sequential as default.
|   'sequential' = Default migration naming (001_add_blog.php)
|   'timestamp'  = Timestamp migration naming (20121031104401_add_blog.php)
|                  Use timestamp format YYYYMMDDHHIISS.

Agains, sorry about this!

Issue #3960, opened 2015-07-07
CI 3.X Router error on _set_default_controller


stefanomozart:
The _set_default_controller method requires the default controller to be a maximum two segment rout (i.e. the controller cannot be in a subdir of the 'controllers' path). Resusing the _set_request inside _set_default_controller corrects the problem

narfbg:
This is by design and explained in the manual.

PR #3961, opened 2015-07-07
Update Router.php

stefanomozart:
Correcting the problem in the _set_default_router, that forced the defult controller to be in the controllers directory


narfbg:
As I said in #3960, this is by design.

PR #3968, opened 2015-07-10
allow add of keys with array

avenirer:
This will allow adding multiple keys using array (http://www.codeigniter.com/user_guide/database/forge.html#adding-keys). Only if user wants, he can use the table columns to set a primary key by setting second parameter as TRUE.

### DIFF
--- a/system/database/DB_forge.php
+++ b/system/database/DB_forge.php
@@ -239,7 +239,7 @@ abstract class CI_DB_forge {
 	 */
 	public function add_key($key, $primary = FALSE)
 	{
-		if ($primary === TRUE && is_array($key))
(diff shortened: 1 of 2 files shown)


narfbg:
This needs a changelog entry, and if the documentation suggests that it is already supported, then it should be a bugfix one.

narfbg:
Didn't really like the changelog entry (and its place - it should be appended to the list, not prepended), so I cherry-picked your first commit and added the message myself.

Thanks for the effort.

avenirer:
Thank you.

narfbg:
This was reverted because it actually broke functionality ... see the commit I referenced above and issue #4173.

PR #3969, opened 2015-07-10
Important note on adding keys add_field() with strings

avenirer:
If you pass strings to the add_field() methods you won't be able to follow those by add_key() methods on those fields.


user_guide_src/source/database/forge.rst
```diff
@@ -143,6 +143,8 @@ string into the field definitions with add_field()
 	$this->dbforge->add_field("label varchar(100) NOT NULL DEFAULT 'default label'");
 
 
+.. note:: Passing strings as fields won't allow you to follow these calls by a add_key() call.
```
narfbg: ..note:: Passing raw strings as fields cannot be followed by ``add_key()`` calls.

PR #3970, opened 2015-07-10
Add unicode support in url_title function

webquail:


system/helpers/url_helper.php
```diff
@@ -500,7 +500,7 @@ function url_title($str, $separator = '-', $lowercase = FALSE)
 		$str = strip_tags($str);
 		foreach ($trans as $key => $val)
 		{
-			$str = preg_replace('#'.$key.'#i', $val, $str);
+			$str = preg_replace('#'.$key.'#ui', $val, $str);
```
narfbg: You need to check for PCRE UTF-8 support first ...

    $str = preg_replace('#'.$key.'#i'.(UTF8_ENABLED ? 'u' : ''), $val, $str);

narfbg:
Also needs a changelog entry.

narfbg:
I meant a feature addition entry as opposed to a bugfix one ... this isn't a bug. I've manually pulled your changes and added the changelog entry as it also needed some formatting. Thanks for the effort.

Issue #3869, opened 2015-05-20
Search highlighting in user guide / documentation


naeon:
While the 3.x documentation looks great overall, there's a pretty huge issue with usability and colors when using the search (on broad topics – e.g. "send email" to look up how to send an email).

On the found pages the highlighting of the search terms breaks every visual hierarchy and distracts and confuses the viewer due to the really intense contrast background.
Example: http://www.codeigniter.com/user_guide/libraries/email.html?highlight=send%20email

![screen shot 2015-05-20 at 17 11 04](https://cloud.githubusercontent.com/assets/5183232/7729111/48575cba-ff13-11e4-8f12-4c6aa2541cfa.png)

How one is supposed to even scan the content for the answer he or she desires is beyond me.
Also when using the search as a replacement for the documentation index, the highlighting (e.g. "input" to quickly access the input class) is completely unnecessary and immensly distracting. 

So at the very least I suggest using a highly transparent color so that the highlighting looks more like that of a textmarker which enables the reader to concentrate on the whole content, not only the highlighted words.

Demo:
![screen shot 2015-05-20 at 17 10 34](https://cloud.githubusercontent.com/assets/5183232/7729130/55df34a2-ff13-11e4-80ae-373fc316cf27.png)

The color `rgba(241, 230, 15, 0.2)` should do it. The horizontal padding of 6px can also be reduced or ommited for the sake of readability.

What do you think?

mwhitneysdsu:
While I agree that the current highlighting style can be painful to read at times, you may need a more objective definition of usability when choosing replacement colors/styles. For instance, the color change you've suggested increases the number of WCAG 2.0 color contrast issues on the page you've chosen for your example, because it is too close to the color used for some of the text on the page (specifically, the green text in the code example).

Of course, the existing highlight style (and the suggested change) currently fails on the red text in the code examples and the orange link text, so the same criteria would suggest that the existing color should not be used, either (or that the color of the highlighted text itself should change, though that might be an additional annoyance).

I agree whole-heartedly with the removal or reduction of the horizontal padding, especially in code examples (and partial-word matches, though those may be harder to differentiate, depending on the code which inserts the spans for the highlighting).

narfbg:
I think you're exaggerating, a lot. But agreed - it is an issue.

narfbg:
Anybody want to take on this and come up with a patch?

mwhitneysdsu:
[...]

PR #3978, opened 2015-07-15
Update theme.css

mrezler:
CSS edit for #3869


narfbg:
Are you familiar with Sphinx and know *for sure* that this is the best way to make the change?

My concerns here: https://github.com/bcit-ci/CodeIgniter/issues/3869#issuecomment-121639129

PR #3985, opened 2015-07-20
Fix link

kenjis:
The current link is 404:
http://www.codeigniter.com/user_guide/general/environments.html


narfbg:
It's just the underscore that had to be removed ... did that instead.

PR #3986, opened 2015-07-20
Path helper remote file inclusion security check fail

rajatsharma94:
The currently implemented security check to make sure the path is NOT a URL can easily be bypassed (gives false negative) for all subdomains. 
Eg "subdomain.domain.com" should ideally show an error but it does not.

The new security check (in addition to the previous one) tries to make a fsockopen connection to validate whether the URL is external or not.

### DIFF
--- a/system/helpers/path_helper.php
+++ b/system/helpers/path_helper.php
@@ -61,6 +61,7 @@ if ( ! function_exists('set_realpath'))
 	function set_realpath($path, $check_existance = FALSE)
 	{
 		// Security check to make sure the path is NOT a URL. No remote file inclusion!
+		// PROBLEM HERE - this can be easily bypassed in case of subdomains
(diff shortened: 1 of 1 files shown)


narfbg:
This is hardly a real *security* check, it's only designed to reject the most obvious URLs and that's just for the developer's convenience, it is not a user input sanitizer function.

`fsockopen()` is a huge overkill for that (especially given that even the path existence checks are optional) and could easily result in false positives - actual domain names are often used as wwwroot directory names.

rajatsharma94:
Okay, just to confirm this again, this is not valid due to the fsockopen() overkill or just simply because that the check is not that important.

What I mean to ask is that if I modify only the regex string to filter out more of those 'obvious URLs' (which won't result in any overkill) then would it be considered useful ?

narfbg:
Both. :)

The regular expression may be modified just as long as you don't make it over-intrusive, i.e. making it reject "subdomain.domain.tld" wouldn't be appropriate due to the possible false positives that I explained in my previous comment.

rajatsharma94:
Yeah got it. As long as I decrease the false negatives and don't increase the false positives it's all good.

PR #3990, opened 2015-07-21
Adding Custom Result object documentation to Query Results docs.

lonnieezell:


user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
 	$query->unbuffered_row('array');	// associative array
 
 *********************
+Custom Result Objects
+*********************
+
+You can have the results returned as an instance of a custom class instead of a stdClass or array,
+as the ``result()`` and ``result_array()`` methods allow.This requires that the class is already
```
mwhitneysdsu: You missed a space between the period and the next sentence.

lonnieezell: Nice catch. Thanks. And fixed.

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
...
+Returns a single row from your query results. The first parameter is the row number of the results.
+The second parameter is the class name to instantiate.
+
+Example::
+
+	$query = $this->db->query("YOUR QUERY");
+
+	if ($query->num_rows() > 0)
+	{
+		$row = $query->custom_row_object(0, 'User);
```
michalsn: Should be `'User'`

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
...
+
+		echo $row->email;   // access attributes
+		echo $row->last_login('Y-m-d');   // access class methods
+	}
+
+You can also use the ``row()`` method in exactly the same way.
+
+Example::
+
+	$row = $query->custom_row_object(0, 'User);
```
michalsn: Same here :)

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
 	$query->unbuffered_row('array');	// associative array
 
 *********************
+Custom Result Objects
+*********************
+
+You can have the results returned as an instance of a custom class instead of a stdClass or array,
+as the ``result()`` and ``result_array()`` methods allow. This requires that the class is already
+loaded into memory. The class will have all values returned from the database set on the class
+as class variables. If your class already has those class variables created and ``protected`` then
```
narfbg: A few problems that I see on this line ...

I'd write 'properties' instead of 'class variables' here - just to use a more proper term; and to eliminate one instance of the word 'class'. That last thing is another that bothers me ... you use 'class' on every line in this paragraph and that appears kinda silly. :) Some occurences may be replaced with 'object' while others may be removed.

Also, what 'created' means here is obvious to you and me, but can otherwise be argued about ... the proper term is 'declared'. And finally, it should be describing non-public as opposed to specificly protected properties. :)

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
...
+you should provide a ``__set()`` method to allow the class variables to be set.
+
+Example::
+
+	class User {
```
narfbg: Our styleguide requires that each opening brace is followed by an empty line.

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
...
+		protected $id;
+		protected $email;
+		protected $username;
+		protected $last_login;
+
+		public function last_login($format)
+		{
+			return date($format, $this->last_login);
```
narfbg: While this is not wrong per se, some would say that we should encourage usage of the DateTime class instead of `date()`. I wouldn't normally raise this as an issue, but since you're also showing usage of `__set()` and `__get()` - that's one of their best use cases.

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,93 @@ the returned value's type::
...
+**custom_result_object()**
+
+Returns the entire result set as an array of instances of the class requested. The only parameter
+is the name of the class to instantiate.
+
+Example::
+
+	$query = $this->db->query("YOUR QUERY");
+
+	if ($query->num_rows() > 0)
```
narfbg: A lot of our drivers need to actually fetch all rows and do a `count()` on the resulting array in order to make `num_rows()` work. Some (if not most) PHP database extensions just don't have buffered result sets ...

Anyway, the result of that is - what's supposed to help performance is now actually hurting it and we should no longer encourage `if ($query->num_rows() > 0)` checks.

lonnieezell: I agree completely. Was just matching what the rest of the docs were showing. Should we open a new pull request to change those examples and get of this practice?

user_guide_src/source/database/results.rst
lonnieezell: Not a problem. I can change that.

user_guide_src/source/database/results.rst
lonnieezell: Good catch. Thanks.

user_guide_src/source/database/results.rst
narfbg: Sure, that would be a good thing to do.

lonnieezell: https://github.com/bcit-ci/CodeIgniter/pull/3993

narfbg: Saw it ... but of course, should be removed from this PR as well. :)

user_guide_src/source/database/results.rst
```diff
@@ -173,6 +173,91 @@ the returned value's type::
...
+
+		public function last_login($format)
+		{
+			return $this->last_login->format($format);
+		}
+
+		public function __set($name, $value)
+		{
+			if ($name === 'last_login')
+            {
```
narfbg: Broken indentation here and the next 2 lines ... I guess you've just copy-pasted from my comment and therefore using spaces instead of tabs.

narfbg:
We also have a soft limit of 76 characters per line on the docs ... It's semi-official and I only noticed it after merging.

Issue #2436, opened 2013-05-05
Feature Request: easy access to validated form data


shijialee:
Hi,

When using CI's Form Validation library, I find that I often have to piece together all the necessary form data to pass to my model. such as:

``` php
$id = $this->blog_m->insert(array(
    'title'                 => $this->input->post('title'),
    'slug'                => $this->input->post('slug'),
    'category_id'    => $this->input->post('category_id'),
    'intro'                => $this->input->post('intro'),
    'body'               => $this->input->post('body'),
    'status'             => $this->input->post('status'),
));
```

That gets very tedious and often introduces typos/bugs.  I don't want to do `$this->blog_m->insert($this->input->post())` because there might be garbage in the postdata that need to be cleaned up, such as `submit` field and fake data submitted by malicious users.

It would be very handy to just get the validated data against the rules and be able to do `$id = $this->blog_m->insert(form_valid_data());`. 

To achieve the above, I run through the postdata against the already defined form validation rule after the validation rules have passed. code as following:

`libraries/MY_Form_validation.php`

``` php
class MY_Form_validation extends CI_Form_validation {
    public function get_valid_data() {
        $valid_data = array();
        foreach ($this->_field_data as $k => $v ) {
            // remove [] if field name is an array.
            $k = preg_replace("/\[\]$/", "", $k);                                                                                                                                                                                             
            if ( array_key_exists($k, $_POST) ) {
                $valid_data[$k] = $v['postdata'];
            }
        }
        return $valid_data;
    }
}
```

`helpers/MY_form_helper.php`

``` php
if ( ! function_exists('form_valid_data')) {
    function form_valid_data() {
        if (FALSE === ($OBJ =& _get_validation_object())) {
            return '';
        }

        return $OBJ->get_valid_data();
    }
}
```

A big plus for me is that I normally have the form filed name matched to the database column name respectively so the form data can be used right away in the model. 

Note that the `form_valid_data()` only contains form fields that have validation rules defined ( I assume you check **everything** submitted by the user :)

I find this little addon very handy and maybe worth to be included in the core.

zoaked:
I found this feature to be incredibly useful so I will go ahead and make a PR for this feature.

yoshida-mediba:
````
    if (!$this->input->post()) {
        $data = $this->session->userdata('form');
        $this->form_validation->set_data($data);
    }

    if ($this->form_validation->run()) {
        // I want $this->form_validation->get_valid_data();
    }
````

[...]

PR #3992, opened 2015-07-22
Form Validation: Add native rule valid_mac

yinlianwei:
MAC validation is very common.
php version gt 5.5.0 use  filter_var others use  regular expressions

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -568,6 +568,31 @@ class CI_Input {
 		return (bool) filter_var($ip, FILTER_VALIDATE_IP, $which);
 	}
 
+	// -------------------------------------------------------------------
+
+	/**
+	 * Validate MAC Adress
+	 *
+	 * @param $mac string	$mac MAC address
+	 * @return bool
+	 */
+	public function valid_mac($mac)
+	{
+		if(empty($mac))
+		{
+			return FALSE;
+		}
+
+		if(version_compare(PHP_VERSION, '5.5.0', ">="))
+		{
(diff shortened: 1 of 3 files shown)


narfbg:
I'd argue that it's not *that* common, not in web applications at least, but sure - it'd be handy.

However, it has no place in the `CI_Input` class and should be in `CI_Form_validation` instead. Also, your regular expression doesn't accept the "0123.4567.89ab" format, which *is* accepted by `FILTER_VALIDATE_MAC`. A changelog entry would be necessary as well.

If you don't care about attribution in the git log, I can commit the improved version myself, but otherwise you'd have to address the issues I described above.

yinlianwei:
I think there is a wider web application usage scenario, at least in my projects, often used mac address check. I suggest that the increase of the function, and modify regular expressions.

narfbg:
Arguing about the use cases is a moot point ... ignore that.

I don't understand what is it that you're suggesting ... you didn't really address any of the issues that I mentioned.

mixinix:
How about validation 'alpha_spaces' for user fullname...?
Like 'King Mataram III' not 'King Mataram 3'.

narfbg:
How about not [commenting on issues irrelevant to your question](https://github.com/bcit-ci/CodeIgniter/issues/2436#issuecomment-213282744) and [searching for similar issues first](https://github.com/bcit-ci/CodeIgniter/pull/3815)?

mixinix:
My bad, thanks.

PR #3993, opened 2015-07-22
Removing use of ->num_rows() in DB results docs.

lonnieezell:
Since some drivers must prefetch the entire result set and count the results, this practice should not be encouraged in the user guide since: 

- `result()`, `result_array()`, and `result_object()` will return an empty array with no results
- `row`, `row_array()` and `row_object()` will return NULL if no result.

### DIFF
--- a/user_guide_src/source/database/results.rst
+++ b/user_guide_src/source/database/results.rst
@@ -33,16 +33,14 @@ If you run queries that might **not** produce a result, you are
 encouraged to test the result first::
 
 	$query = $this->db->query("YOUR QUERY");
-	
-	if ($query->num_rows() > 0)
+
+	foreach ($query->result() as $row)
 	{
-		foreach ($query->result() as $row)
-		{
-			echo $row->title;
-			echo $row->name;
-			echo $row->body;
-		}
+		echo $row->title;
+		echo $row->name;
+		echo $row->body;
 	}
+
 
 You can also pass a string to result() which represents a class to
 instantiate for each result object (note: this class must be loaded)
@@ -83,11 +81,11 @@ one row, it returns only the first row. The result is returned as an
 **object**. Here's a usage example::
 
 	$query = $this->db->query("YOUR QUERY");
-	
-	if ($query->num_rows() > 0)
+
+	$row = $query->row();
+
+	if (is_object($row))
 	{
-		$row = $query->row();
-		
 		echo $row->title;
 		echo $row->name;
 		echo $row->body;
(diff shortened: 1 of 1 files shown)


user_guide_src/source/database/results.rst
```diff
@@ -33,17 +33,15 @@ If you run queries that might **not** produce a result, you are
 encouraged to test the result first::
 
 	$query = $this->db->query("YOUR QUERY");
-	
-	if ($query->num_rows() > 0)
+
```
narfbg: You're inserting a lone tab here ... :)

user_guide_src/source/database/results.rst
```diff
@@ -33,17 +33,15 @@ If you run queries that might **not** produce a result, you are
...
-			echo $row->title;
-			echo $row->name;
-			echo $row->body;
-		}
+		echo $row->title;
+		echo $row->name;
+		echo $row->body;
 	}
 
+
```
narfbg: And an unnecessary empty line here, though I don't really mind it. :)

narfbg:
Um ... you've removed the irrelevant little thing out of 2. :)

user_guide_src/source/database/results.rst
```diff
@@ -83,11 +80,11 @@ one row, it returns only the first row. The result is returned as an
 **object**. Here's a usage example::
 
 	$query = $this->db->query("YOUR QUERY");
-	
-	if ($query->num_rows() > 0)
+
+	$row = $query->row();
+
+	if (is_object($row))
```
narfbg: Also, `isset()` here and instead of the `is_array()` check on line 116 ... Not that this is wrong, `isset()` is just better. :)

lonnieezell: I'm fine with that but truly curious - how is `isset()` better? The results will be either an array or NULL so was checking for the specific cast that I wanted. Always trying to learn more so didn't know if there is a subtlety to how they work that makes it better...or if that's just your personal preference.

user_guide_src/source/database/results.rst
lonnieezell: There's no tab here in the file. This is showing a tab being removed and just a blank line put in its place. Am I missing something?

user_guide_src/source/database/results.rst
narfbg: Well, it's also partly my personal preference (as I already said, it's not wrong as is), but `isset()` is juuust a bit faster because it's a language construct and not a function (no function call overhead).

... and it makes for more consistent examples within this document, but that's a stupid argument. :)

user_guide_src/source/database/results.rst
narfbg: Hmm ... you're right, the tab is in another (very similar) example that you're not modifying. Sorry about that.

PR #3994, opened 2015-07-23
IP checking false positives and no IPv6 check

rajatsharma94:
The currently implemented method marks all IPs between 0.0.0.0 - 999.999.999.999 as valid IP Address. Which generates false positives as any IP after 255.255.255.255 is not a valid IP address.

Also, there is no check for IPv6 IP addresses.

filter_var() solves both the issues.

### DIFF
--- a/system/helpers/path_helper.php
+++ b/system/helpers/path_helper.php
@@ -61,7 +61,7 @@ if ( ! function_exists('set_realpath'))
 	function set_realpath($path, $check_existance = FALSE)
 	{
 		// Security check to make sure the path is NOT a URL. No remote file inclusion!
-		if (preg_match('#^(http:\/\/|https:\/\/|www\.|ftp|[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})#i', $path) || ( function_exists('fsockopen') &&  @fsockopen($path, 80, $errno, $errstr, 30)))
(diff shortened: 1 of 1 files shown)


system/helpers/path_helper.php
```diff
@@ -61,7 +61,7 @@
 	function set_realpath($path, $check_existance = FALSE)
 	{
 		// Security check to make sure the path is NOT a URL. No remote file inclusion!
-		if (preg_match('#^(http:\/\/|https:\/\/|www\.|ftp|[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})#i', $path))
+		if (preg_match('#^(http:\/\/|https:\/\/|www\.|ftp)#i', $path) || (!filter_var($ip, FILTER_VALIDATE_IP) === false))
```
narfbg: There's no `$ip` variable in this function and you're violating our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html) on three occasions:

 - we don't use the `||` operator in CI
 - we put spaces around `!` (which you should remove anyway because it negates the `=== false` condition)
 - we write NULL, TRUE, FALSE in all-capitals (yes, I'm aware of the constants issue, but that's just how it is right now ...)

Also, no parenthesis are necessary.

rajatsharma94:
Oh, I made the pull request on the wrong code. Sorry, I'll close this request and will create a new pull request. :)

PR #3995, opened 2015-07-23
IP checking false positives and no IPv6 check

rajatsharma94:
The currently implemented method marks all IPs between 0.0.0.0 - 999.999.999.999 as valid IP Address. Which generates false positives as any IP after 255.255.255.255 is not a valid IP address.

Also, there is no check for IPv6 IP addresses.

filter_var() solves both the issues.

### DIFF
--- a/system/helpers/path_helper.php
+++ b/system/helpers/path_helper.php
@@ -61,7 +61,7 @@ if ( ! function_exists('set_realpath'))
 	function set_realpath($path, $check_existance = FALSE)
 	{
 		// Security check to make sure the path is NOT a URL. No remote file inclusion!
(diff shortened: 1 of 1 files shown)


system/helpers/path_helper.php
```diff
@@ -61,7 +61,7 @@
 	function set_realpath($path, $check_existance = FALSE)
 	{
 		// Security check to make sure the path is NOT a URL. No remote file inclusion!
-		if (preg_match('#^(http:\/\/|https:\/\/|www\.|ftp|[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})#i', $path))
+		if (preg_match('#^(http:\/\/|https:\/\/|www\.|ftp)#i', $path) OR ( ! filter_var($path, FILTER_VALIDATE_IP) === FALSE))
```
narfbg: OR filter_var($path, FILTER_VALIDATE_IP) === $path

Way simpler. :)

Don't open a new PR ... you're still submitting all the commits.

rajatsharma94: Yeah, didn't think about it in this way.

PR #4004, opened 2015-07-27
parse_url of a valid url was returning FALSE when a colon existed in the path or query

miex0r:
MJZ: i changed this line on 7-27-15 to properly parse a url. It was erroring out if the path or query had a colon in the string. The function was assuming it's a port number following and bombed out... so i added the protocol and port to the front of the string

this string returns FALSE because the parse gets confused "/932?numeric_filters_name[]=Number+of+Members+Holding&numeric_filters_value[]=35:69"

### DIFF
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -201,7 +201,7 @@ class CI_URI {
 			return '';
 		}
 
-		$uri = parse_url($_SERVER['REQUEST_URI']);
+		$uri = parse_url($_SERVER['SERVER_NAME'] . ':' . $_SERVER['SERVER_PORT'] . $_SERVER['REQUEST_URI']);
 		$query = isset($uri['query']) ? $uri['query'] : '';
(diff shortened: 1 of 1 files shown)


narfbg:
Your commit message was quite messy, so I came up with a simpler fix instead + a changelog entry, which is necessary for bug fixes.

Thanks for the effort.

miex0r:
Works for me

narfbg:
I guess yours was better ... #4005.

miex0r:
ah! sorry about that - i should have tried out the dummy/ thing when you
made the change. Out of curiosity, how come you want to avoid using the
hostname from $_SERVER?

Thanks
Mike

narfbg:
I don't, not the hostname in particular - it's the port that is superfluous.

But there are advantages to using a literal string ... it's faster in theory (you don't access a variable), immediately obvious that the string in there doesn't matter and more easily readable.

PR #4016, opened 2015-08-03
Fixed intention from space to tab; Removed empty lines and duplicated spaces

calvintam236:


LouisMilotte:
PSR-2 all indentations should be made using space and not tab. Tabs are not universal and can equate to different measures of spaces pending the application from which they were created in, OS, etc.

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md

If opinions matter in this, I say don't make this change. Empty lines should be used to clearly define changes in logic as well.

calvintam236:
I am following the PHP style guide @ http://www.codeigniter.com/user_guide/general/styleguide.html
Most CodeIgniter files are using tab, not space.

narfbg:
@LouisMilotte CodeIgniter has never followed PSR-2 and if it's up to me alone - it never will.

@calvintam236 

Please note that indentation is not the same thing as alignment. We use tabs for indentation, but alignment can only be done with spaces.  
There's proof of that in your own patch where in some places you're replacing a single tab with two tabs or vice-versa (i.e. the *user_agents.php* file), because you have a tab-width of 4 spaces. If everybody does that, we'd have endless "alter tabs" commits going back and forth between what looks ok to one developer or another.

Also, not every occurence of two spaces is wrong, so to speak. Particularly in texts after a full stop, it means the start of a new paragraph. You also didn't take into account functional behavior and your patch would break the Typography tests.

That being said, you do have some desirable changes and I've manually applied them here: 10fb7d17b2025de4963da8b0108fda4da36ade11

calvintam236:
Many of the files I read in CodeIgniter use tab-width as 4 spaces in order to show correctly. If you are saying the code should use tab width of 2 spaces, you have to update a lot of files to match your current standard.. (I am okay using spaces for alignment. Some of the alignment e.g. note I saw are using tab instead of space so that's why I am matching the style)

narfbg:
I don't know where you got that 2-spaces tab width guess from ... I for one use 8.

But that's the whole point - if something appears uneven with a different tab width, then it's alignment; not indentation and has to use spaces instead. To give an example:

```php

	public function foo()
	{
		$var = 'a static prefix '
		       // The following line (and this comment) is indented
		       // with tabs up to the start of $var above (two tabs)
		       // and aligned with spaces up to the '=' operator
		       .something_dynamic();
	}

```

Yes, you will see many occurences of tabs being used for alignment, but those are old and gradually replaced when that particular piece of code needs to be modified.

PR #4017, opened 2015-08-03
Update intent to tab to match style with pervious standard

calvintam236:


narfbg:
This is alignment, not indentation.

And line 56, where you're adding another tab, is a perfect example of why *alignment* is done with spaces. That line is perfectly fine for me, yet you're modifying it because it appears shorter to you.

Then you're replacing perfectly well-aligned spaces with tabs ... for some kind of consistency I guess. Don't, leave it as it is. Whitespace is our last problem, we don't need mass-updates to alter it.

PR #4022, opened 2015-08-04
check if the $select variable is a array

Jheysoon:
$this->db->select();
can also accepts not just string but also a array


narfbg:
It already accepts arrays, and you're just adding an `is_array()` check for something that's *always* an array.

Jheysoon:
Yeah ..  but the $select can accept other data type

narfbg:
Yes, technically it does and although I don't see the reason why anybody would do that, your patch would just hide their error and leave them wondering why something's not working.

Issue #4023, opened 2015-08-04
ReDoS (Regular Expression Denial of Service) / Code Injection Risk


Calico90:
The second parameter `$separator` in function `increment_string` in file `system/helpers/string_helper.php` is not properly escaped for use in regular expressions. 

Nevertheless the parameter is injected directly into PHP's `preg_match` function on line 256.
`preg_match('/(.+)'.$separator.'([0-9]+)$/', $str, $match);`

This can cause a Denial of Service (ReDoS) and makes it possible to inject code into the regular expression. When the developer used this `increment_string` with a dynamic (user customizable) string and this string is not properly escaped, code injection is possible and a potentially ReDoS can be caused.

I would recommend to use PHP's `preg_quote` function to mitigate this vulnerability. 
Line 256 in file `system/helpers/string_helper.php` should be:
`preg_match('/(.+)' . preg_quote($separator) . '([0-9]+)$/', $str, $match);`

Also without the use of `preg_quote` it's not possible to call this `increment_string` function with one of the following delimiters ` . \ + * ? [ ^ ] $ ( ) { } = ! < > | : -`

----

This vulnerability seem to exists since CodeIgniter versions 2.1.0 and higher and also applies to stable releases 2.2.0 and 3.0.0.

*When this fix is pushed to the next stable release, I'll request a CVE number for this vulnerability and assign it to this issue on GitHub.* Also I will prepare a pull request for this fix.

*Generic information about ReDoS can be found here:*
- https://en.wikipedia.org/wiki/ReDoS
- https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS

Calico90:
I just submit the pull-request here: https://github.com/bcit-ci/CodeIgniter/pull/4024

narfbg:
@Calico90 

I liked your initial approach in #4020 and appreciate your enthusiasm, so I don't want to appear condescending here, but you're now showing something I very much dislike in (some) security researchers - you're too eager to label anything a vulnerability.

I do consider this a valid bug, most notably because the dash character is a hugely popular separator, but this function (or its separator parameter, to be more specific) was never intended to be used with user input, let alone without validation. Saying that this is any kind of a vulnerability is like saying that `preg_match()` itself is vulnerable and needs a CVE ID assigned ...
Sure, in the hands of a complete idiot, anything can lead to a security flaw, but we assume that our users aren't *that* unreasonable.

[...]

PR #4024, opened 2015-08-04
Fix ReDoS-bug in string_helper.php

Calico90:
Fix for ReDoS (Regular Expression Denial of Service) / Code Injection Risk as described here: 
https://github.com/bcit-ci/CodeIgniter/issues/4023

### DIFF
--- a/system/helpers/string_helper.php
+++ b/system/helpers/string_helper.php
@@ -242,7 +242,7 @@ if ( ! function_exists('increment_string'))
 	 */
 	function increment_string($str, $separator = '_', $first = 1)
 	{
-		preg_match('/(.+)'.$separator.'([0-9]+)$/', $str, $match);
+		preg_match('/(.+)' . preg_quote($separator) . '([0-9]+)$/', $str, $match);
 		return isset($match[2]) ? $match[1].$separator.($match[2] + 1) : $str.$separator.$first;
 	}
 }
--- a/system/helpers/string_helper.php
+++ b/system/helpers/string_helper.php
@@ -242,8 +242,8 @@ if ( ! function_exists('increment_string'))
 	 */
 	function increment_string($str, $separator = '_', $first = 1)
 	{
-		preg_match('/(.+)' . preg_quote($separator) . '([0-9]+)$/', $str, $match);
-		return isset($match[2]) ? $match[1].$separator.($match[2] + 1) : $str.$separator.$first;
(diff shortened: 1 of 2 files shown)


system/helpers/string_helper.php
```diff
@@ -242,7 +242,7 @@ function random_string($type = 'alnum', $len = 8)
 	 */
 	function increment_string($str, $separator = '_', $first = 1)
 	{
-		preg_match('/(.+)'.$separator.'([0-9]+)$/', $str, $match);
+		preg_match('/(.+)' . preg_quote($separator) . '([0-9]+)$/', $str, $match);
```
narfbg: Should be `preg_quote($separator, '/')` instead, to account for the specific delimiter. Also, please remove the spaces around concatenation.

narfbg:
This would also need a changelog entry, but please read my comment on #4023 before marking it as a security bugfix. :)

Calico90:
I'll prepare a new PR. Good point about the `preg_quote` delimiter.

narfbg:
No need, just push the changes to the same branch. Almost every PR goes through some changes, we'd have 40000 of them if they were resubmitted anew. :)

Calico90:
Check, just a branch conflict now. Can someone (with write access) fix it? Thanks

narfbg:
We can't fix conflicts, only you can.

This conflict appeared because you've based you patch on a very, very old version of our develop branch (what you have on line 127, we have at 208 - that's a lot of changes).  
In order to resolve the conflict, you'd neet to do a hard reset back to the last commit that you've gotten from our repository, then pull all of our changes since and commit your changes again.

... or delete your fork and re-fork our repo, in which case you probably would need to open a new pull request. :)

Nevertheless, you're now using `preg_quote()` not just inside the regular expression, but also in the return value, which shouln't happen. And as per my commens in #4023, I still insist against labeling this as a vulnerability (which in fact you're doing twice).

If you don't mind, I'd commit the fix & changelog message myself. Unless you delete your fork, I'll cherry-pick your first commit so you'd be properly attributed in the git log.

Calico90:
@narfbg I don't mind if you commit the fix. My bad about the return, I misread the part before else `:` as if it was the condition of the if statement, not the returning value. But it's obvious a shorthand if statement. Thanks! Won't label it as a security vulnerability.

narfbg:
Done, thanks. :)

Calico90:
@narfbg Good call! This will clear things up! :)

PR #4028, opened 2015-08-05
Reduce once $config['query_toggle_count'] checking

bjjay:
This checking can be done by calling set_sections method when initialize profiler .

### DIFF
--- a/system/libraries/Profiler.php
+++ b/system/libraries/Profiler.php
@@ -100,12 +100,6 @@ class CI_Profiler {
 		$this->CI =& get_instance();
 		$this->CI->load->language('profiler');
 
-		if (isset($config['query_toggle_count']))
-		{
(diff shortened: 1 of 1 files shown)


narfbg:
You might as well remove the foreach loop too then.

bjjay:
This key isn't included in the array.

narfbg:
What has that got to do with removing the loop?

bjjay:
The loop is for $_available_sections, and the key isn't in $_available_sections but can be in $config. What's wrong with this?

narfbg:
I'm not saying there's anything wrong with this - it's completely fine. I'm saying you can *remove* the loop, because it's also handled by `set_sections()`.

bjjay:
The loop is setting default sections to display.

narfbg:
Head -> desk.

bjjay:
If the sections are defined in config file, this loop can be removed.

PR #4030, opened 2015-08-05
Update welcome_message.php

DiederikLascaris:
I have changed the styling of the default welcomes page.  It now uses the same styling as the website of codeigniter.com.

I have included external links to the logo and font files. It links to rawgit, it's a CDN service for github files. Although I don’t like a dependency to an external service I feel in this case it's a valid choice since I don't want to bloat up the installation size by including a bunch more files just to fancy up the installation file. 
If the CDN is down it still looks good enough. Only the logo is missing and the heading text is Helvetica again.

One thing I was not quite sure about was using the heading font this way. Perhaps this is screwing to much with the corporate identity as the term "welcome to" is inserted between the fire image and the text part of the logo.

Perhaps an alternative would something like just have the flame and text logo in the red bar and put the "Welcome to CodeIgniter" text below the red bar and use a regular heading style on it (dark bold text).

Anyway, here is a screenshot of the result:

selection_035


narfbg:
External assets are a blocker.

Any styling changes must be achievable via inline CSS alone (yes, this means no images and no custom fonts), and they must also be applied to default error templates.

DiederikLascaris:
I understand. Is an inline base64 image also a dealbreaker?

narfbg:
No, as long as it's inline - that's ok. Although it might be tricky to mask it on browsers that don't support Base64-encoded images, if any.

But unlike with other stylings, the logo (and any reference to CI itself) must stay only on the Welcome page. Obviously, the error templates will still have recognizable styling, but other than that, they shouldn't leak data about the platform.

DiederikLascaris:
Ok, I will give it another go if I have another spare half hour in the next couple of days. A quick search tells me inline images are supported widely (from IE8). 

http://caniuse.com/#search=data

DiederikLascaris:
Ok I've had some to time to make the ajustments. I've created an inline image for the logo in the welcome page. Changed the error messages as you suggested to match the same color schema.

Welcome page:
ci4-welcome

404 error:
ci4-error-404

DB error:
ci4-error-database

Show_error example:
ci4-error-show_error

narfbg:
Looks better (the images in your comment at least).

But please, revert the *coding* style changes ... every second line is an empty one and you're replacing tabs indentation with 4 spaces instead - we don't do that. And more importantly, it's really hard to review and it shouldn't be - a few CSS property changes shouldn't result in a 100-line diff.

DiederikLascaris:
I've made the coding styles changes like you suggested. Still contains alot of diff rules because I added a wrapper around the main container, so I needed to indent the original code.

narfbg:
@DiederikLascaris Why do you need that wrapper?

@jim-parry What do you think about this one? Should it go into 3.0.x or 3.1?

jim-parry:
I see no problem including it in 3.0.x

________________________________

softwarespot:
I see a couple of issues with this PR:

- Inconsistency with hex colours. Use either upper-case or lower-case chars for A-F, not both.
- You honestly could do away with wrapper and use the body if you're wanting to centre (which is what I presume you're trying to do)

narfbg:
This got lost under the radar ... sorry.

@DiederikLascaris If you're still interested, do what @softwarespot suggested and I'll be happy to merge it.

Ardhie1032:
Interesting...

ProfNandaa:
@DiederikLascaris if you are not in position to make this update now, I can do it -- let me know If I can go ahead.

ProfNandaa:
@narfbg -- I just did, feel free to close it if you @DiederikLascaris makes the same changes on this one.

ProfNandaa:
@narfbg @jim-parry - anything else that should be addressed on the PR #4973 ?

jim-parry:
I am fine with this :)

DiederikLascaris:
@narfbg thanks for the reminder on this subject.

I have removed the wrapper and changed the hex colors to lowercase, also changed the color name 'white' to #fff to be more consistent.

narfbg:
Thanks!

PR #4031, opened 2015-08-05
proposal on not to escape output

avenirer:
If someone doesn't want to escape output, can set a forth parameter for form_input() and form_textarea() to TRUE.


narfbg:
Why?

avenirer:
Sorry... I must be drunk...

PR #4041, opened 2015-08-10
remove sess_expire_on_close

Jheysoon:
not found sess_expire_on_close in config.php


narfbg:
It's a fallback mechanism, there are other items that you won't find in the config too.

PR #4044, opened 2015-08-11
Catch redis exception

rochefort:
RedisException occured when connected with ssl port.

 ```
An uncaught Exception was encountered

Type: RedisException

Message: read error on connection

Filename: /Users/trsw/work/php/codeigniter/ci-rest/vendor/codeigniter/framework/system/libraries/Cache/drivers/Cache_redis.php

Line Number: 126
```

There is a possiblity of RedisException of `$this->_redis->auth()`.
Phpredis does not support ssl port, so the above exception occurs when I try to connect to the SSL port.


narfbg:
Applied a different patch due to #4066.

PR #3387, opened 2014-12-04
Update CodeIgniter.php

ghost:
Add else at line 69

PR #4045, opened 2015-08-12
Update CodeIgniter.php

hedcler:
Fixes bug when the configuration is separated into directories.


narfbg:
Duplicate of #3387 
Resolved here: 0d60a21187e2611cc4cf0ef2d7322c9b91c002b9 (referenced above)

PR #4050, opened 2015-08-13
Update Input.php

jpwdesigns:
We encountered a problem with tracking cookies which included the % character in the name. These cookies were not generated by our site but by a third party tracking company. After doing some searching I find that this is not the first time someone ran into this issue.

It is unreasonable to break the entire site with error output simply because a % character shows up in a cookie name. Especially for enterprise sites that have numerous tracking cookies set by third party applications.  

What would be reasonable is to do just about anything else besides hard fail.  This _clean_input_keys function should be more forgiving.

I've added % and ~ to the allowed chars regex. However a better solution would be not to hard fail ever. Rather just ignore that cookie in the application.

For reference:

https://en.wikipedia.org/wiki/HTTP_cookie#Cookie_attributes
The value of a cookie may consist of any printable ASCII character (! through ~, unicode \u0021through \u007E) excluding , and ; and excluding whitespace. The name of a cookie excludes the same characters, as well as =, since that is the delimiter between the name and value. The cookie standard RFC 2965 is more limiting but not implemented by browsers.

To recreate this problem do this:
* Go to a site running CI such as ExpressionEngine etc...
* Open your console
* Enter `document.cookie="asdf%asf=Goodbye world";`
* Refresh page
* Sigh

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -724,7 +724,7 @@ class CI_Input {
 	 */
 	protected function _clean_input_keys($str, $fatal = TRUE)
 	{
-		if ( ! preg_match('/^[a-z0-9:_\/|-]+$/i', $str))
+		if ( ! preg_match('/^~%[a-z0-9:_\/|-]+$/i', $str))
 		{
 			if ($fatal === TRUE)
 			{
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -724,7 +724,7 @@ class CI_Input {
 	 */
 	protected function _clean_input_keys($str, $fatal = TRUE)
 	{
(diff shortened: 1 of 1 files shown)


mwhitneysdsu:
Are you sure this was on CI 3? I've seen [a similar problem reported on CI 2](https://github.com/ci-bonfire/Bonfire/issues/1072), but the person who reported the issue also reported that a modification to the /core/Input.php with the CI3 version of the method(s) cleared it up. I also can't reproduce the problem on my own site (on CI 3.0.1) with the steps you reported.

Additionally, the `_clean_input_keys()` method doesn't just apply to cookies...

jpwdesigns:
@mwhitneysdsu Yeah you're right. This is happening on a version of EE that has CI2 under the hood.

PR #4054, opened 2015-08-14
Fix for HTTP_RAW_POST_DATA on newer PHP; it is often not populated.

willbuckner:
PHP 5.6 complains about deprecation if `$HTTP_RAW_POST_DATA` is used. Fix to use `php://input` if `$HTTP_RAW_POST_DATA` is unset. If `$HTTP_RAW_POST_DATA` is set, the old behavior is used, so this should have no impact if it's set. If unset, the current code would be broken as `$HTTP_RAW_POST_DATA`  is not defined, and now it will work. Thanks!

### DIFF
--- a/system/libraries/Xmlrpcs.php
+++ b/system/libraries/Xmlrpcs.php
@@ -174,15 +174,20 @@ class CI_Xmlrpcs extends CI_Xmlrpc
 	 */
 	function parseRequest($data='')
 	{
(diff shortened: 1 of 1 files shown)


narfbg:
CI2 is in security maintenance mode until October and no longer receives other bugfixes, sorry.

https://github.com/bcit-ci/CodeIgniter/blob/2.2-stable/README.md

PR #4070, opened 2015-08-23
Allow abstract class and namespacing in models and libraries

adamos42:
The loader will allow abstract class load in model and library, but because of abstract you not able to use with ``$this->model/library_name`` instantiate.

Im using this to make a base model or library where the models or libraries can ``override`` the base methods if is loaded, if you make a magick method in base to reach the database methods there, then you can limit the available methods on frontend and override on backend or override the base delete if you don't want ``delete`` from a specific table just ``update`` a ``deleted`` field.

Abstract classes will work if ``ReflectionClass`` is available (``>=PHP5.3``), if not  available then the code working as before.

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -320,7 +320,22 @@ class CI_Loader {
 		}
 
 		$this->_ci_models[] = $name;
-		$CI->$name = new $model();
+		
+		// If available the ReflectionClass (>=PHP5.3)
+		if(class_exists('ReflectionClass'))
+		{
+			// Then check the class is abstract
+			// Because if it is then cannot instantiate
+			$class = new ReflectionClass($model);
+			$abstract = $class->isAbstract();
+		}
+		
+		if(!class_exists('ReflectionClass') || !$abstract)
+		{
+			// Instantiate the class
+			$CI->$name = new $model();
+		}
+		
 		return $this;
 	}
 
@@ -1240,11 +1255,23 @@ class CI_Loader {
 
 		// Save the class name and object name
 		$this->_ci_classes[$object_name] = $class;
-
-		// Instantiate the class
-		$CI->$object_name = isset($config)
-			? new $class_name($config)
-			: new $class_name();
+		
+		// If available the ReflectionClass (>=PHP5.3)
+		if(class_exists('ReflectionClass'))
+		{
+			// Then check the class is abstract
+			// Because if it is then cannot instantiate
+			$class = new ReflectionClass($class_name);
+			$abstract = $class->isAbstract();
+		}
+		
+		if(!class_exists('ReflectionClass') || !$abstract)
+		{
+			// Instantiate the class
+			$CI->$object_name = isset($config)
+				? new $class_name($config)
+				: new $class_name();
+		}
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


adamos42:
Also adding namespace support for models and libraries. If namespaced class available which generated from the path, then that will be added as an instantiate.
Example for model: 
```
<?php
// Namespace from the path: ./application/models/data/
namespace Model\Data;

// Namespaced class extends the root CI_Model
class Content extends \CI_Model
{
    [...]
}
```

narfbg:
No ... the loader methods are supposed to instantiate. If you just need to include something, use `include()`.

adamos42:
@narfbg Where are this rule, on documentation? Simple include not checking the file and the class existance (and not adding to the loaded classses to check duplications), and a loader should able to load valid classes it's not my fault that php can't instantiate abstract class and a FRAMEWORK can't handle it. But Ok, I will remove that commit and always override it with a custom loader.

BUT!

I want to know, why you closing in instant? This pull request contains 2 commit, the second adding PSR valid class loading (without _ converting) allowing to the developers to load namespaced models and libraries if they wanted to programing in OOP in 2015. The namespaces are in since 5.3.0 and the current PHP stable version is 5.6.12. I know the framework must be compatible with 5.3.3, so I don't know why you not allow a 5.3.x feature. Codeigniter does not want to evolve?

You just close a pull request if you found something what you don't like, and you not reading the complete request? In that case I will add one feature/bugfix per pull request.

jim-parry:
1) pull requests should only address one issue or topic; that is mentioned in the "contributing" docs
2) CI doesn't support abstract classes; that may change with CI4. This issue has been addressed numerous times on the forums, with multiple suggested workarounds.
3) Namespace support is planned for CI4. There is nothing wrong with developers using namespaces in their own code, but we don't plan to add that to the CI3 core.
4) CI3 targets a minimum PHP version of 5.2.4, which does not support namespaces. We recommend 5.4 or better.
5) The docs do not explicitly spell it out, but the loader methods attempt to instantiate an instance of the class with the name implied by the source; they imply this by suggesting that you can assign the resulting *object* to a different variable, whose name is passed as a second parameter to the relevant loader method.

adamos42:
@jim-parry Thanks the reasons, sorry for my comment. Where can I join to the CI4 development? Or that is a dream for a few year?
I did not found the "only address one issue or topic" on the documentation or in the contributing.md
My changes for the namespace are compatible with 5.2, see on the travis reports, if you not want use then you not use, it just changing the class if namespaced is exists after include.
Yes the instantiates are working as it before, only the php will check the abstraction with reflection class (if available), and if is abstract then it will load it. If you not want to add into CI3 that's ok, but least we can handle with a nice codeigniter error message and the developer will not trying to solve a problem in the system (because the standard error said the error is there).

jim-parry:
@adamos42 The CI4 development is just getting under way ... http://forum.codeigniter.com/thread-62615.html
We are getting some core bits together, and the CI4 repository organized, before opening things up to the community ... not sure when that will happen, exactly, but it will :)
CI3 will end up being the "legacy" version of CodeIgniter down the road, and we are trying to keep its codebase solid and consistent ... there are probably a few "modern" PHP practices that might be incorporated, but the "good stuff" should end up in CI4.

PR #4072, opened 2015-08-24
Custom Tab Width

CodeIgniter:


narfbg:
What does this do and why do we need it?

CodeIgniter:
Because the current tab width is 8 spaces, it is possible to customize the better.

narfbg:
You're saying nothing with that ...

CodeIgniter:
OK. But CodeIgniter 3.0 Chinese user guide change the width of the 4 spaces.
It is a small improvement to the user guide project.

narfbg:
There is no official Chinese user guide and nothing in the Sphinx config is intended for external use. 

And I still don't understand what does this actually improve ... please explain it in detail.

PR #4075, opened 2015-08-24
Deprecated code in develop branch

vkeranov:
As 3.1.0 is on the road, maybe we should remove these...


jim-parry:
if travis-ci is failing, is there more to do this PR before it can be merged?

narfbg:
The builds are failing because it's only functionality that's being removed here, without the respective test cases.

But that's not why I'm not merging it ... 3.1 isn't imminent and while the docs say that these will be removed in '3.1+', it's not necessarily '3.1'. It's unlikely, but we may still decide to avoid some BC breaks and/or revert deprecations.

vkeranov:
That's true... Some of the docs say "Depecated 3.0.0", while others say "3.1.0" or "3.1+". So you can choose what to remove, because of some BC breaks.

Issue #3752, opened 2015-04-09
Router not setting directory, class or method on time.


xeno:
When using a routing directory in index.php, the routing directory is not being set before _set_routing() is called [in the Router.php constructor method]. The result is a 404 on the default controller and full routes required (including the directory name) in routes.php.

LouisMilotte:
I have received this error if I forget to follow the naming standards. Assuming this is for CI version 3; is your default controller named like so: Controller_name.php? If not, what is the setting for your default controller and what is the name of your default controller?

PR #4077, opened 2015-08-26
3.0 stable

badluck81:
This days i migrated to CI and i found a really ugly bug in router.
I tried to setup $routing['directory'] based on subdomains in index.php like this:

if (($_SERVER['HTTP_HOST']=="www.test.com")  || (($_SERVER['HTTP_HOST']=="test.com")) ){
	$routing['directory'] = 'frontend';
}elseif ( $_SERVER['HTTP_HOST']=="backend.test.com" ){
	$routing['directory'] = 'backend';
}

And here started ugly problems, because in system/core/Router.php the directory is not set first and 
$this->_set_routing(); (in Router.php construct) call _set_default_controller() without calling set_directory() first.
So basically, we cannot setup defalt routes while working with subdomains (with folders works).

The hack i did is in MY_Router.php:
<?php
if (!defined("BASEPATH")) exit("No direct script access allowed");
class MY_Router extends CI_Router {
    function __construct($routing) {
		if (is_array($routing) && isset($routing['directory'])) {
				$this->set_directory($routing['directory']);
		}
		parent::__construct();
    }
}
And default routes worked like a charm.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -803,7 +803,7 @@ class CI_Input {
 
 		if ( ! isset($headers))
 		{
-			empty($this->headers) OR $this->request_headers();
+			empty($this->headers) && $this->request_headers();
 			foreach ($this->headers as $key => $value)
 			{
 				$headers[strtolower($key)] = $value;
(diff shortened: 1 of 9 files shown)


narfbg:
I don't understand why this is a pull request (and with changes unrelated to the described issue), but nevertheless, the bug that you're describing is already fixed in version 3.0.1.

See issue #3752.

PR #4078, opened 2015-08-26
Update Lang.php

sv3tli0:
When multiple lang files are set to load there is no logical return just Null.
This simple update can provide information and to support return result using each lang file as key for 2 dimensional return array.

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -89,12 +89,15 @@ class CI_Lang {
 	{
 		if (is_array($langfile))
 		{
+			$results = array();
 			foreach ($langfile as $value)
 			{
-				$this->load($value, $idiom, $return, $add_suffix, $alt_path);
+				$results[$value] = $this->load($value, $idiom, $return, $add_suffix, $alt_path);
 			}
-
-			return;
+			if($return){
+				return $results;
+			}
+			return TRUE;
 		}
 
 		$langfile = str_replace('.php', '', $langfile);


jim-parry:
This could be an improvement, but the user guide would need to be updated too.

At present, the $return parameter to load(...) suggests that a true value will cause the loaded array of translations to be returned, which is indeed not the case when an array of language files is passed to it.

If multiple files are loaded, the current docs would suggest that $this->language (the array of translations) be returned, whereas sv3tli0's suggestion would have $this->is_loaded (the array of loaded language files) returned.

I think that a better fix would be to return $this->language in the case of an array passes as $langfile, and to return TRUE otherwise.

sv3tli0:
My idea was to change the current `return ;`  because its returning NULL no matter what happens and this is not handful.
I think that it would be good if language loader is returning the same output for multiple entries as it does for 1. 

When there is set flag to Return entries, the loaded languages should be in 2nd dimensional array as there are multiple lang files loaded (and there can be same keys used). Thats why I purpose to use lang file names for keys in that return array. 
Another case which I haven't covered is when some of the keys inside that language files is either invalid or already loaded. Should it return TRUE even if not all of the files are loaded normally?

jim-parry:
Hmmm - if $return is TRUE, then loading a single language file returns the associative array of language settings. If $return is FALSE, then the settings in the specified language file are *added* to the $language property of the Lang object, and the method returns TRUE.

In the case of an array of language files, if the $return parameter is FALSE, then each of the specified language files is loaded in turn, with all of the settings added to the $language property. A later translation with the same key replaces a previously loaded one. This is contrary to your interpretation.

In the case of an array of language files, if the $return parameter is TRUE, that is passed on to the individual invocations of the load method; each will then return the translations for each file as the file is processed, but those are ignored (thrown away, effectively), inside the loop that processes the array of filenames. I don't see that any return value makes sense for this, hence returning NULL (the current behaviour) does not seem inappropriate.

In the first case, an array of filenames, there is actually a single array of translations built, and it might make sense to return that, except that it isexplicitly contrary to the expected behaviour dictated by the $return parameter. Again, NULL does not seem inappropriate.

If you want access to the array of translations after processing a number of translation files, the $language property is public and can be accessed directly.

With multiple files loaded, the $is_loaded property will contain the filenames of those files loaded as associative array keys.

jim-parry:
The only improvement I can see reasonable is to explicitly ignore the $return parameter if an array of filenames is passed, and to return the translation array after processing all of the language files. This would have to be clarified in the user guide too.

Given the impossibility of interpreting the return value right now, if an array of filenames is passed, no one should be relying on any return value in this case, so I would not think that this would be considered a BC change.

mwhitneysdsu:
I'm just trying to see if I follow this, are you thinking something along these lines:

```php
    public function load($langfile, $idiom = '', $return = FALSE, $add_suffix = TRUE, $alt_path = '')
    {
        if (is_array($langfile)) 
        {
            foreach ($langfile as $value) 
            {
                $this->load($value, $idiom, FALSE, $add_suffix, $alt_path);
            }

            return $return ? $this->language : TRUE;
        }
```

This is similar to what I've been using for a while, though the `$this->load()` call just passed `$value`, which can be problematic if someone needs the other parameters.

jim-parry:
That looks like a very elegant improvement :)

sv3tli0:
@mwhitneysdsu returning all loaded translations is not as returning just those from the request.
Either its more clear way. 

@jim-parry I think that this overwriting of translation keys is a little bit wrong in the logic of CI but any improvement in it should be either at major version improvement as 3.1 or it can stay as it is..

jim-parry:
@sv3tli0 The overwriting of keys is the current behaviour, that's all.

@narfbg @lonnieezell What do you think of changing the return value for Lang::load() with an array of filenames, to return the set of translations, per @mwhitneysdsu ? 
I am thinking that would make sense in 3.1.0. The unit tests and user guide would need updating too.

mwhitneysdsu:
I understand that it's not the same, I was just trying to interpret what @jim-parry was saying, and comparing it to what I already use. Personally, I don't like the idea of returning two different types of array from the same method (one with filename keys and array values vs. one with the loaded key/value pairs), because you have to change the way you handle the output based on what type of input you passed in. However, that doesn't hold as much weight because the method already has 3 return types (array, TRUE, NULL) based in part on what you pass to it.

An alternative would be to do an array_merge() on the returned arrays, which would allow you to return only the loaded key/value pairs, rather than returning all of the key/value pairs which have been loaded so far. Again, my primary issue with the original proposal is simply the fact that the array has a different structure from the array returned when loading a single file.

lonnieezell:
Since the docs currently state: `Returns: Array of language lines if $return is set to TRUE, otherwise void` I think it makes perfect sense to return an array of the language lines loaded as @sv3tli0 is proposing. It follows the logic of what it is currently set to do. 

However, updated docs and changelog would need to be in the pull request also. 

There is a small bit of wasted memory in his implementation since we are creating an array that might never hold anything other than a few nulls, but it's probably not worth worrying about.

Oh, and I don't know that this needs to wait for a 3.1 since it's not really a BC break as @jim-parry has said.

narfbg:
I think this is a perfect example of parameter/feature creep ... What once has had a simple purpose, is now bloated with different options and you're struggling to even understand how they work and what the proper behavior is.

There are potential issues that would come up with this addition, due to differences in the resulting array keys depending on `$add_suffix` and it's not even considering possible error conditions ... not happening.

Also, technically, it is a BC break.

PR #4087, opened 2015-08-29
Fix Upgrading from 3.0.0 to 3.0.1

kenjis:
Some config files have been updated at 3.0.1.

See https://gist.github.com/kenjis/90523d05b451f03c5dc6


narfbg:
None of these changes are really necessary for the update ... db_debug in particular has no place in the upgrade instructions.

kenjis:
Why not necessary? They are updated.

narfbg:
Those are minor additions/changes to configuration default values, made simply for convenience. You gain close to nothing from updating them and none of them is necessary for the framework to keep functioning, which is the point of the upgrade guide.

PR #4090, opened 2015-08-31
Add method field_highlight() to Form Helper

jmb-oliveira:
Add the functionality to easily highlight fields on validation errors

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -1005,3 +1005,26 @@ if ( ! function_exists('_get_validation_object'))
 		return $return;
 	}
 }
+
+// ------------------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


jmb-oliveira:
Result

capturar

narfbg:
This is doable via a simple ternary condition, we don't need a function for it.

jmb-oliveira:
It's a helper function.

Isn't the same point as using:

```
echo form_input('name', 'value');
```

Instead of
```
<input type="name" value="value" />
```

?

narfbg:
Yes, but just because useless functions have been implemented in the past, that doesn't mean we should continue to do so.

jmb-oliveira:
That's true. I Agree.
Thx for reply.

PR #4092, opened 2015-09-01
There was an extra = sign in this file

mpmont:
Stable version has an extra = sign in the beginning of a database php file.


narfbg:
My bad, apparently. Thanks!

Issue #4056, opened 2015-08-15
A bug within CI_Input::get_request_header()


ivantcholakov:
CI3 develop branch, Input.php, line 806 https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Input.php#L806

At the moment it is wrong:
```php
			empty($this->headers) OR $this->request_headers();
```

It should be:
```php
			! empty($this->headers) OR $this->request_headers();
```

ivantcholakov:
Yes, this works too.

PR #4097, opened 2015-09-03
Fix #3890 secondary bug with 'OR'

### LINKED ISSUE #3890: get_request_header: wrong case in nginx
When i use the function get_request_header() in nginx keys appears to be in ucfirst mode.
So if i use 'auth' as a key, get_request_header does not match because the key expected is 'Auth'

gmcinalli:
Change `OR` with `AND`, because if `$this->headers` is empty the second statement is not executed.

**EXAMPLE**:
```php
public function test()
{
    $test = [];
    empty($test) AND $test = ['test'];
    print_r($test);
}

// Prints: Array ( [0] => test )

```

```php
public function test()
{
    $test = [];
    empty($test) OR $test = ['test'];
    print_r($test);
}

// Prints: Array ( )

```

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -803,7 +803,7 @@ class CI_Input {
 
 		if ( ! isset($headers))
 		{
-			empty($this->headers) OR $this->request_headers();
+			empty($this->headers) AND $this->request_headers();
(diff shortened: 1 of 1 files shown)


narfbg:
Duplicate of #4056.

PR #4099, opened 2015-09-07
Fix router logic

wenpeng:
If the class does not extend from CI_Controller, you can not access via Route


narfbg:
Controllers not extending `CI_Controller` won't work either way ... this is not necessary.

wenpeng:
It is not necessary but it will be more rigorous.

PR #4107, opened 2015-09-10
Cookie lifetime in Session bugfix

skunkbad:
Since the $expiration variable is set on line 262, there will never be a time when the ternary operator evaluation lets the cookie lifetime be set to 0, unless the config item sess_expiration was specifically set to 0.

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -267,7 +267,7 @@ class CI_Session {
 		}
 		else
 		{
-			$params['cookie_lifetime'] = ( ! isset($expiration) && config_item('sess_expire_on_close'))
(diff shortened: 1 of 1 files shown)


narfbg:
Even if you were right, there'd be nothing wrong with the current logic - there's no bug, nothing broken, nothing to fix.

And your assertion is wrong in the first place, because *sess_expiration* may not exist in the first place (i.e. the user didn't complete the upgrade instructions for 2.x to 3.x) and that's exactly the case this condition is for; *sess_expire_on_close* shouldn't be taken in to account under any other circumstances.

skunkbad:
Eh, I keep forgetting that the default value of config_item is NULL. Nevermind.

PR #4115, opened 2015-09-14
Pass query argument with '=' to QUERY_STRING

feryardiant:
Hi,

I make all CLI argument contains `=` character to `$_SERVER['QUERY_STRING']` so you can get the value using `$_GET['key']` or `$this->input->get('key')`.

Anyway, I know it's not realy nessesary but would be great if we have this feature (I guess).

Best. :grin:

### DIFF
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -284,6 +284,22 @@ class CI_URI {
 	protected function _parse_argv()
 	{
 		$args = array_slice($_SERVER['argv'], 1);
+		$get = array();
+
+		foreach ($args as $key => $value) {
+			if (strpos($value, '=') !== false) {
+				$value = ltrim($value, '-');
+				list($req, $val) = explode('=', $value);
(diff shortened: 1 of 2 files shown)


narfbg:
It's not just unnecessary, it's wrong.

feryardiant:
Let me know the reasons!

On Sep 15, 2015 15:21, Andrey Andreev <notifications@github.com> wrote:

It's not just unnecessary, it's wrong.

feryardiant:
Can't believe we have people who doing something without any reasons. Just share the thoughts man and let em know what's the "wrong" point here.

Thank you.

narfbg:
You know ... if you had been patient, you might've gotten an answer. Now you won't.

Not everybody's patches and ideas will be accepted and nobody is owed an explanation. How's that for a reason?

feryardiant:
So cold. :neutral_face:

PR #4119, opened 2015-09-16
Fix typo in Tutorial

kenjis:


jim-parry:
Verified. Picky, but correct.

PR #4124, opened 2015-09-19
Improve the language support in migration library

zhanghongyi:
In the current migration library, the language feature is not fully utilised. Three of the error messages still use strings. I suggest to add these entries into `migration_lang.php` so that we can provide more support to multiple languages. Meanwhile, languages should be loaded before any error messages showing up.

### DIFF
--- a/system/language/english/migration_lang.php
+++ b/system/language/english/migration_lang.php
@@ -37,7 +37,6 @@
  */
 defined('BASEPATH') OR exit('No direct script access allowed');
 
-$lang['migration_initilization'] = 'Migrations Class Initialized';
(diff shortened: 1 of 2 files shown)


jim-parry:
Looks good to me.

narfbg:
These 3 errors are triggered by `show_error()` (which means they're fatal and stop script execution), while the others that already have translated messages are runtime ones.

I know this doesn't seem important, but it is. `show_error()` will be replaced by exceptions in CI3.1 and using language translation in there can cause a lot of problems.

Only messages that may be seen by an end user should be translated, while fatals should only be seen by application developers, who pretty much must know English or they just won't be able to understand CodeIgniter or programming in general.

PR #4126, opened 2015-09-20
Persist config file settings when resetting form_validation

zoaked:
When checking multiple arrays using form_validation you have to call reset_validation between each separate check due to the instance of the library being a singleton.  The issue comes in when the settings are loaded from a config file as they are initially loaded from a parameter in the constructor, but are set to an empty array when resetting the class.  To get around this issue a copy of the config parameter is made and then the copy is used to reset the rules when clearing.

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -1584,14 +1584,16 @@ class CI_Form_validation {
 	 * @param	bool
 	 * @return	CI_Form_validation
 	 */
-	public function reset_validation($keep_config=false)
+	public function reset_validation($keep_config = FALSE)
 	{
 		$this->_field_data = array();
-		if( ! $keep_config)
-			$this->_config_rules = array();
 		$this->_error_array = array();
 		$this->_error_messages = array();
 		$this->error_string = '';
+		if ( ! $keep_config)
+		{
+			$this->_config_rules = array();
+		}
 		return $this;
 	}
 
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -1581,19 +1581,14 @@ class CI_Form_validation {
 	 * Prevents subsequent validation routines from being affected by the
 	 * results of any previous validation routine due to the CI singleton.
 	 *
-	 * @param	bool
 	 * @return	CI_Form_validation
 	 */
-	public function reset_validation($keep_config = FALSE)
+	public function reset_validation()
 	{
 		$this->_field_data = array();
 		$this->_error_array = array();
 		$this->_error_messages = array();
 		$this->error_string = '';
-		if ( ! $keep_config)
-		{
-			$this->_config_rules = array();
-		}
 		return $this;
 	}
 
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -68,7 +68,6 @@ class CI_Form_validation {
 	 * @var array
 	 */
 	protected $_config_rules	= array();
-	private $_default_config_rules = array();
 
 	/**
 	 * Array of validation errors
@@ -142,7 +141,7 @@ class CI_Form_validation {
 		}
 
 		// Validation rules can be stored in a config file.
-		$this->_default_config_rules = $this->_config_rules = $rules;
+		$this->_config_rules = $rules;
 
 		// Automatically load the form helper
 		$this->CI->load->helper('form');
@@ -1582,12 +1581,14 @@ class CI_Form_validation {
 	 * Prevents subsequent validation routines from being affected by the
 	 * results of any previous validation routine due to the CI singleton.
 	 *
+	 * @param	bool
 	 * @return	CI_Form_validation
 	 */
-	public function reset_validation()
+	public function reset_validation($keep_config=false)
 	{
 		$this->_field_data = array();
-		$this->_config_rules = $this->_default_config_rules;
+		if( ! $keep_config)
+			$this->_config_rules = array();
 		$this->_error_array = array();
 		$this->_error_messages = array();
 		$this->error_string = '';
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -68,6 +68,7 @@ class CI_Form_validation {
 	 * @var array
 	 */
 	protected $_config_rules	= array();
+	private $_default_config_rules = array();
 
 	/**
 	 * Array of validation errors
@@ -141,7 +142,7 @@ class CI_Form_validation {
 		}
 
 		// Validation rules can be stored in a config file.
-		$this->_config_rules = $rules;
+		$this->_default_config_rules = $this->_config_rules = $rules;
 
 		// Automatically load the form helper
 		$this->CI->load->helper('form');
@@ -1586,7 +1587,7 @@ class CI_Form_validation {
 	public function reset_validation()
 	{
 		$this->_field_data = array();
-		$this->_config_rules = array();
+		$this->_config_rules = $this->_default_config_rules;
 		$this->_error_array = array();
 		$this->_error_messages = array();
 		$this->error_string = '';
--- a/user_guide_src/source/changelog.rst
+++ b/user_guide_src/source/changelog.rst
@@ -16,13 +16,13 @@ Release Date: Not Released
 -  General Changes
 
    -  Updated the *application/config/constants.php* file to check if constants aren't already defined before doing that.
+   -  Added optional parameter to ``reset_validation()`` in :doc:`form validation<libraries/form_validation>` that stops ``_config_rules`` from being cleared.
 
 Bug fixes for 3.0.2
 -------------------
 
 -  Fixed a bug (#2284) - :doc:`Database <database/index>` method ``protect_identifiers()`` breaks when :doc:`Query Builder <database/query_builder>` isn't enabled.
 -  Fixed a bug (#4052) - :doc:`Routing <general/routing>` with anonymous functions didn't work for routes that don't use regular expressions.
--  Fixed a bug (#4126) - Added optional parameter to ``reset_validation()`` in ``Form_validation`` that stops ``_config_rules`` from being cleared.
 
 Version 3.0.1
 =============
--- a/user_guide_src/source/changelog.rst
+++ b/user_guide_src/source/changelog.rst
@@ -16,13 +16,13 @@ Release Date: Not Released
 -  General Changes
 
    -  Updated the *application/config/constants.php* file to check if constants aren't already defined before doing that.
-   -  Added optional parameter to ``reset_validation()`` in :doc:`form validation<libraries/form_validation>` that stops ``_config_rules`` from being cleared.
 
 Bug fixes for 3.0.2
 -------------------
 
 -  Fixed a bug (#2284) - :doc:`Database <database/index>` method ``protect_identifiers()`` breaks when :doc:`Query Builder <database/query_builder>` isn't enabled.
 -  Fixed a bug (#4052) - :doc:`Routing <general/routing>` with anonymous functions didn't work for routes that don't use regular expressions.
+-  Fixed a bug (#4126) - :doc:`Form Validation Library <libraries/form_validation>` was needlessly clearing ``$_config_rules`` when calling ``reset_validation()`` method.
 
 Version 3.0.1
 =============
--- a/user_guide_src/source/changelog.rst
+++ b/user_guide_src/source/changelog.rst
@@ -22,7 +22,6 @@ Bug fixes for 3.0.2
 
 -  Fixed a bug (#2284) - :doc:`Database <database/index>` method ``protect_identifiers()`` breaks when :doc:`Query Builder <database/query_builder>` isn't enabled.
 -  Fixed a bug (#4052) - :doc:`Routing <general/routing>` with anonymous functions didn't work for routes that don't use regular expressions.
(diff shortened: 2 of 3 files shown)


jim-parry:
Hmm - this presumes that one wants to preserve those settings.
Better would be to add a parameter ... reset_validation($keep=false) and to treat accordingly.

This should have a changelog entry, and the user guide would need to be updated, IMHO.

zoaked:
Hmmm...  Would there ever be a case where a config loaded from a file would want to be changed?  The way we use it is that all of the validations are stored in that file so no matter what we need/want to check it is already loaded.

I would be happy to add the items you requested.  Forgive my ignorance, but this is the first time doing this.  Am I correct in thinking that user_guide_src/source/libraries/form_validation.rst would be where the user guide changes should go?  Not sure where to locate the changelog.

jim-parry:
user_guide_src/source/libraries/form_validation.rst should be updated if the functionality changes, which you are proposing.
user_guide_src/source/changelog.rst should have a (potentially) single line entry indicating that some functionality has changed.

As to config changing, I have seen cases with several sets of validation rules ... one pre-loaded, and then reset & load a (possibly slightly) different second set. I just would not assume that the same config should automatically be reloaded.
Having a parameter $keep, which defaults to false, makes sure that the proposed changes do not break anyone's existing code, which is a good thing :)

zoaked:
In that case would you think it would be more beneficial for the masses to just skip clearing the current config instead of restoring it back to a copy of the original?  This would enable a reset to occur for any configuration instead of only for those loaded from a file.

jim-parry:
A parameter to clear/keep the config settings, with the default being consistent with the current behaviour, would make most sense to me.

zoaked:
How does that look - is this ok?

jim-parry:
Getting better :)

This isn't a bug ... your PR was originally described as an enhancement. In that case, the changelog entry would be a new feature, not a bug fix.

IIRC, a bug fix reference would be for an issue # previously reported, not for a problem with the PR itself.

I find your user guide additions confusing ...
1) You insist on passing TRUE, above line 717. This is opinion, not direction.
2) " This parameter must be true if your rules are being loaded from a config file" ... er, isn't your point to provide optional keeping of config rules? This suggests that the only reasonable value is TRUE if you have a config file. That is opinion, rather than direction.
3) The class reference for the method, lines 1079+ in the original .rst, need to be corrected as well.

Finally, your PR is not signed ... this is mentioned at the bottom of http://www.codeigniter.com/user_guide/contributing/index.html

system/libraries/Form_validation.php
```diff
@@ -1581,12 +1581,14 @@ public function encode_php_tags($str)
 	 * Prevents subsequent validation routines from being affected by the
 	 * results of any previous validation routine due to the CI singleton.
 	 *
+	 * @param	bool
 	 * @return	CI_Form_validation
 	 */
-	public function reset_validation()
+	public function reset_validation($keep_config=false)
```
narfbg: Spaces around the `=` operator are necessary, and `false` needs to be `FALSE` instead.

This is described in our styleguide: http://www.codeigniter.com/userguide3/general/styleguide.html

system/libraries/Form_validation.php
```diff
@@ -1581,12 +1581,14 @@ public function encode_php_tags($str)
...
 	{
 		$this->_field_data = array();
-		$this->_config_rules = array();
+		if( ! $keep_config)
```
narfbg: There must be a space after `if`, and also all `if` statements need curly braces (again, per our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html)).

Also, to make this more readable, this condition should be either before all other resets or after them.

user_guide_src/source/changelog.rst
```diff
@@ -22,6 +22,7 @@ Bug fixes for 3.0.2
 
 -  Fixed a bug (#2284) - :doc:`Database <database/index>` method ``protect_identifiers()`` breaks when :doc:`Query Builder <database/query_builder>` isn't enabled.
 -  Fixed a bug (#4052) - :doc:`Routing <general/routing>` with anonymous functions didn't work for routes that don't use regular expressions.
+-  Fixed a bug (#4126) - Added optional parameter to ``reset_validation()`` in ``Form_validation`` that stops ``_config_rules`` from being cleared.
```
narfbg: This is a new feature, not a bugfix ... and we reference libraries via `:doc:` syntax, as seen in all other changelog entries.

user_guide_src/source/libraries/form_validation.rst
```diff
@@ -714,6 +714,8 @@ file. You can organize these rules into "groups". These groups can
 either be loaded automatically when a matching controller/method is
 called, or you can manually call each set as needed.
 
+.. important:: Make sure that when using the ``reset_validation()`` method you pass ``true`` as a parameter.  If you do not then all the validation rules loaded from the config file will be cleared as well!
```
narfbg: Remove this.

user_guide_src/source/libraries/form_validation.rst
```diff
@@ -1076,8 +1078,9 @@ Class Reference
 		Permits you to set an array for validation, instead of using the default
 		``$_POST`` array.
 
-	.. php:method:: reset_validation()
+	.. php:method:: reset_validation($keep_config=false)
```
narfbg: Same as line 1587 in Form_validation.php

user_guide_src/source/libraries/form_validation.rst
```diff
@@ -1076,8 +1078,9 @@ Class Reference
...
 
+		:param	bool	$keep_config: Will leave configured validation rules alone when true or clear them when false.  This parameter must be true if your rules are being loaded from a config file
```
narfbg: 'Whether to reset validation rules from config files'

It's a short parameter description ... further explanations should be added in the method description below, if necessary.

zoaked:
@jim-parry - I read over the information you provided about signing the commits, but it appears that they refer to some command line application.  I have been using Git's build in file editor to make my changes.  Is it possible to sign off as requested from there or will I need to install and setup the local applicaiton?

@narfbg - Thank you for your comments.  I believe I have addressed all of the items you requested.

user_guide_src/source/changelog.rst
```diff
@@ -16,6 +16,7 @@ Release Date: Not Released
 -  General Changes
 
    -  Updated the *application/config/constants.php* file to check if constants aren't already defined before doing that.
+   -  Added optional parameter to ``reset_validation()`` in :doc:`form validation<libraries/form_validation>` that stops ``_config_rules`` from being cleared.
```
narfbg: This is just plain ugly ... you didn't make even a little effort of taking another changelog line as an example. Better remove it and leave it for me to add later.

system/libraries/Form_validation.php
```diff
@@ -1581,15 +1581,19 @@ public function encode_php_tags($str)
...
 	 */
-	public function reset_validation()
+	public function reset_validation($keep_config = FALSE)
 	{
 		$this->_field_data = array();
-		$this->_config_rules = array();
 		$this->_error_array = array();
 		$this->_error_messages = array();
 		$this->error_string = '';
+		if ( ! $keep_config)
```
narfbg: There needs to be an empty line before this one, and after the closing brace on line 1596

narfbg:
We still don't have an example of why this is really useful though ...

mwhitneysdsu:
For this to be useful, two conditions have to come into play:
1) You have to be using config files to store your validation rules in groups (using run($group) rather than run() to perform validation)
2) You have to be performing multiple validations (requiring use of reset_validation())

I think each of these is relatively uncommon, so the combination of the two becomes even less common. However, I have seen complaints about the reset_validation() method resetting the _config_rules property (which, of course, I could not track down with a quick check of 2 or 3 places). 

I've had enough reasons to use reset_validation() on my own, so, even though I don't use config files to store my validation rules, I do think it would be very frustrating if I did. After all, the property is only set in the constructor, and the library currently provides no way of restoring the rule groups once they've been cleared out by the reset_validation() method. Sure, you can use set_rules() to set validation rules on the individual fields, but you can't use the $group parameter of the run() method once the _config_rules property has been cleared out.

narfbg:
I see ... in that case I'd have to disagree with @jim-parry and say that this is indeed a bugfix and we don't need it to be optional.

It's just that whoever implemented the `reset_validation()` method (might've been me, don't remember) just made it reset all properties and never even considered what `$_config_rules` is there for.

zoaked:
I had attempted to describe the issue that this PR was addressing in my first post, but perhaps was not clear enough.  Hopefully now it is a bit clearer as to why I was trying to be so direct/forceful in the documentation about needing to have that parameter set when using config files.

@mwhitneysdsu - Thank you for providing the detailed explanation!

@narfbg - I can revert the proposed fix back to my original commit if that is what you think would be best.  Would there be any advantage in other situations of allowing people to choose to keep the configured validation rules during a reset_validation()?

@jim-parry - In your last comment there was mention of needing to sign off on the commits.  Is it possible to do so from Git's build in editor or do I need to download&configure some local command line utility?

Once a conclusion has been reached on what the most flexible and/or best fix is I will go back in and update the code accordingly. (& also address code format comments)

user_guide_src/source/changelog.rst
zoaked: @narfbg - My apologies for not getting this formatting correct - I have never seen this syntax before...  (& searching for "rst :doc:" and variations of that did not come up with any results.  I do not know what this style is called)

I will not make excuses, but do want to let you know that I did compare what was there to other nearby lines and tried to copy their style.  Specifically: "... :doc: \` Database \<database/index\> \`  ...."

It appeared to use :doc: as a prefix for pointing to a document, the hyperlink text to display, followed by <path/to/document>.  

Please forgive my ignorance.  If you could provide some specific, constructive, comments I would be happy to make the correct changes.  However if you feel that it would be easier to just do it yourself I can also understand that.

jim-parry:
The documentation tool we use is http://sphinx-doc.org/
It was originally made for python docs, but has been adapted to many purposes.
Documentation sources are created using http://docutils.sourceforge.net/rst.html, although I understand that support for markdown format is underway.
The style we use was customized from https://pypi.python.org/pypi/sphinx_rtd_theme

user_guide_src/source/changelog.rst
zhanghongyi: @zoaked you can always use `make html` in the user_guide_src folder to generate the html documents from the source. In this way, you can make sure that your changes are safe. And then you can check the generated files in the build folder to see if the result meets your expectation. 

Just add a tip here in case you did not know. :)

narfbg: @zoaked The RST syntax is fine ... I didn't say it was incorrect; I said that it is ugly and that you didn't take other changelog messages as examples. It's a styling/consistency issue.

Yet, you're still ignoring that last part. You don't need to google it (in fact, that couldn't possibly help you), you just need to search for "form validation" in the changelog.rst file and you'd see that it is always written as "Form Validation Library" (word 'Library' in there and capital letters) and always followed-up by a single space.
If you looked for another message that mentioned a property/variable, you'd also see that they are prefixed with the dollar sign.

I do insist on writing the message myself for this one.

narfbg:
@zoaked They don't have to choose, `$_config_rules` shouldn't ever be touched by `reset_validation()` in the first place ...

Just remove the `$this->_config_rules = array();` line from that method and that's it.

user_guide_src/source/changelog.rst
zoaked: @zhanghongyi - Thanks for the tip!

zoaked:
@narfbg - I just updated the PR to reflect your suggestion

user_guide_src/source/changelog.rst
```diff
@@ -22,6 +22,7 @@ Bug fixes for 3.0.2
 
 -  Fixed a bug (#2284) - :doc:`Database <database/index>` method ``protect_identifiers()`` breaks when :doc:`Query Builder <database/query_builder>` isn't enabled.
 -  Fixed a bug (#4052) - :doc:`Routing <general/routing>` with anonymous functions didn't work for routes that don't use regular expressions.
+-  Fixed a bug (#4126) - :doc:`Form Validation Library <libraries/form_validation>` was needlessly clearing ``$_config_rules`` when calling ``reset_validation()`` method.
```
narfbg: I insisted on writing this message myself and I pointed out styling issues at the time, but there's more to it than that ...

Now that we have a proper bugfix patch, the message needs to be altered a bit. The word 'needlessly' is too soft for describing something that was broken (as opposed to just being unnecessary) and `$_config_rules` doesn't mean much to most people. So it should be something like this instead:

    - Fixed a bug (#4126) - :doc:`Form Validation Library <libraries/form_validation>` method ``reset_validation()`` discarded validation rules from config files.

In addition, you're targeting the *develop* branch (which is 3.1.0-dev) with a fix for 3.0.x that should be in *3.0-stable* instead. These branches aren't, and couldn't possibly be in complete sync at all times and difference in changelogs are the most often cause for merge conflicts, which means I'd have to do extra work in order to merge this one into the 3.0-stable branch ...

So yeah, I know you want to do it all yourself (which is great and I do appreciate it), but that doesn't necessarily mean less work for others and can be quite the opposite.

zoaked:
Done.

narfbg:
Nope, thanks.

PR #4127, opened 2015-09-20
Added method to pull config fields out of post data

zoaked:
#2436 - Created a PR for this very useful feature.  In short it only pulls fields out of post data that also have a validation configuration.

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -418,11 +418,11 @@ class CI_Form_validation {
 	 public function get_valid_data()
 	 {
 		$valid_data = array();
-		foreach ($this->_field_data as $k => $v )
+		foreach ($this->_field_data as $k => $v)
 		{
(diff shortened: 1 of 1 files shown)


jim-parry:
Good idea. This should also have a changelog entry & update for the user guide, if it is to go ahead.

narfbg:
What if the validated data isn't in `$_POST`?

zoaked:
Is posting code in an issue considered to be providing it to the project?  I did not create this method and am wondering if the original author posting it to the above issue means it is being made available or not?  In my opinion it seems like it does, but I want to also make sure that all your project policies are being followed.

narfbg:
It is, but if you're just submitting a patch that was already provided (in whatever form) by another issue and we haven't applied it, it means we didn't like it in the first place.
One of the reasons why is the questions I asked above, which you didn't answer.

Closing here; you can discuss it in #2436 if you want to.

PR #4128, opened 2015-09-21
Automate documentation building in Travis-ci

zhanghongyi:
Developers need to manually generate documents for now. It would be cool if we enable travis-ci to take care of this. My changes will allow travis-ci to autobuild documentation sources. Through this, we can verify whether the commits for documentations are valid. Furthermore, if it is viable, we should automate deployment of the generated documents as well.

Currently, I only tested the generation of html pages. Shall I include more?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,19 @@ env:
 
 sudo: false
 
+install:
+  - apt-get install python-sphinx=1.2.3
+  - apt-get install python-sphinxcontrib.phpdomain  
+
 before_script:
   - sh -c "if [ '$TRAVIS_PHP_VERSION' = '5.2' ]; then pear channel-discover pear.bovigo.org && pear install bovigo/vfsStream-beta; else composer install --dev --no-progress; fi"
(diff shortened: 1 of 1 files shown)


jim-parry:
Can it handle epub books? PDF documents?

narfbg:
I'm not comfortable with this. Our tests suite is complex enough as it is and I'd like to keep it focused on actual code.

jim-parry:
The topic of integrating a documentation build into travis-ci somehow has come up several times in the past, specifically in conection with releases. How would you suggest that be better handled?

narfbg:
Why does it need to be "better handled"? Neither Travis builds, nor any kind of automated doc builds would show any more warnings than what I already see during the manual on-release build. And whatever I see then is easily fixable on the spot ...

IMO, this is a non-existing problem.

zhanghongyi:
As @narfbg pointed out, my pull request here might not be the best solution, so I raised an issue #4132 to open a thread for discussion. I hope we can reach to an agreement that whether we need it or not through the discussion.

narfbg:
Closing this. Even if some kind of similar automation happens, it won't be via Travis.

PR #4129, opened 2015-09-21
Enhance/lang fallback

jim-parry:
Added language fallback handling.

When loading a language file, CodeIgniter will load first the english version, if appropriate, and then the one appropriate to the language you specify. This lets you define only the language settings that you wish to over-ride in your idiom-specific files.

This has the added benefit of the language facility not breaking if a new language setting is added to the built-in ones (english), but not yet provided for in one of the translations.

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * CodeIgniter
  *
@@ -53,14 +54,21 @@ class CI_Lang {
 	 *
 	 * @var	array
 	 */
-	public $language =	array();
+	public $language = array();
 
 	/**
 	 * List of loaded language files
 	 *
 	 * @var	array
 	 */
-	public $is_loaded =	array();
+	public $is_loaded = array();
+
+	/**
+	 * Refactor: base language provided inside system/language
+	 * 
+	 * @var string
+	 */
+	public $base_language = 'english';
 
 	/**
 	 * Class constructor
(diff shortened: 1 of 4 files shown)


vkeranov:
Sorry for the offtopic, but I like to see the same coding style all over the framework. I am following the development since version 1.7 and I do not see the point of adding spaces between dots. There are no spaces in the other files and you're replacing $foo.$bar with $foo . $bar. The same is with ( ! $foo => (!$foo

narfbg:
The fallback to 'english' was only allowed because it is the only language provided by default with CI; it was otherwise a contentious decision and still arguable if having a fallback isn't just plain wrong ...

No such enhancements should be allowed.

jim-parry:
Hmmm - a fallback capability would prevent a new or changed key language setting from breaking an already working application, which sounds like a good idea to me.
What is the downside of such a fallback?

narfbg:
Well, you call it "break prevention" and I call it "error detection". :)
And it's also illogical to magically fallback to another language when you clearly wanted something else instead.

jim-parry:
Would this be more acceptable if there was a config setting to optionally enable it?

I don't see the illogic. If you are trying to say something in a non-native language, and you don't know the word(s) for it, then you would express it in your native language, until you learned the proper way (eg adding a new language setting to your translation).

It seems very logical to me that adding a new language setting should not break an application using the existing translations until they have been corrected.

narfbg:
As I said, the fallback to English was only ever accepted because CI comes with the 'english' language pack and therefore we can guarantee that every translation string in the framework itself has a meaningful fallback.
So there's that ... a somewhat acceptable compromise. But why fallback to another language?

And why "not break" until you've corrected your translations? That's a contradiction by itself ... it's broken either way.

vkeranov:
Why not remove (hide) all possible php errors and warnings then? Because the user is trying to achieve something with CI and he/she doesn't know php yet...

jim-parry:
Uncle - we can agree to disagree :)
I'll move this topic over to the translations repo.

ivantcholakov:
If this topic gets interesting in the future, here you are a visualized idea about the languages:
languages
It has been implemented through database settings, but it could be covered by the basic config settings, why not.

A finished application has "user interface language" and "data language" diversity, but this is beyond what a framework provides, I think.

About the fallback implementation vs. draconian checks - if there is no consensus, then a new configuration option may be introduced. For me language fallback would be convenience during the development. I would prefer to check and finish the language files near the end of a project.

Issue #2849, opened 2014-01-29
Default controller subdirectory (worked in 2.1.4 but not in the latest 3.0.0/release)


zentaos:
Hello,

I just updated to the latest 3.0.0/release build but the default controller doesn't working anymore if the controller is in a subdirectory.

$route['default_controller'] = 'page/news';
$route['404_override'] = 'error_404';
$route['translate_uri_dashes'] = FALSE;

I hope you can fix this.

Thank you.

seputrodimas:
so on the CI3 you "can't" change default_controller on sub-directory? 
is this feature really disabled? :(
dude... really? :(

narfbg:
Yes, really.

Please ask your questions on the [forum boards](http://forum.codeigniter.com/), not here.

PR #4141, opened 2015-09-25
added directory check

lucasfong:
seems directory check for default controller is missing


narfbg:
#2849

This is also explained in the upgrade instructions for 3.0.0.

PR #4144, opened 2015-09-28
Fix fatal error for PHP 5.3 on Windows.

dcarbone:
There is an issue with PHP 5.3 running on Windows where utilizing the MCRYPT_DEV_URANDOM algorithm results in a Fatal error with the message "Could not gather sufficient random data".

This proposed change tests for 5.3 on windows, and attempts to utilize a different algorithm.

The official documentation outlining the function http://php.net/manual/en/function.mcrypt-create-iv.php claims that MCRYPT_DEV_URANDOM should be sufficient, however with earlier versions of PHP 5.3 this is simply un-true.

This is especially bothersome to developers deploying to a RHEL 5 or 6 environment, but developing on Windows.  At the moment we are stuck on PHP 5.3.3, and this bug causes all kinds of issues.

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -575,7 +575,14 @@ class CI_Security {
 		}
 
 		// Unfortunately, none of the following PRNGs is guaranteed to exist ...
-		if (defined('MCRYPT_DEV_URANDOM') && ($output = mcrypt_create_iv($length, MCRYPT_DEV_URANDOM)) !== FALSE)
+		if (is_php('5.3') && stripos(php_uname('a'), 'win') === 0)
+		{
+			if (defined('MCRYPT_RAND') && ($output = mcrypt_create_iv($length, MCRYPT_RAND)) !== FALSE)
+			{
+				return $output;
+			}
+		}
+		else if (defined('MCRYPT_DEV_URANDOM') && ($output = mcrypt_create_iv($length, MCRYPT_DEV_URANDOM)) !== FALSE)
 		{
 			return $output;
 		}


ivantcholakov:
It seems you experience this bug: https://bugs.php.net/bug.php?id=55169
It has been fixed as of PHP 5.3.7

dcarbone:
@ivantcholakov: This is true, however official RHEL 5 & 6 distros have kept php at version 5.3.3.  I cannot reliably develop locally with a later version of PHP as there are features added in 5.3.4 and 5.3.6 in particular that, if used, would simply not work once deployed.

One possible alteration that could be made would be to test for any PHP version below 5.3.7 on Windows as opposed to any PHP 5.3.x

narfbg:
Your patch may avoid a fatal error, but results in predictable output (it may say 'RAND', but it's not at all random), which is insecure - for that alone I cannot accept it.

I could talk about this a lot, but long story short: CI is no longer committed to making itself work on ancient PHP versions (as stated in the README.md file - no guarantees are given) and there's no easy way around this one either, so there's that - we just won't fix this one, sorry.

The easiest way for you to work around this would be to uninstall mcrypt (or never install it in the first place), but you really shouldn't test on Windows in the first place if you're supposed to run the app on Linux. Plus, there's EPEL for RedHat 6 ... you're not really stuck with 5.3.3.

dcarbone:
Believe me I am aware of all of this, and I understand your decision.  My position is an increasing fringe case (I am unable to utilize anything other than Windows locally, and am not allowed utilize any unofficial RHEL package channels), but it is the current state of affairs for many developers working within academia.

At any rate, thank you for looking at it at least.

PR #4148, opened 2015-09-30
Documentation: Generate pulldown menu using sphinx toctree

zhanghongyi:
The current content of pulldown menu is hard-coded which means it would be hard to maintain in the future. I take a look at Sphinx and found that it has a built-in function toctree() which provides the table of contents as a HTML list. So I think we can utilise this function to generate our pulldown menu. 

I tried my best to make it look similar as before, here is what I got(still a bit different though):
screenshot

The changes are tested on Chrome, Firefox and Safari, but not IE...

**  Some indentations look different from my text editor. If this idea is acceptable, I will keep improving them later.


zhanghongyi:
@jim-parry I modified your codes of pulldown menu a bit. Could you take a look and give me some comments? :)

user_guide_src/source/_themes/sphinx_rtd_theme/breadcrumbs.html
```diff
@@ -14,6 +14,9 @@
           <a href="{{ pathto('_sources/' + sourcename, true)|e }}" rel="nofollow"> View page source</a>
         {% endif %}
       </li>
+    <div style="float:right;margin-left:5px;" id="closeMe">
```
narfbg: You're obviously using 2-space indentation ... we use tabs.

jim-parry:
This looks a lot cleaner than the hack that Rufnex and I came up with!!
It works well, as far as I can tell :)

user_guide_src/source/_themes/sphinx_rtd_theme/breadcrumbs.html
zhanghongyi: @narfbg All HTML codes in the current theme that we use seem to use 2-space indentation. To be consistent with the theme, is it okay to keep 2-space indentation here(only in HTML)?

narfbg: It's never OK with me, but I don't care much about random HTML in the docs as long as it's consistent.

But that was my point in disguise - that it's not consistent. Look at line 16 above ... it's a closing tag and you're opening a new one right after it at a different indentation level.

zhanghongyi:
Just updated the codes and fixed the indentations. Hopefully it is better now :)

narfbg:
The "Contributing to CodeIgniter" section needs to be in the first column, not the last one.

jim-parry:
@andrey I disagree with you. Hongyi's approach has generalized the pulldown menu, based on the sphinx doctree for our user guide. The pulldown menu layout was originally hand-coded to match the pulldown menu style in the CodeIgniter 2 user guide, which was hand/hard-coded too IIRC.

If the pulldown menu style isn't "suitable", I think the solution is to modify the TOC order in the user guide, not to program exceptions.

narfbg:
I'm not saying his approach is wrong.

I'm saying the hacky hard-coded pulldown menu displayed that section where we *want* it and this one does not. Whether that's due to the TOC itself being in the wrong order is a different question.

Btw, that `@andrey` guy is not me. :)

jim-parry:
Oops re the tag.
What do you think of modifying the TOC order - would that not address the "consistency problem" without having to program an exception?

narfbg:
Works for me, yes. I'm not in favor of making logical exceptions either.

jim-parry:
Should the TOC order be changed as part of this PR or a separate one?

narfbg:
Can be done separately after this is merged, yes. I just don't want us to forget about that ... will you do it?

jim-parry:
Will do :)

zhanghongyi:
@jim-parry thanks for taking over the doctree rearranging. 
@narfbg thanks for cleaning the css.
And thank you all for your time to review it. :)

PR #4151, opened 2015-10-01
Add missing NULL to defaults for Input methods

gdhnz:
Added $index = NULL to the following methods:
* post_get
* get_post
* cookies
* server

It avoids the php warning when getting the various input arrays.  Matches the behaviour of the get and post methods.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -259,7 +259,7 @@ class CI_Input {
 	 * @param	bool	$xss_clean	Whether to apply XSS filtering
 	 * @return	mixed
 	 */
-	public function post_get($index, $xss_clean = NULL)
+	public function post_get($index = NULL, $xss_clean = NULL)
 	{
 		return isset($_POST[$index])
 			? $this->post($index, $xss_clean)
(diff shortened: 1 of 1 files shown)


narfbg:
This is intentional, and it doesn't make sense for `get_post()`, `post_get()` at all.

gdhnz:
It may not make sense for get_post() or post_get() but it does make sense for cookie() and server().  This is a regression from v2.2.4 behaviour (where it uses '' instead of null).

narfbg:
Well, first of all - you talk about `cookie()`, but your patch doesn't even touch it. And it doesn't touch it because it already works that way, so I don't see why you're talking about `cookie()` at all.

Secondly, the behavior in 2.x with the default value (of '' there and NULL in 3.x) is completely different - 2.x returned FALSE in that case, 3.x returns the whole array - that's not a regression; it's an ehnancement.

And finally, because of how it worked in 2.x, you can't possibly say that calling `server()` without passing a parameter to it has ever been useful, so ... how does that make sense?

gdhnz:
My mistake adding cookie in the description.

I don't understand why get(), post(), and cookie() have the ability and was aiming for some consistency across all those types of methods.

PR #4167, opened 2015-10-11
disable pulldown menu on mobile devices

zhanghongyi:
I would like to suggest that we disable pulldown menu on small screens since this feature is designed for desktop devices and it does not function well on mobile devices.

For example, when enabling pulldown menu on mobile devices, the bar icon will not disappear and a blank side bar will show up after clicking the bar icon.
screen shot 2015-10-11 at 16 02 33
And while the pulldown menu is helpful on large screens, it is not really easy to view on mobile devices.

As the recent release started to offer Epub version of docs so that users could have one more alternative on mobiles, I think it would be safe to remove this functionality on small screen.

### DIFF
--- a/user_guide_src/source/_themes/sphinx_rtd_theme/static/js/theme.js
+++ b/user_guide_src/source/_themes/sphinx_rtd_theme/static/js/theme.js
@@ -25,7 +25,7 @@ $(document).ready(function () {
     $('#closeMe').toggle(
         function ()
         {
-            setCookie('ciNav', true, 365);
+            setCookie('ciNav', 'true', 365);
             $('#nav2').show();
             $('#topMenu').remove();
             $('body').css({background: 'none'});
@@ -35,7 +35,7 @@ $(document).ready(function () {
         },
         function ()
         {
-            setCookie('ciNav', false, 365);
+            setCookie('ciNav', 'false', 365);
             $('#topMenu').remove();
             $('#nav').hide();
             $('#nav2').hide();
@@ -50,14 +50,19 @@ $(document).ready(function () {
         //$('#nav').slideToggle();
     }
     // END MODIFICATION ---
+
 });
 
 // Rufnex Cookie functions
 function setCookie(cname, cvalue, exdays) {
+    // expire the old cookie if existed to avoid multiple cookies with the same name
+    if  (getCookie(cname) != false) {
+        document.cookie = cname + "=;expires=Thu, 01 Jan 1970 00:00:00 GMT";
+    }
     var d = new Date();
     d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
     var expires = "expires=" + d.toGMTString();
-    document.cookie = cname + "=" + cvalue + "; " + expires;
+    document.cookie = cname + "=" + cvalue + "; " + expires + "; path=/";
 }
 function getCookie(cname) {
     var name = cname + "=";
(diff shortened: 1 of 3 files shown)


narfbg:
Where does the 768px value come from? And if it's just about hiding the pulldown menu on mobiles, why are you changing alt, title texts?

jim-parry:
Experimented ... 768px is the width at which the sphinx-generated side menu collapses to a hamburger icon. The pulldown menu looks silly and unusable from that point smaller, and I agree it should be hidden.
Depending on the settings at the time the sidemenu collapses, the left margin might also become unusable.
Screenshots attached.
screenshot-2
screenshot-3

zhanghongyi:
768px is defined as the tablet size in the theme and @jim-parry have confirmed that it is the size when layout changed. By removing the icon, I guess there is no other way to trigger pulldown menu, so we do not have to worry the left margin anymore, right?

I changed the title&alt just to make them more accurate so that it will show "switch layout" when mouse is over. I feel the change is too small so I did not make a separate pull request. Forget to mention it in the pull request. :(

jim-parry:
I think we might have to worry about the left margin... if the pulldown menu is currently active, and the browser window is resized, should we not re-enable the side menu and disable the pulldown? If we don't, then it would still be hidden, and clicking the hamburger would result in a large white strip down the left, would it not? I think that is what happened in my earlier screenshots.
I suspect there could be a bit more to it than just removing the pulldown icon.

zhanghongyi:
hmm makes sense. I did not think of the resizing scenario.

narfbg:
Sorry, but I want the 'classic' -> 'switch' change reverted. Doesn't matter if it's a small, medium or big change - if you want to change it, it has to happen in another PR.

zhanghongyi:
I finally got some time to update this pull request. :)

@narfbg I have reverted alt&title changes. 

@jim-parry for the resizing issue, I came up with a workaround to solve it. It seems that after changing the css with jQuery, the css media query is not effective anymore. So my workaround is to trigger the margin change when window is resized...It works, but it is not an elegant solution though.

user_guide_src/source/_themes/sphinx_rtd_theme/static/js/theme.js
```diff
@@ -25,7 +25,7 @@ $(document).ready(function () {
     $('#closeMe').toggle(
         function ()
         {
-            setCookie('ciNav', true, 365);
+            setCookie('ciNav', 'true', 365);
```
narfbg: Why?

zhanghongyi: The behaviour does not change because javascript string concatenation will automatically cast true to "true". I changed it just for the consistency with the getCookie() function so that what we set would be the same as what we get. As cookie values should be strings, I think it would be better to pass a string to the setCookie function rather than rely on string concatenation to do the trick.

But the previous codes work as well. If you think this change is unnecessary, I can revert it.

narfbg: No, it's fine like this ... it's just weird when you change booleans to strings, but in this case it seems that it is the current code that relies on automated type conversions.

However, if you're changing it to string here, why still using boolean on line 59 (which *you* are inserting)?

zhanghongyi: If I did not get you wrong, line 59 is to get cookies? If the getCookie function returns a boolean type false, it means there is no cookies set for this name. Otherwise, we need to delete the old cookie first.

narfbg: I know what it means, but it's the only place you're checking for a boolean false instead of the string 'false' ... which might even mean it doesn't work (if it does, JavaScript is an even worse language than I thought).

narfbg: On a second thought, it may work without JS being that horrible, but the code sucks. If the same function can return 'false', 'true' and boolean false, and if those are its *only* possible return values ... it just sucks; it's confusing and illogical. At least change the strings to 'yes' and 'no' or something.

zhanghongyi: I see what you mean. That is also part of the reason that I changed setCookies false to "false" because actually `getCookies() == false` means that the cookie does not exist. I think the simple way out is to change the return of getCookies() to empty string `""` when the cookies are not set. Then all the returns would be of the same type.

zhanghongyi: @narfbg what do you think?

narfbg: Well, I don't know why you insist on using 'true', 'false' (this is something that I would avoid), but I guess an empty string in that case works, yes.

zhanghongyi: fixed in the new commit. :)

narfbg:
Looks fine to me now.

@jim-parry I'll leave this for you to merge, unless you've got something to comment on.

jim-parry:
Looks good to me too :)

PR #4168, opened 2015-10-12
Don't trigger double queries.

apenchev:
No longer runs query twice when working with a column that contains "select". For instance:
INSERT INTO params (param) VALUES ('param_selector');


narfbg:
We no longer accept non-security fixes for CI2, sorry.

PR #4172, opened 2015-10-15
Update postgre_driver.php

russ-ab:
Updating function "_execute()" - Currently this function returns a "resource" when the query is performed successfully. When failing it throw a PHP exception (which doesn't happen when running on MySQL). Instead of using "pg_query()" now it uses the function "pg_send_query()", thus saving the result in a variable that will be accessible through the error capture function. If the error code is 0, it means there were no errors and it'd the outcome, otherwise it'd return 'false' (following the same behavior when running on MySQL).

Updating function "error()" - From now saving the result resource it's possible to retrieve what were the query's error Code and Message. Before the error Code was not being taken and the flow diverted due to the exception even the Message was not available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,4 +48,5 @@ matrix:
 branches:
   only:
     - develop
+    - 3.0-stable
     - /^feature\/.+$/
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -344,6 +344,9 @@ $config['encryption_key'] = '';
 |
 |	Whether to match the user's IP address when reading the session data.
 |
+|	WARNING: If you're using the database driver, don't forget to update
+|	         your session table's PRIMARY KEY when changing this setting.
+|
 | 'sess_time_to_update'
 |
 |	How many seconds between CI regenerating the session ID.
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -238,7 +238,15 @@ class CI_Config {
 
 		if (isset($protocol))
 		{
-			$base_url = $protocol.substr($base_url, strpos($base_url, '://'));
+			// For protocol-relative links
+			if ($protocol === '')
+			{
+				$base_url = substr($base_url, strpos($base_url, '//'));
+			}
+			else
+			{
+				$base_url = $protocol.substr($base_url, strpos($base_url, '://'));
+			}
 		}
 
 		if (empty($uri))
@@ -293,7 +301,15 @@ class CI_Config {
 
 		if (isset($protocol))
 		{
-			$base_url = $protocol.substr($base_url, strpos($base_url, '://'));
+			// For protocol-relative links
+			if ($protocol === '')
+			{
+				$base_url = substr($base_url, strpos($base_url, '//'));
+			}
+			else
+			{
+				$base_url = $protocol.substr($base_url, strpos($base_url, '://'));
+			}
 		}
 
 		return $base_url.ltrim($this->_uri_string($uri), '/');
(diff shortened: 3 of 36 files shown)


narfbg:
`pg_query()` throwing an exception? First time I hear that, and the PHP manual doesn't mention anything like that either.

russ-ab:
You can test yourself.

Connect to a valid **Postgres DB** and try the following code:

```php
$sql = 'SSELECT datname FROM pg_database';
$query = $this->db->query($sql);

if ($query) {
    echo 'Ok';
} else {
    echo 'Error';
}
```

That will throw an exception which will be caught by CI_Exceptions->log_exception() with:
**Message:** *pg_query(): Query failed: ERROR:  syntax error at or near &quot;SSELECT&quot;
LINE 1: SSELECT datname FROM pg_database ^'*
**Severity:** *2*

narfbg:
I wouldn't ask if I was working with PostgreSQL on a daily basis in order to have the environment ...

Anyway, I just installed one just to test this, and here's what I got:

<pre>
$ php -a
Interactive mode enabled

php > $res = pg_connect('host=localhost dbname=testdb user=testuser password=testpass');
php > var_dump(pg_query($res, 'SSELECT foo FROM bar'));
PHP Warning:  pg_query(): Query failed: ERROR:  syntax error at or near "SSELECT"
LINE 1: SSELECT foo FROM bar
        ^ in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. pg_query() php shell code:1
bool(false)
php >
</pre>

No exception, just an E_WARNING message and boolean FALSE returned, as is expected by the CI code. Tested within CI as well and it's still just a warning that's being triggered.

The error is indeed logged by `CI_Exceptions::log_exception()`, but that's just the method's name ... it doesn't log only exceptions.

ivantcholakov:
I tried pg_query() too, a PHP error is triggered, not an exception.

psql (PostgreSQL) 9.4.5
PHP 5.6.5

```php
<?php defined('BASEPATH') OR exit('No direct script access allowed');

class Test_postgre extends CI_Controller {

    public function index() {

        $sql = 'SSELECT foo FROM bar';

        $pg_host = 'localhost';
        $pg_port = '5432';
        $pg_user = 'postgres';
        $pg_pass = '*****';
        $pg_db   = 'pg_db';

        $conn = pg_connect("host=$pg_host port=$pg_port dbname=$pg_db user=$pg_user password=$pg_pass");
        var_dump(pg_query($conn, $sql));
    }
}
```

```
A PHP Error was encountered
Severity: Warning
Message: pg_query(): Query failed: ERROR: syntax error at or near "SSELECT" LINE 1: SSELECT foo FROM bar ^
Filename: controllers/Test_postgre.php
Line Number: 16
```

russ-ab:
Indeed is an ERROR and not an EXCEPTION, this is the expected behavior? Just ignore the warnings?

And about the error code?

narfbg:
What about the error code?

To be honest, "pg_query() throwing an exception" is the only thing I understood from your description and that's why I only replied to that. I don't really understand what's your goal here, or what's the problem that you're trying to solve.

russ-ab:
How do i retrieve the error code from a sql query execution?

narfbg:
Unfortunately, `pg_last_error()` doesn't provide it and that's why it is not being returned by the `error()` method.

So, you're suggesting a bit of refactoring that would allow better error feedback ... OK, we can work towards that.

Issue #4059, opened 2015-08-16
Allow CamelCase on Model names


jsnoriegam:
As the title implies... I know the documentation specify that only the first leeter in the model class name should be uppercase, but i don't really see the purpose on that and i don't think model class names need to be that strict.

I've removed the 293th line in core/loader.php lib on my projects and everything looks fine for the moment:

```php
$model = ucfirst(strtolower($model));
```

But maybe that break something somewhere else...

So if this is not against CI philosophy, please consider it.

Thanks.

narfbg:
The question isn't whether they *need* to be that strict, but to have a single, consistent naming convention, i.e. all classes are named in Ucfirst.

`strtolower()` isn't applied to other classes, so I've removed that one.

ProfNandaa:
Sounds good. Code also looks more beautiful with `Your_Model` as opposed to `Your_model`, IMHO. Thanks!

vimalmistry:
I am getting error on linux hosting with model name like LoginModel.php
While in my local xampp its working fine.
note. I am using HMVC extention.

jim-parry:
@vimalmistry That is a feature and not a bug. Your point should have been raised on the forum. The user guide says "Class names must have the first letter capitalized with the rest of the name lowercase."

vimalmistry:
@jim-parry Thanks for reply. I made changes in my all code. Now I am using Login_model.php. :)

PR #4176, opened 2015-10-16
Lowercase is not required for model names anymore

dirtycoder:
From version 3.0.2, only ucfirst is used at models name.

As seen here:
https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Loader.php#L293

### DIFF
--- a/user_guide_src/source/general/models.rst
+++ b/user_guide_src/source/general/models.rst
@@ -81,8 +81,8 @@ The basic prototype for a model class is this::
 	}
 
 Where **Model_name** is the name of your class. Class names **must** have
-the first letter capitalized with the rest of the name lowercase. Make
-sure your class extends the base Model class.
+the first letter capitalized. Make sure your class extends the base Model
+class.
 
 The file name must match the class name. For example, if this is your class::
 


narfbg:
The `strtolower()` call was removed because it was unnecessary, not because the requirements are changed.

dirtycoder:
Wasn't the change made to allow capitals in the middle of a models name?

Maybe I misunderstood this commit:
https://github.com/bcit-ci/CodeIgniter/commit/825fab7370a28b6c05da126842dd8df25e51e026

narfbg:
https://github.com/bcit-ci/CodeIgniter/issues/4059#issuecomment-131699436

My only real concern is that 'lowercase' is converted to 'Lowercase', because lowercase was the requirement prior to CI3, and because a lot of our users write everything in lowercase no matter what you tell them.
I don't like unnecessary function calls, so that's why I removed it.

But other than that, our requirement is still Ucfirst and not One_Two. If you use the latter, you're just ignoring our requirements and I hope some other hidden condition breaks your app. :D

dirtycoder:
Oh, now I see, thanks.
Nah, I kept the consistency, just thought that the documentation was forgot to be updated.
And I hope that any hidden condition breaks our apps, haha. :stuck_out_tongue_winking_eye:

PR #4183, opened 2015-10-20
Router can now use overridable search paths

NeoLegends:
Previously, the router class was limited to loading controllers from APPPATH. For greater flexibility, it can now search multiple folders for controllers, kind of like the Loader.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -399,13 +399,13 @@ if ( ! is_php('5.4'))
 	$class = ucfirst($RTR->class);
 	$method = $RTR->method;
 
-	if (empty($class) OR ! file_exists(APPPATH.'controllers/'.$RTR->directory.$class.'.php'))
+	if (empty($class) OR ! file_exists($RTR->folder.'controllers/'.$RTR->directory.$class.'.php'))
 	{
 		$e404 = TRUE;
 	}
 	else
 	{
-		require_once(APPPATH.'controllers/'.$RTR->directory.$class.'.php');
+		require_once($RTR->folder.'controllers/'.$RTR->directory.$class.'.php');
 
 		if ( ! class_exists($class, FALSE) OR $method[0] === '_' OR method_exists('CI_Controller', $method))
 		{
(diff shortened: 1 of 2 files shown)


narfbg:
Sorry, but no. Everything has its limits.

NeoLegends:
Why not? I'm sure we can integrate it in a satisfactionary way.

narfbg:
No, we can't; I can never be satisfied with it. :)

"Why not?" is the wrong question to ask ... I should be asking you "why so?" instead. And I'm not asking simply because I know you can't possibly give me an answer that I'd like.

Sorry to put it that way, but I've been asked "why not?" a thousand times and it gets tiresome to be nice and explain everything to everybody, one at a time. Some additions just don't make sense.

casapex:
hola mundo

2015-10-20 13:10 GMT-05:00 Andrey Andreev <notifications@github.com>:

NeoLegends:
@narfbg Scenario is as follows: We're using CodeIgniter in a small to medium-sized company to power multiple large websites. CI separates most of the code very nicely, but we found there's still room for improvement. Through using a layered system we're able to really separate application logic from our custom DB logic and from the upstream CI core.

This separation goes so far as that it allows laying independently versioned website code versions on top of independently versioned BaseController / BaseModel versions on top of an independently versioned CI core. The core will be kept clean of modifications to be able to apply official updates regularly.

Most of this works "out of the box", since the CI_Loader _has_ overridable search paths. The only exception to this is the CI_Router / CodeIgniter.php for controller loading. Since the latter one is no class, we cannot inject the versioned search paths.

This is where my changes come into play. They just add that little bit of flexibility people that use a custom controller path need. For people who do not use them, this should work as expected, since APPPATH is the default search path.

mwhitneysdsu:
While I don't think it's likely to happen, I do think it would be nice to have a single method of defining a set of base directories which would be searched by the (4*) classes/functions/files in CI which act as class/file locators/loaders. I don't think increasing the router's involvement in the process is a good idea, though I think I understand why you did that.

Additionally, though it's a bit of a tangent, when you start messing with the search paths in the loader you can run into some issues. The biggest issue I ran into was with `_ci_load_stock_library()`'s [handling of](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Loader.php#L1097) `_ci_library_paths`. If you insert a path between the 2 default paths, like `array(APPPATH, BFPATH, BASEPATH)`, it ends up searching `APPPATH` twice and not checking the custom path at all (I guess that's one reason the class has a comment saying `Don't mess with them.` at the top of the list of property definitions). I ended up taking a [slightly slower approach](https://github.com/ci-bonfire/Bonfire/blob/develop/bonfire/core/BF_Loader.php#L108) of searching for the `BASEPATH`/`APPPATH` and unsetting them if they were found, though I don't have hard data on how much slower it actually is, and haven't spent much time looking for a faster approach.

* If I haven't missed any, the 4 are:
1. [/system/core/CodeIgniter.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/CodeIgniter.php#L402)
2. [/system/core/Common.php load_class() function](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Common.php#L141)
3. [/system/core/Router.php (multiple methods)](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Router.php#L302)
4. [/system/core/Loader.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Loader.php)

Config and Lang include `load()` methods, each with its own level of knowledge of the directory structure, but they are at least limited to loading files specific to the class, and the loader passes requests to load those file types to the respective classes.

I understand, more or less, why extremely similar functionality is spread across these 4 locations (due primarily to the order in which things are loaded/instantiated in CodeIgniter.php), but it is a little frustrating when you need to do something as deceptively simple as define a third base directory/search path for the loader.

PR #4185, opened 2015-10-21
Update evil_attributes

erknrio:
evil_attributes remove javascript event handler but also remove data-*on* attributes for example: data-section: 5. This update remove the same attributes of w3schools: http://www.w3schools.com/jsref/dom_obj_event.asp


narfbg:
Neither 'data-section' nor 'data-onanything' is being removed. I don't know where you're getting this from ...

PR #4188, opened 2015-10-23
Update database.php

DevotionGeo:
Change $db['default']['dbdriver'] from 'mysql' to 'mysqli'. (/application/config/database.php)
If adding a single 'i' makes a difference (better database driver + no notice/error messages for deprecated driver), why not add it.


jim-parry:
Makes sense to me :)

narfbg:
It may make sense, but it's not a security-related patch, which is the only thing we accept for CI2.

jim-parry:
My bad, sorry! I didn't realize he was talking about CI2. 'mysqli' is indeed the default in CI3 :-0

PR #4190, opened 2015-10-23
Change _config['save_path'] variable to match as described in the use…

bigmekk:
…r_guide sess_save_path


narfbg:
Configuration variables don't necessarily match internal code variables. This would break the session class.

Issue #1534, opened 2012-06-26
SQL EXPLAIN badly needed (profiler or output)


iazi:
There seems to be no option with codeigniter's active record to enable EXPLAIN before SQL statements. It would be nice having it in the profiler for example.
Joan

Brian-Webb:
+1

DavidOtis:
It seems to me that this should be handled in the individual db drivers because each different database has different explain syntax. Instead of printing them in the in the profiler, I think an explain function should just return the results which would allow users to print/log/email or do whatever else they want to with them. Thoughts?

jim-parry:
Should there be an explain() method in the DB_Driver, with optional parameters?
No expectation about what it returns, nor about the parameter interpretation ... just an expectation that each database driver will implement it in an appropriate fashion.

DavidOtis:
Yes, I think there should be an explain() method in the DB_Driver, but I'm not sure what optional parameters it would need. Shouldn't it just take in the query to be explained?

narfbg:
"No, I don't think we should have [put functionality name here]" - said no user ever. :)

narfbg:
If someone wants to have a go at this - submit a pull request. No use having this issue just sitting open though.

PR #4194, opened 2015-10-26
Adds SQL Explain option to query()

DavidOtis:
This PR addresses #1534. I've implemented an explain function for all of the drivers with simple explain syntax. Others will require a little more thought. I haven't touched the PDO subdrivers or any databases that require settings changes to show query plans like cubrid or sqlsrv.


narfbg:
If this feature is ever implemented (and I don't really see that happening, to be honest), it doesn't belong in `query()`. Sorry.

webXperts:
ok

Issue #4189, opened 2015-10-23
Refactor for exceptions


jim-parry:
Enhancement Project: replace usage of show_error() with exceptions.
A class or function, instead of calling show_error to display an error message and then exit, should instead throw a PHP exception.

Benefits: improved error handling, better testability. 

Related: enhance or replace error-handling writeup in the user guide.

Procedure: 
- In general, replace every show_error() call with a throw new Exception(), or as appropriate, one of the other predefined exceptions.
- Translations would be mostly dropped in that process as they don't make sense for exceptions ... cases where a translation may be necessary will be discussed/decided on the fly.
- And just to make the transition smoother, show_error() itself should be altered to throw Exception.

Process: 
- Tackle the affected classes one at a time, making sure that the unit testing and user guide are updated to match.
- Changes must adhere to the CI style guide.
- Each converted class, together with its unit testing and documentation changes, should be handled as a separate PR, referencing this issue.

I will create a feature branch for this, so that it doesn't block the main branch.

Affected classes/helpers:
- system/core/Common (contains the show_error function cited above)
- system/core/Config
- system/core/Lang
- system/core/Loader
- system/core/Router
- system/core/Security
- system/core/URI
- system/database/DB
- system/database/DB_Forge
- system/database/DB_query_builder
- system/database/DB_utility
- system/database/drivers/pdo/pdo_driver
- system/helpers/path_helper
- system/libraries/Driver
- system/libraries/Encryption
- system/libraries/Ftp (also uses lang->line)
- system/libraries/Migration (also uses lang->line)
- system/libraries/Pagination
- system/libraries/Xmlrpc
- system/libraries/Xmlrpcs

narfbg:
In relation to #4200 ... let's leave `show_error()` as is for now.

Changing it to just throw an exception may not be as good as I initially thought. It may still happen, but after everything else is done.

narfbg:
Final: `show_error()` remains unchanged and a few such calls may remain ... The one triggered by CSRF stays for sure. See #4229.

narfbg:
@DavidOtis Look above and you'll see 4 pages of mentions that nobody cares about ... Please reference this issue *only* from your PR descriptions (one reason to actually write a description instead of leaving them empty) and not in every commit message.

PR #4200, opened 2015-10-29
refactor error handling in /core/Common

zhanghongyi:
Change show_error function in /core/Common to just throw new exception according to issue #4189.

- make show_error() simply throw new exceptions. If an error template rendering is needed, users can call the rendering function when an exception is caught.
- remove test mocking function for show_error()
- the error handling part of user guide is modified accordingly
- a trivial test case is added which tests nothing more than PHP just to validate my changes.(It could be removed after the change is finalized.)

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -393,11 +393,13 @@ if ( ! function_exists('show_error'))
 	/**
 	 * Error Handler
 	 *
-	 * This function lets us invoke the exception class and
-	 * display errors using the standard error template located
-	 * in application/views/errors/error_general.php
-	 * This function will send the error page directly to the
-	 * browser and exit.
+	 * This function will be deprecated in version 3.1.0.
+	 * It is encouraged to use PHP exceptions in your application.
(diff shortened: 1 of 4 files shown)


system/core/Common.php
```diff
@@ -393,11 +393,13 @@ function is_cli()
 	/**
 	 * Error Handler
 	 *
-	 * This function lets us invoke the exception class and
-	 * display errors using the standard error template located
-	 * in application/views/errors/error_general.php
-	 * This function will send the error page directly to the
-	 * browser and exit.
+	 * This function will be deprecated in version 3.1.0.
```
narfbg: You can just remove this description altogether.

narfbg: Or ... on another note, don't do anything with it for now.

narfbg:
Please read this: https://github.com/bcit-ci/CodeIgniter/issues/4189#issuecomment-153000903

Don't close this PR though ... it may be merged later.

zhanghongyi:
okay I see. I will wait for other changes to be done first :)

PR #4205, opened 2015-11-01
Added function show_flashdata() to system/libraries/Session/Session.php

DevotionGeo:
Normally we call a flashdata in the view file as follows,

<?php if($this->session->flashdata('login_success')): ?>
  <p class="alert success"><?= $this->session->flashdata('login_success') ?></p>
<?php endif; ?>

But by using this function the view file will be more cleaner and its easier to call as follows,
<?= $this->session->show_flashdata('login_success', 'alert success'); ?>

for only a single if-else block of flashdata its OK, but when we have 5 or more if-else blocks in our view for the flashdata, then it looks very cluttered, this function is to clean it up.

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -888,9 +888,11 @@ class CI_Session {
 	}
 	
 	/**
+	 * Show flashdata
+	 * 
 	 * Calling flasdata, and avoiding if-else check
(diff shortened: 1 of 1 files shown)


narfbg:
No ... this is wrong in at least 3 different ways.

PR #4206, opened 2015-11-01
Update user_agents.php

MartySVK:
added Chrome OS


application/config/user_agents.php
```diff
@@ -29,6 +29,7 @@
 	'win95'				=> 'Windows 95',
 	'windows phone'			=> 'Windows Phone',
 	'windows'			=> 'Unknown Windows OS',
+        'CrOS'				=> 'Chrome OS',
```
narfbg: You're using spaces for indentation here ... we use tabs.

PR #4207, opened 2015-11-02
Update readme.

jim-parry:
Make CodeIgniter website references use https, and update the IRC link to the freenode page.


readme.rst
```diff
@@ -53,11 +53,11 @@ agreement <https://github.com/bcit-ci/CodeIgniter/blob/develop/user_guide_src/so
 Resources
 *********
 
--  `User Guide <http://www.codeigniter.com/docs>`_
+-  `User Guide <https://www.codeigniter.com/docs>`_
 -  `Language File Translations <https://github.com/bcit-ci/codeigniter3-translations>`_
 -  `Community Forums <http://forum.codeigniter.com/>`_
 -  `Community Wiki <https://github.com/bcit-ci/CodeIgniter/wiki>`_
--  `Community IRC <http://www.codeigniter.com/irc>`_
+-  `Community IRC <http://webchat.freenode.net/?channels=%23codeigniter>`_
```
narfbg: Might as well have https here too :)

jim-parry:
I checked, and yes, freenode can take advantage of https.
I did not put https for the forum ... it breaks the retrieval of assets.
I think there might be one more place I have to change inside the forum before that will be a go.

zhanghongyi:
Should the Github description be updated as well?
screen shot 2015-11-06 at 23 49 28

Also, I noticed every file referenced a link of the user guide, which still uses http. I wonder if they should be updated. I am willing to do the work if it is a need.

jim-parry:
@zhanghongyi You are correct! Global serach & replace should fix that - I have updated my PR.

narfbg:
The webchat link was the only thing worth modifying and now I have to use Ctrl+F to find it ...

PR #4208, opened 2015-11-02
correct call of  use of CI->session->userdata()

juanitomint:
User data was called wrong in the Profiler class


narfbg:
How's that wrong?

PR #4212, opened 2015-11-04
Fix count_all_results when ordering

adam-bell:
Fixes error introduced in v3 when there is a pending qr_order_by (used typically in pagination to get a count before results)

Method adapted from v2 DB_active_rec line 911

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1379,9 +1379,20 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			$this->from($table);
 		}
 
+  		$tmp_order = $this->qb_orderby;
+        	$tmp_order_cache = $this->qb_cache_orderby;
+        	$this->qb_orderby = array();
+        	$this->qb_cache_orderby = array();
+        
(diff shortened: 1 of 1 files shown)


narfbg:
You're saying it fixes an error, but you're not saying what error that is ...

adam-bell:
Using the sqlsrv driver, query builder generates the following query for `count_all_results`

`SELECT COUNT(*) AS "numrows" FROM "tenants" LEFT JOIN "properties" "p" ON "p"."id" = "tenants"."property_id" WHERE "tenants"."deleted_date" =0 AND "tenants"."company_id" = '72' ORDER BY "tenants"."property_id" ASC`

In SQL server, this generates the error

`Column "tenants.property_id" is invalid in the ORDER BY clause because it is not contained in either an aggregate function or the GROUP BY clause`

In the previous version of CodeIgniter, the order by clause was omitted during compilation of the `count_all_results` query, to avoid this error.

MySQL seems more forgiving of this issue, but omitting the `ORDER BY` for `count_all_results` will not change the output for other database drivers

PR #4213, opened 2015-11-04
Fix issue introduced in v3 with QB Cache

adam-bell:
If CI_DB_query_builder qb_caching is TRUE and a query has been executed prior to the session save, the parameters can bleed into the session saving to DB.

Ideally, developers will use `stop_cache()` and `flush_cache()` but this issue did not occur in v2 and was not noted in upgrade instructions.

Only potential improvement to the pull request may be to store the cached query, flush, then reapply, in case the session write is called before the end of the page.

### DIFF
--- a/system/libraries/Session/drivers/Session_database_driver.php
+++ b/system/libraries/Session/drivers/Session_database_driver.php
@@ -230,6 +230,7 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 			return FALSE;
 		}
 
+		$this->_db->flush_cache();
 		$this->_db->where('id', $session_id);
 		if ($this->_config['match_ip'])
 		{


narfbg:
This stupid "caching" is the perfect example of user-requested functionality that should've been rejected.

But here's the thing ... while an easy `flush_cache()` call may prevent sessions from breaking, it may in fact cause you more trouble - when the session breaks, you *immediately know* there's something wrong, and that's not necessarily the case if you silently discard the cache.
So, while I agree that there's a problem, I'm not sure if there's an appropriate solution for it.

adam-bell:
Yeah - it's not ideal, but as the errors only occurred since the migration to CI v3, it should be mentioned in the upgrade docs or some fix pushed to the release

mwhitneysdsu:
Since the constructor of the session database driver throws an exception if $this->_db_cache_on, wouldn't it make sense for the driver to throw an exception if qb_caching is true, as well? That would probably require a check every time the driver tries to access the database, though, and I'm not really sure it's worth it just to remind someone not to call the session while they're using QB caching.

narfbg:
@adam-bell CI2 did utilize the query builder, so it's not correct to say that this is a breaking change in CI3. It may be when it affected *your* application, but it's not at all something new ...

Except of course that it is now a completely different Session library, which is very thoroughly explained in the upgrade instructions. We can't document every tiny detail about it.

Either way, I don't think this particular change is appropriate, so I'm closing the PR.

@mwhitneysdsu Such check would be useless, because this "cache" is started at runtime.

PR #4216, opened 2015-11-05
refactor exceptions for core/Lang

zhanghongyi:
related to issue #4189 

When the lang file could not be found, an exception will be thrown. Test cases already exist. By doing this, the error page would not be automatically rendered. If there is a need, user should call error page rendering in their exception handler(not sure if this is an issue). 

No particular changes are needed for the user guide.

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -150,7 +150,7 @@ class CI_Lang {
 
 		if ($found !== TRUE)
 		{
-			throw new RuntimeException('CI Error: Unable to load the requested language file: language/'.$idiom.'/'.$langfile);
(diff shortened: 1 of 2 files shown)


system/core/Lang.php
```diff
@@ -150,7 +150,7 @@ public function load($langfile, $idiom = '', $return = FALSE, $add_suffix = TRUE
 
 		if ($found !== TRUE)
 		{
-			show_error('Unable to load the requested language file: language/'.$idiom.'/'.$langfile);
+			throw new RuntimeException('CI Error: Unable to load the requested language file: language/'.$idiom.'/'.$langfile);
```
narfbg: Remove the 'CI Error: ' prefix, it's not necessary.

zhanghongyi: Fixed.

PR #4217, opened 2015-11-05
Build base_url correctly if SERVER_ADDR is IPv6

natesilva:
Now that `base_url` is being calculated using `SERVER_ADDR` instead of `HTTP_HOST`, it no longer works on IPv6 hosts. This fixes the problem by surrounding literal IPv6 addresses with proper [square brackets]. IPv4 addresses are not affected.

`http://::1/some/controller` ← this is not valid
`http://[::1]/some/controller` ← this is valid

### DIFF
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -98,6 +98,7 @@ class CI_Config {
 				{
 					$server_addr = $_SERVER['SERVER_ADDR'];
 				}
+
 				$base_url = (is_https() ? 'https' : 'http').'://'.$server_addr
 					.substr($_SERVER['SCRIPT_NAME'], 0, strpos($_SERVER['SCRIPT_NAME'], basename($_SERVER['SCRIPT_FILENAME'])));
 			}
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -90,7 +90,15 @@ class CI_Config {
 		{
 			if (isset($_SERVER['SERVER_ADDR']))
 			{
-				$base_url = (is_https() ? 'https' : 'http').'://'.$_SERVER['SERVER_ADDR']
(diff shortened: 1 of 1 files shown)


natesilva:
Note that this still doesn’t work if the server is not running on the default port number (a problem affects both IPv4 and IPv6). In theory we could fix this by using `$_SERVER['SERVER_PORT']`, but it is not safe to do so, [according to the PHP docs](http://php.net/manual/en/reserved.variables.server.php):

> Under the Apache 2, you must set UseCanonicalName = On, as well as UseCanonicalPhysicalPort = On in order to get the physical (real) port, otherwise, this value can be spoofed and it may or may not return the physical port value. It is not safe to rely on this value in security-dependent contexts.

system/core/Config.php
```diff
@@ -90,7 +90,15 @@ public function __construct()
 		{
 			if (isset($_SERVER['SERVER_ADDR']))
 			{
-				$base_url = (is_https() ? 'https' : 'http').'://'.$_SERVER['SERVER_ADDR']
+				if ((bool) filter_var($_SERVER['SERVER_ADDR'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6))
+				{
+					$server_addr = '[' . $_SERVER['SERVER_ADDR'] . ']';
```
narfbg: No spaces around concatenation please.

system/core/Config.php
```diff
@@ -90,7 +90,15 @@ public function __construct()
...
+				}
+				else
+				{
+					$server_addr = $_SERVER['SERVER_ADDR'];
+				}
```
narfbg: Needs an empty line after this one.

narfbg:
In addition to those, you may as well just `strpos()` for a colon character instead of using `filter_var()`.

PR #4218, opened 2015-11-05
Added MIME types for JPEG2000

schwemmer:
Added all MIME types and file extensions to support JPEG2000 images as
allowed file type.


narfbg:
Good, but the alignment is inconsistent with the other lines.

schwemmer:
Yes, I noticed, but only when I saw the diff on Github, didn't see it in Sublime Text nor the Github Desktop. Sorry for that, will fix it.

Issue #1327, opened 2012-05-04
Any way to set config items in a section?


michaelbrooks:
You can load config variables into sections with `$this->config->load('some_config', TRUE)`, and you can get these config variables using `$this->config->item('variable', 'some_config')`, but I can't find a way to set them manually using `$this->config->set_item()`, for example.

PR #4219, opened 2015-11-06
Added ability to set a config item thats contained in an array

REJack:


narfbg:
Duplicate of #2991.
Also relevant: #1327

PR #4220, opened 2015-11-07
refactor exceptions for Loader

zhanghongyi:
- Change show_error() to throw exceptions in system/core/Loader
- Changes in tests
    - "CI: Error" prefix is removed.
    - fixed one test of test_library(): the previous test tried to create a library without class, but it did not call `ci_vfs_create()` properly(no `ucfirst()` is used), so it actually was testing non-existent library files.
   - added a test case to test loading an empty library file. A RuntimeException with message `'Non-existent class: '.ucfirst($lib)` is excepted.
   - added a test case to test loading a new library with previously used object name. An exception with message `"Resource '".$obj."' already exists and is not a ".$class." instance."` is expected.
- No changes are needed for user guides

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -586,7 +586,7 @@ class CI_Loader {
 				$base_helper = BASEPATH.'helpers/'.$helper.'.php';
 				if ( ! file_exists($base_helper))
 				{
-					show_error('Unable to load the requested file: helpers/'.$helper.'.php');
+					throw new RuntimeException('Unable to load the requested file: helpers/'.$helper.'.php');
 				}
 
 				include_once($base_helper);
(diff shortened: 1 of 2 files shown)


narfbg:
The way I see it, the test was doing exactly what it was supposed to do.

Either way, the only changes to tests you're supposed to do here are those related to the exceptions change ... don't do anything else.

zhanghongyi:
okay, I won't modify test cases in the future pull requests of this task. 

But I would argue the test was not doing what it was supposed to do as far as I could tell. The first modified one wanted to test loading library without classes. So it created a empty file of library and tried to load it. It should expect an exception message of "Non existent class...". But the previous test created a file with the name "bad_test_lib.php" rather than "Bad_test_lib.php" so the framework could not find the file and return an exception message "Unable to load requested class...". It was actually testing non-existent files. 

And the last added test is aiming at testing one exception which is not tested before. For this one, I can remove it if you think enriching test cases is not the target of this issue. 

Will you accept these changes to tests if I separate them into another pull request?

narfbg:
Sure. I'm not saying they're bad changes, just that they don't belong in a PR that refactors error handling.

zhanghongyi:
referencing #4189

PR #4228, opened 2015-11-09
Update mimes

fcosrno:
Add Libre Office file types.

Reference: https://www.openoffice.org/framework/documentation/mimetypes/mimetypes.html


narfbg:
OpenOffice documents may also be detected as ZIP or XML (because that's what they in fact are) ...

PR #4229, opened 2015-11-10
refactor error handling in /core/Security

zhanghongyi:
- refactor error handling
- remove error message prefix
- no changes for user guide


narfbg:
This is a condition triggered by user input, and we don't have a mechanism to treat similar exceptions in a pre-defined way, so this one in particular should stay as is.

That also means #4200 can't be merged ... I was afraid something like this would come up, which is why I put it on hold for now.

PR #4235, opened 2015-11-11
New numeric_spaces filter on Form Validation

meletisf:
**Description**
The numeric_spaces rule returns true if it gets numbers and spaces.

**Use case scenarios**
A post code has 1 space in it, ex. 19 012 which causes the numeric filter to fail. In that case the numeric_spaces filter would return true.


narfbg:
There are many rules that would be similar to this one (e.g. somebody may use dashes instead of spaces) and we simply can't add countless rules for every specific use case. You can use regexp_match for that or create your custom rules.

PR #4239, opened 2015-11-12
Grammar correction in database configuration guide

craigdanj:
"it" changed to "in" on line 10. "in" seems like the correct word to be used there. Small change.


craigdanj:
Branch deleted

PR #4240, opened 2015-11-14
add unix socket support to redis session driver

sskaje:


system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -85,10 +85,24 @@ public function __construct(&$params)
 		{
 			log_message('error', 'Session: No Redis save path configured.');
 		}
+		elseif (strpos($this->_config['save_path'], 'unix://') === 0 && preg_match('#(?:unix://)?([^:?]+)(\?.+)?#', $this->_config['save_path'], $matches))
```
narfbg: The `strpos()` is unnecessary here ... the regular expression does the same check anyway.

narfbg: ... or it would do it, if you fix the expression to treat it as non-optional, which should be done anyway.

Also, `([^:?]+)` is just copied from the TCP version and it doesn't make sense here because you don't have a port number (and thus the colon character is not a separator), which also makes the next catching group unnecessary.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -128,7 +142,11 @@ public function open($save_path, $name)
 		}
 
 		$redis = new Redis();
-		if ( ! $redis->connect($this->_config['save_path']['host'], $this->_config['save_path']['port'], $this->_config['save_path']['timeout']))
+		if ($this->_config['save_path']['type'] == 'unix' && !$redis->connect($this->_config['save_path']['path']))
```
narfbg: We put spaces around the `!` operator, and preferrably use compare by type (`===` vs. `==`).

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -128,7 +142,11 @@ public function open($save_path, $name)
...
+		{
+			log_message('error', 'Session: Unable to connect to Redis with the configured settings.');
+		}
+		else if ( ! $redis->connect($this->_config['save_path']['host'], $this->_config['save_path']['port'], $this->_config['save_path']['timeout']))
```
narfbg: We also write `elseif`, no space. :)

narfbg:
In addition to the inline comments, a changelog entry is also necessary.

sskaje:
@narfbg code updated.

narfbg:
Still no spaces on *both* sides of `!` uses and still no changelog entry.

sskaje:
Ahhah, I thought changelogs are maintained by maintainers only.  
:D

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -128,11 +142,19 @@ public function open($save_path, $name)
...
+				log_message('error', 'Session: Unable to connect to Redis with the configured settings.');
+			}
+		}
+		elseif ( ! $redis->connect($this->_config['save_path']['host'], $this->_config['save_path']['port'], $this->_config['save_path']['timeout']))
 		{
 			log_message('error', 'Session: Unable to connect to Redis with the configured settings.');
 		}
-		elseif (isset($this->_config['save_path']['password']) && ! $redis->auth($this->_config['save_path']['password']))
+
+		if (isset($this->_config['save_path']['password']) && ! $redis->auth($this->_config['save_path']['password']))
```
narfbg: Changing this from an `elseif` to a separate `if` breaks the intended logic ... it will now try to authenticate even after the connect attempt has failed.

narfbg:
That's rarely the case, although I do have to correct almost any user-supplied one ...

PR #4244, opened 2015-11-16
Add encoding rule

toyamarinyon:
`^[`(ascii:27) -> `=1B`

This rule is require for e-mail written by Japanese characters that are
use `ISO-2022-JP`.


narfbg:
I went a bit further than that as I didn't like just adding one character specifically for Japanese ... see the commit referenced above. Thanks for the report.

PR #4246, opened 2015-11-16
Prevent Inspection Error

pittyplatsch:
Not completely sure what's going on here, but my IDE (PhpStorm) would show an inspection error here without the check.


narfbg:
Report it to phpStorm then ... CI handles this by a `function_exists()` check already.

PR #4251, opened 2015-11-18
Added alt attribute functionality …

dimitribalazs:
It hasn't been possible to set the alt attribute on the image. Added img_alt key on the $defaults array and integrated the $img_alt variable into the img string.


narfbg:
Why is the default changed from empty to "captcha"?

Issue #4255, opened 2015-11-20
xss_filtering (deprecated? )


DevotionGeo:
The warning message in the CodeIgniter's comments for $config['global_xss_filtering'] says,
" WARNING: This feature is DEPRECATED and currently available only for backwards compatibility purposes! "

What does that mean? Is only global xss filtering (setting xss filtering globally through config.php) deprecated or the xss filtering is completely deprecated and the method xss_filtering() will be removed from the security class completely?

cygmris:
Instead of enabling global xss filtering, consider using $this->securit->xss_clean() when rendering data which is a better way to not to mass up users' input,.
as [Documents](http://www.codeigniter.com/user_guide/libraries/input.html?highlight=xss) says,
> The ‘global_xss_filtering’ setting is DEPRECATED and kept solely for backwards-compatibility purposes.  XSS escaping should be performed on output, not input!

narfbg:
This is a bug tracker, not a Q&A website ... you know that.

PR #4256, opened 2015-11-20
Update config.php, form_helper.php and form_helper.rst (add form_email)

DevotionGeo:
The warning message should be clearer, because from the current message one can easily be mis-guided and think that the xss_filtering is completely removed, as I was misguided. Look at https://github.com/bcit-ci/CodeIgniter/issues/4255


DevotionGeo:
@narfbg the form_email function is so needed. A lot of time I use form_input and forget type="email" and miss the cool html5 validation for email field.

PR #4257, opened 2015-11-21
config directory constants

NoFeel:
when I needed to keep all my files in 3 settings folders for my different environments I threw me an error that could not read the " constants.php " file , because they always tried to read from the root "config /".

Hope that helps

Greetings


ivantcholakov:
-1

"... because they always tried to read from the root "config/" ..." - this is an essential idea. https://github.com/bcit-ci/CodeIgniter/commit/0d60a21187e2611cc4cf0ef2d7322c9b91c002b9

narfbg:
Duplicate of #3387.

PR #4259, opened 2015-11-22
Activated garbage collection

achraf52:
The garbage collection in the files driver doesn't seem to do it's job, so I'm calling it in the files driver constructor and it seems to work.

### DIFF
--- a/system/libraries/Session/drivers/Session_files_driver.php
+++ b/system/libraries/Session/drivers/Session_files_driver.php
@@ -97,6 +97,9 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 		{
 			$this->_config['save_path'] = rtrim(ini_get('session.save_path'), '/\\');
 		}
+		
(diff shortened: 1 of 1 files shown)


narfbg:
It's not supposed to be run on every request.

achraf52:
So how often and where does the function gets executed? I couldn't find any place in the CodeIgniter system files when it's periodically run.

narfbg:
http://php.net/manual/en/session.configuration.php#ini.session.gc-probability

achraf52:
I know the GC probability, but where it's implemented in CodeIgniter? I want the exact location of the system file that implements the GC probability mechanism to run the aforementioned function.

narfbg:
It's not in CodeIgniter, PHP runs it.

achraf52:
Hum... I tried changing the GC probability values in my PHP.ini file to be run 100% but it doesn't work, I'll do more tests and see if I can get positive results and get GC to run.

PR #4263, opened 2015-11-23
#4189 added :throws: directive to user guide for core/Config

DavidOtis:


narfbg:
You're using spaces as separators ... use tabs.

Also, the message should be reduced to:

> If no (valid) configuration file is found

... and that should really be applied to the docblock in the actual PHP file as well, I guess I missed that the first time around.

narfbg:
Sorry, but you've barely addressed anything ... I actually didn't notice that you used spaces for indentation, wasn't talking about that.

I'll leave inline comments.

user_guide_src/source/libraries/config.rst
```diff
@@ -204,6 +204,7 @@ Class Reference
 		:param	string	$item: config item name
 		:returns:	Config item value with a trailing forward slash or NULL if not found
 		:rtype:	mixed
+		:throws: RuntimeException   Throws a RuntimeException if no (valid) configuration file is found.
```
narfbg: You have to use a single tab on each side of ':throws:'

Also, I meant that you should remove the 'Throws a RuntimeException' part.

system/core/Config.php
```diff
@@ -115,7 +115,7 @@ public function __construct()
 	 * @param	bool	$use_sections		Whether configuration values should be loaded into their own section
 	 * @param	bool	$fail_gracefully	Whether to just return FALSE or display an error message
 	 * @return	bool	TRUE if the file was loaded correctly or FALSE on failure
-	 * @throws	RuntimeException		Throws a runtime exception if the config file can't be found or doesn't have a valid configuration array.
+	 * @throws	RuntimeException		Throws a RuntimeException if no (valid) configuration file is found.
```
narfbg: It appears that there are two tabs before the description here ... it should be only one.

narfbg:
"If no (valid) configuration file is found"

Exactly as I wrote it; no additions or changes to it

narfbg:
**I**f no (valid) configuration file is found

narfbg:
... both places.

PR #4265, opened 2015-11-23
Removed call to show_error() in core/URI. The function that ca…

DavidOtis:
…lled it, filter_uri(), doesn't have an entry in the user guide.


narfbg:
This is an error message displayed to users ... exceptions are not for that.

Also, please write proper issue descriptions instead of just trying to put it all in the title.

PR #4267, opened 2015-11-25
encodefromlegacy patch

renedekat:
Fixed the encode_from_legacy method as it didn't decode properly. The mcrypt_decode method has changed from the 1.x version.
This update includes legacy methods for decode and mcrypt_decode. They are declare private.

Also added public/private to all methods and updated some PHPDoc

### DIFF
--- a/system/libraries/Encrypt.php
+++ b/system/libraries/Encrypt.php
@@ -67,7 +67,7 @@ class CI_Encrypt {
 	 * @param	string
 	 * @return	string
 	 */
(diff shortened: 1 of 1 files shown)


narfbg:
CI2 is no longer supported, sorry.

renedekat:
Sure, but people might still want to migrate from older version and they'll run into the same issue as I did.

Perhaps you can merge it into the 3.x branch.

PR #4268, opened 2015-11-25
Update Encrypt.php

renedekat:
Fixed encode_from_legacy. The legacy decryption used a different mcrypt_decode function which is now in place.

### DIFF
--- a/system/libraries/Encrypt.php
+++ b/system/libraries/Encrypt.php
@@ -188,6 +188,43 @@ class CI_Encrypt {
 		return $this->mcrypt_decode(base64_decode($string), $this->get_key($key));
 	}
 
+		/**
+	 * Decode from legacy
+	 *
+	 * Reverses the above process
+	 *
+	 * @access	public
+	 * @param	string $string
+	 * @param	string $key
+	 * @return	string $legacy_mode
+	 */
+	private function _decode_from_legacy($string, $key = '', $legacy_mode = MCRYPT_MODE_ECB)
+	{
+		// set mode temporarily to what it was when string was encoded with the legacy
+		// algorithm - typically MCRYPT_MODE_ECB
+		$current_mode = $this->_get_mode();
+		$this->set_mode($legacy_mode);
+
+		$key = $this->get_key($key);
+
+		if (preg_match('/[^a-zA-Z0-9\/\+=]/', $string))
+		{
+			return FALSE;
+		}
+
+		$dec = base64_decode($string);
+
+		if (($dec = $this->_mcrypt_decode_from_legacy($dec, $key)) === FALSE)
+		{
+			return FALSE;
+		}
+
+		$this->set_mode($current_mode);
+
+		return $this->_xor_decode($dec, $key);
+
+	}
+
 	// --------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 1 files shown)


narfbg:
How's that fixing it?

renedekat:
It fixes it by using the legacy mcrypt_decode function which uses another $init_vector

I've added the output of a little test. The decoded legacy uses the old method and decoded new uses your encode_from_legacy up until the point that it starts encrypting it. If I were to use that it would encode �!D��\�3M*���4�  instead of Lloyds Bank.

Encrypted legacy: mKkflDfN7UjlTxkYc3uXljNUdhGYwZZeS17F5ueKoK0= 
Decoded Legacy: Lloyds Bank 
Decoded New: �!D��\�3M*���4� 

The patch is also future proof as it uses the legacy methods to decode (not publicly exposed). So if you want to change the mcrypt_encode function for some reason, that would also mean you would change the mcrypt_decode which will again break the encode_from_legacy. After all it uses the NEW mcrypt_decode function instead of the legacy one.

narfbg:
Perhaps I was unclear with my question, sorry about that. What I really meant is - what is the problem that you're trying to fix?

The only thing I get from your last comment is that it uses a different IV, but that's during *decryption*, which can't possibly help with anything.

renedekat:
I had to use the encode_from_legacy method, but that didn't decrypt it properly so all the encrypted data was useless. 

With my fix for the encode_from_legacy function, the data that was encrypted using CI 1.7.2 is now working with the latest CI.

renedekat:
Just to summarise:

Project in CI 1.7.2 which had encrypted data.

Upgrade the CI framework which requires that encrypted data be RE-encrypted using the "encrypt_from_legacy" function.
That function didn't produce the correct result and is fixed by my patch.

I don't know how I can make myself clearer than that.

narfbg:
Ok ... but that gives me as much info as "it didn't work".

The IV that you're now generating on the fly is irrelevant in ECB mode, and it can only break things in other modes - so that's neither what helping you, nor anything that makes sense for me to merge.

The other changes look like just moving code around to me ... The only thing I suspect may be relevant is the `get_key()` call point, but that doesn't explain why are you making that many alternations to the code instead of just moving that one line.

narfbg:
Of course, it doesn't help that I don't have a CI1.7.whatever version to look at ...

renedekat:
The old mcrypt_decode function is defined as:

	private function mcrypt_decode($data, $key)
	{
		$init_size = mcrypt_get_iv_size($this->_get_cipher(), $this->_get_mode());
		$init_vect = mcrypt_create_iv($init_size, MCRYPT_RAND);
		return rtrim(mcrypt_decrypt($this->_get_cipher(), $key, $data, $this->_get_mode(), $init_vect), "\0");
	}

Which properly decrypts strings that were encrypted using the 1.7.x branch.

The new mcrypt_decode function returns a different result when using legacy encrypted data, which is what you do when using encode_from_legacy.

As I said earlier, the code is now also future proof as you can change the mcrypt_encode and mcrypt_decode functions at will. It won't affect the encode_from_legacy function for people that are trying to upgrade.

narfbg:
Well, I don't have a 1.7.2, but I *do* have a 1.6.2, which contains the following:

<pre>
        function mcrypt_decode($data, $key)
        {
                $data = $this->_remove_cipher_noise($data, $key);
                $init_size = mcrypt_get_iv_size($this->_get_cipher(), $this->_get_mode());

                if ($init_size > strlen($data))
                {
                        return FALSE;
                }

                $init_vect = substr($data, 0, $init_size);
                $data = substr($data, $init_size);
                return rtrim(mcrypt_decrypt($this->_get_cipher(), $key, $data, $this->_get_mode(), $init_vect), "\0");
        }
</pre>

So, I find it quite strange that 1.7.2 would have the code that you've pasted ... it just doesn't make sense.

renedekat:
Well the version I am working with uses that mcrypt_decode function. Not sure what happened, but this patch fixed it for me.

narfbg:
Ok ... I found an 1.7.2 in my archives and I can confirm that this is what its `mcrypt_decode()` function looks like:

<pre>
        function mcrypt_decode($data, $key)
        {
                $data = $this->_remove_cipher_noise($data, $key);
                $init_size = mcrypt_get_iv_size($this->_get_cipher(), $this->_get_mode());

                if ($init_size > strlen($data))
                {
                        return FALSE;
                }

                $init_vect = substr($data, 0, $init_size);
                $data = substr($data, $init_size);
                return rtrim(mcrypt_decrypt($this->_get_cipher(), $key, $data, $this->_get_mode(), $init_vect), "\0");
        }
</pre>

So I can only assume that you're using a manually modified version.

renedekat:
Ok, thanks for that. It must have been changed long ago before I started working on it.

Apologies for the confusion.

PR #4274, opened 2015-11-27
p.footer -> .footer

Von101:
Element (p.footer) is overqualified.


narfbg:
There's no reason to make this change.

PR #4275, opened 2015-11-27
Create foreign keys with CI_DB_Forge

greensn0w:
Given we have this table:

```
Table: assignments
+-----------+------+--------+-----------+-------+
|   Field   | Type | Length | Unsigned? | Null? |
+-----------+------+--------+-----------+-------+
| id        | INT  |     11 | Y         | N     |
| parent_id | INT  |     11 | Y         | Y     |
+-----------+------+--------+-----------+-------+
```

Until now you had to add a foreign key like this:
```php
[...]
$this->dbforge->create_table('assignments');
$this->db->query('ALTER TABLE `assignments` ADD FOREIGN KEY (`parent_id`) REFERENCES `assignments` (`id`)');
```

With this PR you can use a convenience function to add a foreign key:
```php
$this->dbforge->add_foreign_key('parent_id', 'assignments', 'id');
$this->dbforge->create_table('assignments');
```
<hr/>

`->add_foreign_key(..)` takes these arguments:
```php
$this->dbforge->add_foreign_key($index_name, $reference_table, $reference_column, $on_update = 'no action', $on_delete = 'no action')
```
The case of `$on_update` and `$on_delete` is ignored and will be converted to upper case.
<hr/>

All key "blueprints" are stored in an array.

No key is created until `->create_table(...)` is executed

### DIFF
--- a/system/database/DB_forge.php
+++ b/system/database/DB_forge.php
@@ -75,6 +75,12 @@ abstract class CI_DB_forge {
 	public $primary_keys	= array();
 
 	/**
+	 * Foreign Keys data
+	 * @var array
+	 */
+	public $foreign_keys	= array();
+
+	/**
 	 * Database character set
 	 *
 	 * @var	string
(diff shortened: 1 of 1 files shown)


narfbg:
Ignoring smaller details, your patch only works for MySQL - it must work for all drivers.

steve-goodwin:
Any further work on this sn0w? This is a feature i've been waiting for :)

sn0w:
@steve-goodwin:
I have no idea how to port this to other drivers, as I'm only using MySQL :disappointed: 

Anyway I could create a library or a `MY_DB_Forge` class and postpone the PR.

We could then eventually merge this back into the core as soon as all drivers are supported.

PR #4278, opened 2015-11-29
add documentation of exceptions for Loader

zhanghongyi:
add documentation for #4220 
part of #4189 

In Loader.php, documentation is added for every function which directly triggers an exception.

In the user guide, documentation of exceptions is added for every function which is possible to throw an exception during the execution(directly or indirectly through internal functions).

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -230,6 +230,10 @@ class CI_Loader {
 	 * @param	string	$name		An optional object name to assign to
 	 * @param	bool	$db_conn	An optional database connection configuration to initialize
 	 * @return	object
+	 * @throws	RuntimeException	If the name of model is already being used by another resource
+	 * @throws	RuntimeException	If model file exists but does not declare class
+	 * @throws	RuntimeException	If unable to load specified model
+	 * @throws	RuntimeException	If class already exists but does not extend CI_Model
 	 */
 	public function model($model, $name = '', $db_conn = FALSE)
 	{
@@ -558,6 +562,7 @@ class CI_Loader {
 	 *
 	 * @param	string|string[]	$helpers	Helper name(s)
 	 * @return	object
+	 * @throws	RuntimeException	If unable to load the requested file
 	 */
 	public function helper($helpers = array())
 	{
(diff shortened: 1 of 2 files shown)


zhanghongyi:
Any suggestion for this PR? @narfbg

narfbg:
"If unable to load the requsted file" is a bit too generic to be used in multiple places.

However, I didn't comment so far because I'm not completely sure about another thing ... I don't quite like the multiple `@throws` and `:throws:` per single method.

zhanghongyi:
Based on PHP's exception documentation https://pear.php.net/manual/en/rfc.error-handling.documentation.php, it is allowed to use mutiple `@throws` per single method to document different usages of exceptions. Though I am not sure about the sphinx documentation. In [documentation of sphinxcontrib-phpdomain](http://pythonhosted.org/sphinxcontrib-phpdomain/reference.html#directive-php:method), it does not require a description for exception. 

So maybe we could keep multiple `@throws` in php file and remove the descriptions in user guide to keep it simple?

narfbg:
I'm sure multiple instances are allowed, but I'm not sure I actually want them. :)

I'd probably treat it in the same way as `@returns` - come up with a short, but clear generalization of all the conditions, like "In case of failure". The specific reason for that failure is provided by the exception message anyway, so we don't need to be overly-verbose with the documentation.

zhanghongyi:
I have changed the multiple `throws` into the simple version "In case of failure" as suggested. But I am not sure about your concerns of the generic message "unable to load the requested file". Should that be changed as well?

narfbg:
Well, if it's written for the `view()` method, it could say "view" instead of "file". Same for `helper()`, where in fact two files may be loaded at once. And so on ...

If the message was again "In case of failure", that'd be ok, but a "file" may have different meanings depending on the context. I suppose it's not incorrect per se, but it just feels a bit off.

zhanghongyi:
I have updated the user guide message to be "In case of failure" to be more generic. But for the php documentation, I would suggest to keep the message "If unable to load requested file" because it correspond to the exact exception message.

PR #4281, opened 2015-12-02
Allowing redis cache to search keys and return their values

davidplappert:
This code update allows those who use Redis to search for keys and get their values returned. It helps when you branches in Redis for storing values.

Code Example:
```
$this->load->driver('cache',array('adapter' => 'redis'));
$this->cache->save('user-5:event-1', $event1, 0);
$this->cache->save('user-5:event-2', $event2, 0);
$this->cache->get("user-5:event-*");
//returns values for both 'user-5:event-1' and 'user-5:event-2'
```


narfbg:
Sorry, but no ... Whatever you'd be using this for, there surely are simpler ways for it, and we don't accept driver-specific features anyway.

PR #4286, opened 2015-12-03
My First Commits

Sosiashvili:
My First Commit ^_^


jim-parry:
A pull request out of the blue that has no functional impact, and the purpose of which is your practicing github? See https://github.com/bcit-ci/CodeIgniter/blob/develop/contributing.md and http://www.codeigniter.com/user_guide/contributing/index.html#tips-for-a-good-issue-report

PR #4287, opened 2015-12-03
Load compatibility features

Sosiashvili:
I think it is more convenient


jim-parry:
More convenient for you, perhaps, but personal preference.
The change does not materially affect the framework.
It looks like you are changing for the sake of changing, not to solve a problem.

PR #4289, opened 2015-12-04
update return function mysql_to_unix

phlyper:
problem when using with **php 7**

### DIFF
--- a/system/helpers/date_helper.php
+++ b/system/helpers/date_helper.php
@@ -389,12 +389,12 @@ if ( ! function_exists('mysql_to_unix'))
 
 		// YYYYMMDDHHMMSS
 		return mktime(
-			substr($time, 8, 2),
-			substr($time, 10, 2),
-			substr($time, 12, 2),
-			substr($time, 4, 2),
-			substr($time, 6, 2),
-			substr($time, 0, 4)
+			intval(substr($time, 8, 2)),
+			intval(substr($time, 10, 2)),
+			intval(substr($time, 12, 2)),
+			intval(substr($time, 4, 2)),
+			intval(substr($time, 6, 2)),
+			intval(substr($time, 0, 4))
 		);
 	}
 }


narfbg:
I don't see a problem: https://3v4l.org/pPbmW

peternijssen:
As soon as the seconds are gone, it breaks in PHP 7, while running fine in PHP 5.6; https://3v4l.org/IeeTf

narfbg:
You're not supposed to pass a timestamp that doesn't have seconds ...

peternijssen:
I know. But the method allows me to .... ;) a typecast/intval would be a smart choice because ints are expected.

huglester:
So I have the same problem here. https://github.com/bcit-ci/CodeIgniter/issues/4533
It was running as expected in php 5.3+ up to php 5.6

This won't be fixed? and left 'as-is' untill more issues will be created?

narfbg:
It's not broken, you're not using it properly.

And this is already obvious by the previous comments ... asking these questions won't change anything.

enlivenapp:
I'm certainly not trying to get a mud slinging contest started but I did want to ask a pretty specific question.

I understand the OP wasn't using the function correctly, (not including the seconds) and looking at @huglester's issue which he didn't really explain anything except it didn't work.  

So, how are people using it incorrectly (and please explain how to use it correctly) when in the function it's almost all string manipulation:

    $time = str_replace(array('-', ':', ' '), '', $time); 

Then when working directly with `mktime()` directly, it's using `substr()`.  So, as far as I can tell, it's type cast from the get-go as a string with no change of using integers.

Am I completely off base here?

narfbg:
Yes you are. Passing a timestamp without seconds is the incorrect usage ...

Honestly, I don't understand what's so hard to get about this. The UNIX timestamp *itself is a seconds count*.

enlivenapp:
I think you misunderstood what I'm saying.  I'm agreeing the OP used it incorrectly by not including the seconds which I already mentioned.

What I'm getting at is `mktime()`  params *requires* input as `(int)` in PHP7 and the current `mysql_to_unix()` function passes it *strings* simply because it's using str_replace().

to use the `mysql_to_unix()` function we *must* pass a string to it which immediately fails with `mktime()`.

So, I'll ask again, if we're using it incorrectly, how do we use it correctly?

narfbg:
<pre>
$ php -a
Interactive mode enabled

php > var_dump(PHP_VERSION);
string(22) "7.0.8-0ubuntu0.16.04.3"
php > var_dump(mktime('1', '2', '3', '4', '5', '6'));
int(1144188123)
php > 
</pre>

The problem that you're describing doesn't exist.

astrolac:
Ok! As we can see in PHP manual mktime expect to receive int arguments.
Let me ask, why you put strings into it? I suppose that mktime can convert strings to int in some cases, but what if it will be empty string or false? I think the right way is a direct convert strings into int in mysql_to_unix:
```
		return mktime(
			intval(substr($time, 8, 2)),
			intval(substr($time, 10, 2)),
			intval(substr($time, 12, 2)),
			intval(substr($time, 4, 2)),
			intval(substr($time, 6, 2)),
			intval(substr($time, 0, 4))
		);
```
I understand that mysql_to_unix expect to receive MySQL Timestamp as string 'YYYY-MM-DD HH:MM:SS'. But maybe will be better if mysql_to_unix will check incoming string and throw the exception if something wrong with it? Оtherwise some developers who use CI (and me too) can have some misunderstanding seeing this error.
In addition, with direct convert, if substr return false and input string format was not be checked, intval just give 0 and such error will not occur (it will be mean that hours, minuts and seconds is equal to 0).

narfbg:
> As we can see in PHP manual mktime expect to receive int arguments.
> Let me ask, why you put strings into it?

Because it expects int-*like* values; i.e. the string '2017' works just as well as the 2017 integer. Explicit casting, especially via `intval()`, adds unnecessary overhead.
If this wasn't the case, this function would've never worked.

Typing isn't the problem; why do I have to explain this countless times?

> I understand that mysql_to_unix expect to receive MySQL Timestamp as string 'YYYY-MM-DD HH:MM:SS'. But maybe will be better if mysql_to_unix will check incoming string and throw the exception if something wrong with it? Оtherwise some developers who use CI (and me too) can have some misunderstanding seeing this error.

*'YYYY-MM-DD HH:MM:SS' is the MySQL timestamp format*. The function is obviously designed to accept only this format, and it's completely natural that there's a problem if there's a mismatch.

Can we check if there's a format mismatch? Perhaps, a regular expression could both check and catch the numbers for processing. But a mismatch should cause a boolean FALSE to be returned, not an exception.

> In addition, with direct convert, if substr return false and input string format was not be checked, intval just give 0 and such error will not occur (it will be mean that hours, minuts and seconds is equal to 0).

This means sweeping it under the rug; just because you don't see an error message, that doesn't mean there isn't a problem. Doesn't matter how many people suggest this, it's not happening.

narfbg:
... and when you look at this function carefully, it turns out to be a can of worms that I'm not sure we want to keep at all.

It is old, from pre-PHP-5.2 times, probably pre-PHP-4 too. Nowadays, everybody can do this:

    DateTime::createFromFormat(
        'Y-m-d H:i:s',
        $mysqlTimestamp,
        new DateTimeZone('UTC') // The MySQL timestamp format DOES assume UTC
    )
    ->format('U');

I've written that on multiple lines for brevity, but it is in fact a one-liner.

Also, see that comment about MySQL timestamps and UTC above? `mysql_to_unix()` doesn't take that into account; unless you've set your `date.timezone` is set to UTC, the function is essentially broken, as `mktime()` will take into account your local timezone.

PR #4291, opened 2015-12-06
fix phpdoc

b-kaxa:
fix phpdoc with Router.php, URI.php

This PR to big open source project is for the first time for me.
So I will apologize if I mistake PR process.

### DIFF
--- a/system/core/Router.php
+++ b/system/core/Router.php
@@ -118,6 +118,7 @@ class CI_Router {
 	 *
 	 * Runs the route mapping function.
 	 *
+	 * @param array $routing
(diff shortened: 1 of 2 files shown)


system/core/Router.php
```diff
@@ -118,6 +118,7 @@ class CI_Router {
 	 *
 	 * Runs the route mapping function.
 	 *
+	 * @param array $routing
```
narfbg: We use tabs for separation.

b-kaxa: Oh, sorry.

PR #4292, opened 2015-12-06
Refactor exception handling in database/DB

DavidOtis:


narfbg:
You need to alter the appropriate unit tests as well ... they are now failing with your changes.

system/database/DB.php
```diff
@@ -193,7 +194,11 @@ class CI_DB extends CI_DB_driver { }
 	// Load the DB driver
 	$driver_file = BASEPATH.'database/drivers/'.$params['dbdriver'].'/'.$params['dbdriver'].'_driver.php';
 
-	file_exists($driver_file) OR show_error('Invalid DB driver');
+	if( ! file_exists($driver_file))
```
narfbg: There needs to be a space between if and the opening parenthesis, i.e. `if (`

system/database/DB.php
```diff
@@ -193,7 +194,11 @@ class CI_DB extends CI_DB_driver { }
...
+	{
+		throw new RuntimeException('Invalid DB driver');
+	}
+
```
narfbg: You're inserting whitespace here; the line should be completely blank.

system/database/DB.php
```diff
@@ -114,7 +115,7 @@ function &DB($params = '', $query_builder_override = NULL)
 		 */
 		if (($dsn = @parse_url($params)) === FALSE)
 		{
-			show_error('Invalid DB Connection String');
+			throw new RuntimeException('Invalid DB Connection String');
```
narfbg: This should be an `InvalidArgumentException` instead.

system/database/DB.php
```diff
@@ -47,6 +47,7 @@
  * @param 	string|string[]	$params
  * @param 	bool		$query_builder_override
  *				Determines if query builder should be used or not
+ * @throws	RuntimeException	If (valid) configuration file or DB driver doesn't exist.
```
narfbg: Just 'In case of failure'.

PR #4296, opened 2015-12-07
Detecting if database connection was not successful

halmai:
Hi All, 
I had a problem when I wanted to connect to the database from my code. I used this line:

    $this->load->database();

Unfortunately this way I cannot check if the database connection was successful because the `database()` method of the CI_Loader class doesn't return FALSE in case of the faulty connection attempt. This is impractical and contradicts to the comment of the method:

     * @return    object|bool    Database object if $return is set to TRUE,
     *                    FALSE on failure, CI_Loader instance in any other case

So, I added the appropriate check before returning from the method. 

After this, I can check the connection in my code this way:

    $res = $this->load->database();
    if (!$res) 
    {
        throw new Exception("Database connection error");
    }

I hope this modification helps other CI users. I know that the migration to throwing exceptions is in progress for the next major CI versions but this modification is a small one and it doesn't require modifying the existing projects relying on the present version of CI. So, I think the next minor version change could contain this bugfix. It would make me really happy. :)

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -361,6 +361,10 @@ class CI_Loader {
 
 		// Load the DB class
 		$CI->db =& DB($params, $query_builder);
+		if (!$CI->db->conn_id) 
+		{
+			return FALSE;
+		}
 		return $this;
 	}
 


narfbg:
A "failure" in this context means failure to instantiate a database driver, not failure to connect.

Either way, such an exception should be thrown by the `DB()` function and not the loader class. It should also be a `RuntimeException` and needs to be documented (while here you're just patching the code). If you wish to submit another pull request to account for these notes, please also read our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html) as you'd need to follow that as well.

halmai:
Sorry for the code styling thing, I thought I was careful enough. Next time I will be even more accurate.

About the solution: Why do you prefer exceptions? In general I prefer them too but the present version of CI uses FALSE values to indicate errors. I wanted to follow this pattern. 

I think the general exception-based behaviour of CI should come with a more major version number jump, for example 4.0, while this little bugfix should be compatible with the existing CI applications. That's why I used this approach. 

The exception-based solution should be treated like this:

    try 
    {
        $this->load->database();
    } 
    catch (Exception $e)
    {
        // handle error
    }
    

With my proposal, it should be handled like this:

    if ( ! $this->load->database())
    {
        // handle error
    }
 
I think the latter one is closer to the usual way how we handle errors in CI. What do you think?

If you still prefer the exception-based solution then I will do it. In that case I would replace the 

    return FALSE;

with 

    throw new RuntimeException('Database connection failure.');

Do you want me to do this?

narfbg:
> About the solution: Why do you prefer exceptions? In general I prefer them too but the present version of CI uses FALSE values to indicate errors. I wanted to follow this pattern.

CI hasn't taken advantage of exceptions up until now, but in many cases it really has to, which is why we're committed to doing it in 3.1.

> I think the general exception-based behaviour of CI should come with a more major version number jump, for example 4.0, while this little bugfix should be compatible with the existing CI applications. That's why I used this approach.

4.0 will be a complete rewrite, so if we follow that logic, it means no behavioral improvements whatsoever - that's just not an option. We haven't pledged to follow SemVer and until the 4.0 release, we'll just treat 3.1, 3.2, etc. as major releases.

> I think the latter one is closer to the usual way how we handle errors in CI. What do you think?
>
> (... and the rest of your reply ...)

I assume you didn't really understand my previous comment.

If `CI_Loader::database()` returns FALSE, it means that it failed to load the appropriate database class - it should not have any knowledge of a connection status. That is why returning FALSE (or throwing an exception) in *there* is not an option.
If anything fails, it should be a function located somewhere under system/database/, and for your use-case to be handled in that scenario, an exception is the only option.

halmai:
> If CI_Loader::database() returns FALSE, it means that it failed to load the appropriate database class - it should not have any knowledge of a connection status. That is why returning FALSE (or throwing an exception) in there is not an option.

I can already see, thanks for re-explaining. The misleading thing was that the `...->load->database()` not only *loads* the database class but it *connects* to the db too. This contradicts to the Separation of Concerns principle but I can live with this. The point is that this makes it difficult to return information about two different types of errors:
1. not finding the class 
2. finding the class but failing to connect to the db with it.

I already understand that you don't want to mix the report of these errors. 

> If anything fails, it should be a function located somewhere under system/database/, and for your use-case to be handled in that scenario, an exception is the only option.

OK, then let's throw an exception from the `initialize()` function of the `CI_DB_driver` class, instead of the present `return FALSE`. OK? After this, the developer can check the above possible errors in the following way:
1. if FALSE returned then the class was not found
2. if exception was thrown then the connection was not successful. 
Is this what you would prefer?

narfbg:
> I can already see, thanks for re-explaining. The misleading thing was that the ...->load->database() not only loads the database class but it connects to the db too. This contradicts to the Separation of Concerns principle but I can live with this.

From a practical POV it does indeed do that, but code-wise speaking, `CI_Loader::database()` doesn't initialize connections; the DB classes do that themselves - there's no separation of concerns issue with that.

> OK, then let's throw an exception from the initialize() function of the CI_DB_driver class, instead of the present return FALSE. OK? After this, the developer can check the above possible errors in the following way:
1. if FALSE returned then the class was not found
2. if exception was thrown then the connection was not successful.
Is this what you would prefer?

The thing is ... we're in the middle of refactoring a lot of stuff to throw exceptions and failure to load a database class MUST throw one, so we'll probably end up with exceptions in both cases.

`CI_DB_driver::initialize()` throwing a `RuntimeException` is fine, yes.

halmai:
I created the new Pull Request as we discussed. Could you check it if you are happy with it, please? Thanks.

PR #4298, opened 2015-12-07
System and Application folder should have full directory path.

vrkansagara:
System and Application folder should have full directory path.


narfbg:
They are already converted to absolute paths ... that's why index.php is over 300 lines long.

vrkansagara:
No man that it not converted to absolute path you explicit add the __DIR__ for that.

I think this issue would be open so other member can discuss on it.

narfbg:
Yes it is, right here: https://github.com/bcit-ci/CodeIgniter/blob/514896e1b0df12bc82dac785fa7c9ed4f62585e6/index.php#L198

PR #4299, opened 2015-12-07
#4189 Refactor error handling in database/DB_utility

DavidOtis:


zhanghongyi:
Will it be more appropriate to throw InvalidArgumentException?

narfbg:
`InvalidArgumentException` would be the proper one here.

However, it would be even better if the parameter is typed to `CI_DB_result`, which would eliminate the need for these `if` conditions ... I'll do that later.

PR #4300, opened 2015-12-08
throwing RuntimeException instead of returning FALSE

halmai:
As we discussed yesterday with narfbg, I modified my last solution. Now we throw a RuntimeException instead of returning FALSE.

I also added this modification to the changelog.

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -361,6 +361,10 @@ class CI_Loader {
 
 		// Load the DB class
 		$CI->db =& DB($params, $query_builder);
+		if (!$CI->db->conn_id) 
+		{
+			return FALSE;
+		}
 		return $this;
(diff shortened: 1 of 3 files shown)


narfbg:
That's good so far (changelog message needs some styling), but with an exception thrown there, you may as well remove the `log_message()` call and the `db_debug` condition - an uncaught exception will both log and display a fatal error in dev mode.

narfbg:
On another note ... I just noticed a call to `initialize()` that obviously will need to be handled differently with this change, and there are probably more similar cases to cover.

If you don't mind, I'll manually pull your commit (so you are credited in the commits) and will refactor the rest myself. Is that ok with you?

halmai:
Yes, thank you for all. :)

PR #4304, opened 2015-12-08
Fix mismatched brackets in documentation

scherepn:
Sphinx generates incomplete/incorrect HTML output because there are
mismatched brackets in the documentation.

Signed-off-by: Phil Scherer <pscher1@umbc.edu>

### DIFF
--- a/user_guide_src/source/libraries/sessions.rst
+++ b/user_guide_src/source/libraries/sessions.rst
@@ -842,7 +842,7 @@ Class Reference
 		.. note:: This method is DEPRECATED. Use ``userdata()``
 			with no parameters instead.
 
-	.. php:method:: get_userdata()
+	.. php:method:: &get_userdata()
 
 		:returns:	A reference to ``$_SESSION``
 		:rtype:	array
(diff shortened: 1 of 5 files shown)


user_guide_src/source/libraries/sessions.rst
```diff
@@ -842,7 +842,7 @@ Class Reference
 		.. note:: This method is DEPRECATED. Use ``userdata()``
 			with no parameters instead.
 
-	.. php:method:: &get_userdata()
```
narfbg: This ampersand is not here by accident, why are you removing it?

I don't see this, nor any of the extra/missing brackets causing build errors or incorrect HTML. Inaccurate documentation - sure, but nothing really broken.

scherepn: Sorry, I included this file by accident. I made this change while debugging and forgot to revert. I was tracking down an issue with the upstream sphinx php plugin. It can't handle method "names" that start with an '&'.

The resulting documentation looks ok-ish when rendered by a browser, but it's missing permalinks, and syntax highlighting. As an example, here's the HTML that's generated by well-formed documentation:
````HTML
<dt id="CI_DB_driver::simple_query">
  <code class="descname">simple_query</code>
  <span class="sig-paren">(</span>
  <em>$sql</em>
  <span class="sig-paren">)</span>
  <a class="headerlink" href="#CI_DB_driver::simple_query" title="...snip...">¶</a>
</dt>
````

Here's what mismatched brackets generate:
````HTML
<dt id="CI_DB_driver::simple_query">
  <code class="descname">simple_query($sql])</code>
</dt>
````

PR #4305, opened 2015-12-09
Fix alignment

huaiping:
Fix alignment


narfbg:
That's not fixing it ... it's just adjusting it to your personal preferences.

Issue #4187, opened 2015-10-22
Drivers not fully documented


Charles-Garrison:
In order to get a driver working properly, you must create an array of valid drivers, however this can not be found anywhere in the documentation.

Example of setting the valid drivers array
----------------------------------------------------
$this->valid_drivers = array( 'diver_name', 'another_driver_name' );

Documentation URL
--------------------------
http://www.codeigniter.com/user_guide/general/creating_drivers.html

narfbg:
You say that like the page in question says anything useful at all ... it doesn't. :)

These "drivers" have proven to be confusing for people, and not at all as useful as initially thought. As such, I don't want more additions to encourage people to use them. If the Cache library didn't use `CI_Driver`, I'd erase all of it right now.

Charles-Garrison:
Hi Andrey,

Although I understand what you are saying, and I tend to mostly agree, I do
feel like leaving the documentation incomplete is more confusing an a
bigger issue than encouraging people to use legacy code.  What about a
middle-ground solution?  Fully documenting the usage, but including message
that discourages the usage of Drivers since they will likely be phased out
in the coming versions?

Thanks,

Charles

narfbg:
Well, I could agree to that, but here's the problem ... I'm not sure if there's anybody left who knows exactly how the "drivers" work in order to document them. You'd probably expect me to know it, but I don't; I don't even remember who introduced the feature.

Charles-Garrison:
Fair enough. I don't know how to use the documentation system. So if I documented the drivers, would you consider plugging it in?






On Mon, Oct 26, 2015 at 3:22 AM -0700, "Andrey Andreev" <notifications@github.com> wrote:












Well, I could agree to that, but here's the problem ... I'm not sure if there's anybody left who knows exactly how the "drivers" work in order to document them. You'd probably expect me to know it, but I don't; I don't even remember who introduced the feature.

narfbg:
Given that your additions are suitable - sure, I'll consider it.

xiers:
Hi, there are some examples for CI.
https://github.com/oliverhr/CI2_Drivers_example/
https://github.com/xiers/CI3_Drivers_example/

 useful ?

PR #4314, opened 2015-12-14
Bugfix: child class in valid drivers

skunkbad:
$child will never be in $valid_drivers unless backwards compatibility with CI2 and the docs are wrong.

### DIFF
--- a/system/libraries/Driver.php
+++ b/system/libraries/Driver.php
@@ -103,7 +103,7 @@ class CI_Driver_Library {
 		$child_name = $this->lib_name.'_'.$child;
 
 		// See if requested child is a valid driver
-		if ( ! in_array($child, $this->valid_drivers))
+		if ( ! in_array($child_name, $this->valid_drivers))
 		{
 			// The requested driver isn't valid!
 			$msg = 'Invalid driver requested: '.$child_name;
--- a/tests/codeigniter/libraries/Driver_test.php
+++ b/tests/codeigniter/libraries/Driver_test.php
@@ -31,7 +31,7 @@ class Driver_test extends CI_TestCase {
 	public function test_load_driver()
 	{
 		// Create driver file
-		$driver = 'basic';
+		$driver = 'Driver_basic';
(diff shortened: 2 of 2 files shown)


skunkbad:
I don't have time to keep messing around with the tests. The reason for the pull request is that Drivers usage is not the same between CI2 and CI3, but the docs are the same. So either the docs are wrong for CI3, or there's a need to go in and make drivers work like CI2. The problem being that the array of valid drivers has changed the way valid drivers are listed.

narfbg:
What part of the documentation suggests this?

skunkbad:
Actually, now that you mention it, the documentation for drivers is so poor, that one would have to resort to tutorials on the internet, or asking questions on the forum to get them working. So you are right, and I am wrong, because the documentation doesn't ever mention that the $valid_drivers array exists, or how to use it.

That said, anyone who used CI2 drivers successfully would have their drivers immediately fail on CI3 because the usage of $valid_drivers has changed. Nowhere on the Upgrading from 2.2.x to 3.0.x page is this mentioned, which isn't a surprise because $valid_drivers isn't mentioned anywhere else in the docs.

It would be really nice to include a little documentation about the $valid_drivers array, because that is an essential thing to know about drivers. Is mere documentation the proper solution, or should the CI3 valid drivers be set the same was as CI2?

narfbg:
Changing the code to match CI2's behavior would obviously be a BC break for CI3 at this point, so closing the PR.

However, you might find the discussion in #4187 useful.

skunkbad:
I see that you don't like the Drivers feature, and it's your opinion that it may not be as useful as originally thought, and that you'd just like to get rid of the Drivers feature. That sounds like a great solution, because nobody is willing to make documentation changes that allow it to be easily used, and in your own words, you don't even know how it works.

I used it once, and it wasn't anything that couldn't have been accomplished in others ways. I do like that using drivers kept all of the files in a single sub-directory of application/libraries.

PR #4320, opened 2015-12-17
allow adding variables to the language line

avenirer:
I would like to propose the use of "placeholders" when needing to insert variables inside a translated string. This will be done like in example below:
```php
//inside the language file...
$lang['homepage_hello_known'] = 'Hello, [username]! How are you today? ';
```php

```php
//when calling the language line
$this->lang->line('homepage_hello_known',TRUE,array('username'=>'a username'));
```

If it will be approved, I will work on the documentation

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -200,7 +200,7 @@ class CI_Lang {
 		
 		if(!empty($placeholders))
 		{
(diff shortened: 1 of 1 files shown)


narfbg:
Duplicate of #1502, #3223.

avenirer:
If it is a duplicate, does that mean that I should work on the original proposals (either #1502, or #3223) to see it implemented? or?...

PR #4322, opened 2015-12-21
Fixes PDO connect when DSN is provided without username & password

ricpelo:
If $db['default']['dsn'] is set (including 'user' and 'password' fields), but $db['default']['username'] == '' or $db['default']['password'] == '', CodeIgniter fails to connect. This commit fixes it.

### DIFF
--- a/system/database/drivers/pdo/pdo_driver.php
+++ b/system/database/drivers/pdo/pdo_driver.php
@@ -130,7 +130,14 @@ class CI_DB_pdo_driver extends CI_DB {
 
 		try
 		{
-			return new PDO($this->dsn, $this->username, $this->password, $this->options);
+			if (empty($this->username) OR empty($this->password))
+			{
+				return new PDO($this->dsn, NULL, NULL, $this->options);
+			}
+			else
+			{
+				return new PDO($this->dsn, $this->username, $this->password, $this->options);
+			}
 		}
 		catch (PDOException $e)
 		{


narfbg:
I don't understand why would you want not to use these, but if you really do, you can obviously set them to NULLs. So where's the gain here?

ricpelo:
I really do want to have all the connection params into the DSN, which is one only conf param, much better than three (DSN, user & pass). Setting user & password to NULL is a way to workaround the issue, but setting them to '' (an empty string) is obviously another way to say: "no user & password here; see in the DSN". IMHO, an empty string says the same as NULL. On the other hand, documentation should be more explicit about using NULLs if user & password are into the DSN. Could save some hours of trying to figure why CodeIgniter don't connect to my DB.

mwhitneysdsu:
`if (empty($this->username) OR empty($this->password))` would also be true if either `$this->username` or `$this->password` is `'0'` (a string containing 0).

The fact that PDO treats an empty value in these parameters as different from NULL is probably enough reason to justify not treating them the same in CI.

As a programmer, it's important to get into the mindset of differentiating between NULL and an empty value, especially in loosely-typed languages with type juggling and constructs like `empty()`. Code is usually easier to maintain when you don't have to consult the [type comparison tables](http://php.net/manual/en/types.comparisons.php) to figure out why the code behaves the way it does.

narfbg:
What @mwhitneysdsu said ... Plus, what you want to do is hardly common and I don't really see a benefit (especially given caveats such as `empty('0')` calls).

Also, empty string is the default value and it's far more likely that somebody just forgot to fill-in the credentials if it stays as is. Setting to NULL is explicit, and explicit is good. :)

PR #4323, opened 2015-12-21
Refactored the `CI_Log` line formatting

jspreddy:
- Extracted the `CI_Log` line formatting into a protected
  function `_format_log_line()` so as to make it easy to
  change the log line format via extending the class and
  overriding the method.

### DIFF
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -204,7 +204,7 @@ class CI_Log {
 			$date = date($this->_date_fmt);
 		}
 
-		$message .= $level.' - '.$date.' --> '.$msg."\n";
+		$message .= $this->_format_log_line($level, $date, $msg);
 
 		flock($fp, LOCK_EX);
 
@@ -227,4 +227,20 @@ class CI_Log {
 		return is_int($result);
 	}
 
+	/**
+	 * Format the log line.
+	 *
+	 * This is for extensibility of log formatting.
+	 * If you want to change the log format,
+	 * extend the CI_Log class and override this method.
+	 *
+	 * @param $level
+	 * @param $date
+	 * @param $msg
+	 *
+	 * @return string formatted log line with a new line character '\n' at the end.
+	 */
+	protected function _format_log_line($level, $date, $msg){
+		return $level.' - '.$date.' --> '.$msg."\n";
+	}
 }
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -232,14 +232,13 @@ class CI_Log {
 	/**
 	 * Format the log line.
 	 *
-	 * This is for extensibility of log formatting.
-	 * If you want to change the log format,
-	 * extend the CI_Log class and override this method.
+	 * This is for extensibility of log formatting
+	 * If you want to change the log format, extend the CI_Log class and override this method
 	 *
 	 * @param	string	$level 	The error level
 	 * @param	string	$date 	Formatted date string
 	 * @param	string	$msg 	The log message
-	 * @return	string	formatted log line with a new line character '\n' at the end.
+	 * @return	string	Formatted log line with a new line character '\n' at the end
 	 */
 	protected function _format_line($level, $date, $message)
 	{
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -154,8 +154,8 @@ class CI_Log {
 	 *
 	 * Generally this function will be called using the global log_message() function
 	 *
-	 * @param	string	the error level: 'error', 'debug' or 'info'
-	 * @param	string	the error message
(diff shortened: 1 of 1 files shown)


narfbg:
Your code doesn't comply with our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html) (you could've just taken the current code as an example) and if this is for extensibility - why a protected method instead of a property?

jspreddy:
Thanks for the catch. I'll fix that.

I not only need to change the order of level, date and message but also change the separators and add more items to the log line.
I also want to add filler characters to the log level ("DEBUG","xINFO") to make the logs cleaner and appear in order.

How do I do this via a property?

narfbg:
I suspected you'd say you want to change the order of these tokens, but ... why?

jspreddy:
In our company we have a ton of different applications built in many different ways. All of which write logs in a specific format. I am working on a couple of new services and want to write logs from codeigniter in the same format we've been using in all other places.

jim-parry:
You could extend the CI_Log class, using application/core/MY_Log, with your custom formatting there.

jspreddy:
Correct. That is exactly what I want to be able to do.

But I don't see a way to extend `CI_Log` without having to rewrite the whole of `write_log()` function. Since that function is the meat of `CI_Log`.

This patch will make it easier to make my own log line format.
It will extract the code that formats the log line into its own function inside `CI_Log`.
log_php_-_codeigniter_-____dev_github_codeigniter_

So that I can extend it in my `MY_Log` class like so.
my_log_php_-_svc-fileserv-core_-____dev_passport_svc-fileserv-core_

jspreddy:
@jim-parry @narfbg any updates on this?

narfbg:
@jspreddy Please be patient, it's holiday season. :)

On topic: I don't really like it, but what the hell ... it's not like `CI_Log` does much else anyway.

I'll leave some inline notes however ... the patch is still inconsistent with our other code.

system/core/Log.php
```diff
@@ -227,4 +227,20 @@ public function write_log($level, $msg)
 		return is_int($result);
 	}
 
+	/**
```
narfbg: We put the following between separate methods:

system/core/Log.php
```diff
@@ -227,4 +227,20 @@ public function write_log($level, $msg)
...
+	 * Format the log line.
+	 *
+	 * This is for extensibility of log formatting.
+	 * If you want to change the log format,
+	 * extend the CI_Log class and override this method.
+	 *
+	 * @param	string	the error level: 'error', 'debug' or 'info'
```
narfbg: I now see that you've actually taken the `write_log()` docblock as an example for these ... But it seems that *we* didn't correct that one in particular - sorry about that.

There's supposed to be a variable name between the type and description (see [this one](https://github.com/bcit-ci/CodeIgniter/blob/5afa348b48a93f24957377dc12f86ae64665b944/system/libraries/Encryption.php#L352) for example) and the description should start with a capital letter.

Also, the code in this method obviously doesn't care about 'error', 'debug', 'info' or something else - it will blindly put anything passed to it, so it should become just "The error level", no details.

system/core/Log.php
```diff
@@ -227,4 +227,20 @@ public function write_log($level, $msg)
...
+	 * @param	string	formatted date string
+	 * @param	string	the log message
+	 * @return	string	formatted log line with a new line character '\n' at the end.
+	 */
+	protected function _format_log_line($level, $date, $msg)
```
narfbg: The 'log' part of the method name is superfluous ... we know it's about logging, it can be just '_format_line'.
Also, `$message` would be better than `$msg`.

Those are stylistic "mistakes" that we've done ourselves in the past, but we try to avoid them now, so ... :)

jspreddy:
roger, will-fix.

jspreddy:
@narfbg Anything else I need to change?

system/core/Log.php
```diff
@@ -227,4 +227,22 @@ public function write_log($level, $msg)
...
+	 * Format the log line.
+	 *
+	 * This is for extensibility of log formatting.
+	 * If you want to change the log format,
+	 * extend the CI_Log class and override this method.
+	 *
+	 * @param	string	$level 	The error level
+	 * @param	string	$date 	Formatted date string
+	 * @param	string	$msg 	The log message
+	 * @return	string	formatted log line with a new line character '\n' at the end.
```
narfbg: Just this here:

- Sentence should start with a capital "F"
- Use double quotes around \n (it's not a newline character if it's between single quotes)

While at it, you may also remove the dot at the end ... for consistency with other lines. :)

PR #4327, opened 2015-12-23
Removed deprecated function valid_email

pmbaldha:
Use PHP's filter_var() instead


pmbaldha:
I am pulling request after removing valid email function from test case

PR #4328, opened 2015-12-23
Removed deprecated function valid_email

pmbaldha:
I think that valid_email function is not widely used and it should be removed from latest version. Developer should use filter_var($email, FILTER_VALIDATE_EMAIL) instead of valid_email. I am puling request with removing from test case also.


narfbg:
Sorry, but we don't accept PRs to remove functionalities based on "I don't think it is widely used" (and in fact, I believe it is widely used).

It's going to be removed anyway, but not just this function - the whole email helper collection is deprecated - and pending a decision whether that will happen in 3.1 or later.

Issue #4329, opened 2015-12-23
update_string (and possibly others) clobber SQL with backticks in 3.0.1 (worked fine in 2.x)


QuakePhil:
Consider this snippet:

    $sql = $this->db->update_string('test', array('column' => 'value'), 'id = ?');

It will generate

    $sql = "UPDATE `test` SET `column` = 'value' WHERE `id` = `?`"

which then becomes something like

    "UPDATE `test` SET `column` = 'value' WHERE `id` = `10`"

Mysql will then error out with "Unknown column 10" because the value in the where clause is backticked in error.

QuakePhil:
I have fixed this here https://github.com/QuakePhil/CodeIgniter/commit/273a818ac98568404339ebb228d03e14e149e084

and issued a PR

narfbg:
I don't understand how that question mark becomes a 10?

PR #4330, opened 2015-12-23
Fix issue 4329

QuakePhil:


narfbg:
Properly referencing #4329.

PR #4333, opened 2015-12-25
Add support for subresource integrity

vfalconi:
The idea behind this PR is to implement subresource integrity checks on linked resources (scripts, CSS, etc), as highlighted by [GitHub Engineering](http://githubengineering.com/subresource-integrity/).

The primary use case for this is when an application or website loads a script or a stylesheet hosted by a CDN or some other third-party. The `integrity` attribute contains a hash of a local copy of the linked file, which is used to verify that the linked file is the version that the author expects to have loaded, the idea being as to prevent malicious code from being injected.

The integrity check also works on locally-stored files, though the risk of those files being modified in transit is probably lower.

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -849,3 +849,26 @@ if ( ! function_exists('function_usable'))
 		return FALSE;
 	}
 }
+
+// ------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


vfalconi:
Just realized my commits weren't signed. Closing and resubmitting.

PR #4334, opened 2015-12-26
Fixes an issue with uri_string detection.

BrunoDeBarros:
When `$_SERVER['REQUEST_URI']` doesn't have a leading slash, it leads to the dummy URL being generated as `http://dummytest`, instead of `http://dummy/test` as expected, which breaks uri_string detection. This fix makes sure the leading slash is there at all times.

### DIFF
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -203,7 +203,7 @@ class CI_URI {
 
 		// parse_url() returns false if no host is present, but the path or query string
 		// contains a colon followed by a number
-		$uri = parse_url('http://dummy/'.ltrim($_SERVER['REQUEST_URI'], '/'))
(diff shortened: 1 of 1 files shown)


narfbg:
I'm not aware of a case where a `$_SERVER['REQUEST_URI']` wouldn't have a leading slash and not be empty. Can you share more details?

BrunoDeBarros:
Feel free to reject the PR; it was due to some code that was modifying the REQUEST_URI prior to CI loading in a project I inherited.

ivantcholakov:
$_SERVER['REQUEST_URI'] was empty on IIS6, but this is in ancient times. I don't know other similar cases.

PR #4338, opened 2015-12-28
fix pdo odbc cant't use quote function bug

capturePointer:


narfbg:
Your patch seems to be based on the assumption that ODBC is only used with MSSQL ... it's not. 

ODBC is an abstraction layer that can connect to literally every RDBMS and there's no way to implement such functionalities in a platform-abstract way, which is why it lacks features.

PR #4342, opened 2015-12-28
Refactor: logic to get the rule's error message

jspreddy:
Refactored the logic for getting an error message related to a rule.
Don't want to have the same logic in two places.

priority wise:
1. check if a custom message is defined through validation config row.
2. check if a custom message has been set using the set_message() function
3. check if we have an error message in lang file
4. DEPRECATED support for non-prefixed keys, in lang file again
5. error message not found -> default to `"form_validation_error_message_not_set"`

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -637,7 +637,7 @@ class CI_Form_validation {
 				// Set the message type
 				$type = in_array('required', $rules) ? 'required' : 'isset';
 
-				$line = $this->_get_error_message($type, $row["field"]);
+				$line = $this->_get_error_message($type, $row['field']);
 
 				// Build the error message
 				$message = $this->_build_error_msg($line, $this->_translate_fieldname($row['label']));
@@ -808,7 +808,7 @@ class CI_Form_validation {
 				}
 				else
 				{
-					$line = $this->_get_error_message($rule, $row["field"]);
+					$line = $this->_get_error_message($rule, $row['field']);
 				}
 
 				// Is the parameter we are inserting into the error message the name
@@ -840,8 +840,7 @@ class CI_Form_validation {
 	 * Get the error message for the rule
 	 *
 	 * @param 	string $rule 	The rule name
-	 * @param 	string $field
-	 *
+	 * @param 	string $field	The field name
 	 * @return 	string
 	 */
 	protected function _get_error_message($rule, $field)
@@ -856,7 +855,7 @@ class CI_Form_validation {
 		{
 			return $this->_error_messages[$rule];
 		}
-		elseif (FALSE !== ($tmp = $this->CI->lang->line('form_validation_' . $rule)))
+		elseif (FALSE !== ($tmp = $this->CI->lang->line('form_validation_'.$rule)))
 		{
 			return $tmp;
 		}
@@ -866,7 +865,7 @@ class CI_Form_validation {
 			return $tmp;
 		}
 
-		return $this->CI->lang->line('form_validation_error_message_not_set'). '(' . $rule . ')';
+		return $this->CI->lang->line('form_validation_error_message_not_set').'('.$rule.')';
 	}
 
 	// --------------------------------------------------------------------
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -637,7 +637,7 @@ class CI_Form_validation {
 				// Set the message type
 				$type = in_array('required', $rules) ? 'required' : 'isset';
 
-				$line = $this->_get_raw_error_message($type, $row);
+				$line = $this->_get_error_message($type, $row["field"]);
 
 				// Build the error message
 				$message = $this->_build_error_msg($line, $this->_translate_fieldname($row['label']));
@@ -808,7 +808,7 @@ class CI_Form_validation {
 				}
 				else
 				{
-					$line = $this->_get_raw_error_message($rule, $row);
+					$line = $this->_get_error_message($rule, $row["field"]);
 				}
 
 				// Is the parameter we are inserting into the error message the name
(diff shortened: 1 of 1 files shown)


narfbg:
It's not the same logic.

jspreddy:
@narfbg I understand that you are busy but can you help me with this? 
The only difference I could find is the difference in the default message used if error message line was not set anywhere in the system.

Here we are setting the line to a custom string message.
refactor__logic_to_get_the_rule_s_error_message_by_jspreddy_ _pull_request__4342_ _bcit-ci_codeigniter

Here we are using a similar message from the lang file.
refactor__logic_to_get_the_rule_s_error_message_by_jspreddy_ _pull_request__4342_ _bcit-ci_codeigniter



I believe the rest of the logic is the same as the code that I refactored from.

## The new extracted function
refactor__logic_to_get_the_rule_s_error_message_by_jspreddy_ _pull_request__4342_ _bcit-ci_codeigniter

## Block1 (823 - 840)
refactor__logic_to_get_the_rule_s_error_message_by_jspreddy_ _pull_request__4342_ _bcit-ci_codeigniter

If we remove the `!` and interchange the code inside else and elseif blocks, we get _Block2_
refactor__logic_to_get_the_rule_s_error_message_by_jspreddy_ _pull_request__4342_ _bcit-ci_codeigniter


## Block2 (640 - 655)
refactor__logic_to_get_the_rule_s_error_message_by_jspreddy_ _pull_request__4342_ _bcit-ci_codeigniter

narfbg:
My bad, I only looked at the removed code and it's not obviously the same logic, so I closed it. Reopening and I'll leave inline comments for the stuff that you need to update.

However, unless you absolutely have to, please don't use screenshots in the future ... the steps numbering and arrows did help, but if I had to copy any of that code for debugging purposes, I wouldn't be able to.

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,47 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
 	// --------------------------------------------------------------------
 
 	/**
+	 * Get the error message for the rule
+	 *
+	 * @param 	string the rule name.
```
narfbg: The format is:

    @param	type	$variable_name	description

... and descriptions start with capital letters, no period at EOL

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,47 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+	 * @param 	array
+	 * @return 	string
+	 */
+	private function _get_raw_error_message($rule_name, $row)
```
narfbg: We don't use private methods; use protected, so that it is extendable.

Also:

 - it should be named just `_get_error_message()`
 - `$rule_name` should be just `$rule` (who says it's a name? it may be a callable)
 - `$row` doesn't make sense in this context; it should be `$field`

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,47 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+	{
+		$error_message = '';
```
narfbg: This is unnecessary, you set the variable name at all times anyway.

On another note, instead of setting it only to return it after the if block, you could just return straight from inside the conditions (and you wouldn't need the `else` in that case).

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,47 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+		if (isset($this->_field_data[$row['field']]['errors'][$rule_name]))
+		{
+			$error_message = $this->_field_data[$row['field']]['errors'][$rule_name];
+		}
+		// check if a custom message has been set using the set_message() function
+		elseif (isset($this->_error_messages[$rule_name]))
+		{
+			$error_message = $this->_error_messages[$rule_name];
+		}
+		// check if we have an error message in lang file
```
narfbg: This comment line is completely redundant, the code is really obvious.

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,47 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+		elseif (FALSE !== ($line_tmp = $this->CI->lang->line('form_validation_'.$rule_name)))
```
narfbg: ... it's also obvious that it's a temporary variable, why did you remove it to `$line_tmp`?

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,47 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+		{
+			$error_message = $line_tmp;
+		}
+		// DEPRECATED support for non-prefixed keys, lang file again
+		elseif (FALSE !== ($line_tmp = $this->CI->lang->line($rule_name, FALSE)))
+		{
+			$error_message = $line_tmp;
+		}
+		// error message not found
```
narfbg: Another redundant comment.

narfbg:
On another note ... 3.1 won't support PHP 5.2, so a closure (instead of a hard-declared method) would be just fine here and probably more performant.

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,41 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+		if (isset($this->_field_data[$field]['errors'][$rule]))
+		{
+			return $this->_field_data[$field]['errors'][$rule];
+		}
+		// check if a custom message has been set using the set_message() function
+		elseif (isset($this->_error_messages[$rule]))
+		{
+			return $this->_error_messages[$rule];
+		}
+		elseif (FALSE !== ($tmp = $this->CI->lang->line('form_validation_' . $rule)))
```
narfbg: https://github.com/jspreddy/CodeIgniter/commit/dd149da93a2064937e490977343d152207b73815#commitcomment-15986814 was actually meant for here ... sorry.

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,41 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
 	// --------------------------------------------------------------------
 
 	/**
+	 * Get the error message for the rule
+	 *
+	 * @param 	string $rule 	The rule name
+	 * @param 	string $field
```
narfbg: Missing description here.

system/libraries/Form_validation.php
```diff
@@ -865,6 +837,41 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
+	 *
```
narfbg: ... and we don't separate phpdoc stuff with empty lines. :)

system/libraries/Form_validation.php
```diff
@@ -637,21 +637,7 @@ protected function _execute($row, $rules, $postdata = NULL, $cycles = 0)
...
-				{
-					$line = $this->_error_messages[$type];
-				}
-				elseif (FALSE === ($line = $this->CI->lang->line('form_validation_'.$type))
-					// DEPRECATED support for non-prefixed keys
-					&& FALSE === ($line = $this->CI->lang->line($type, FALSE)))
-				{
-					$line = 'The field was not set';
-				}
+				$line = $this->_get_error_message($type, $row["field"]);
```
narfbg: This is irrelevant, but we generally use single quotes unless doubles are necessary.

I don't really mind, but since you have to make other adjustments, you may as well change this line and 811, for consistency.

narfbg:
I notice that you didn't leave me a comment after your latest changes. Please do that next time as commits don't issue notifications ...

... unless you write `@somebody` in the commit message, which you did, but shouldn't. Now each time somebody merges that commit in another repo I'll get spammed.

system/libraries/Form_validation.php
jspreddy: Could the description be "The field name"

system/libraries/Form_validation.php
jspreddy: good call. Will keep that in mind.

jspreddy:
@narfbg Made the changes.

PR #4346, opened 2015-12-29
Validate date strings in all formats

shanavas786:
Checks whether the string is a valid date. The default format is 'Y-m-d'. 

Format should be compatible with php's [DateTime::createFromFormat](http://php.net/manual/en/datetime.createfromformat.php)

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -1132,6 +1132,26 @@ class CI_Form_validation {
 	// --------------------------------------------------------------------
 
 	/**
+	 * Exists
+	 *
+	 * Check if the input value already exists
+	 * in the specified database field
+	 *
+	 * @param	string	$str
+	 * @param	string	$field
+	 * @return	bool
+	 */
+	public function exists($str, $field)
+	{
+		sscanf($field, '%[^.].%[^.]', $table, $field);
+		return isset($this->CI->db)
+			? ($this->CI->db->limit(1)->get_where($table, array($field => $str))->num_rows() === 1)
+				: FALSE;
+	}
+
+	// --------------------------------------------------------------------
+
+	/**
 	 * Minimum Length
 	 *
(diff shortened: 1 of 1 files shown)


narfbg:
We may discuss `valid_date()` (although it carries the same negatives as time, currency and other locale-specific values, so it will be hard to convince me), but we've rejected `exists()` numerous times before already and I'm not accepting it now.

Which is why you shouldn't submit multiple features at the same time. :)

shanavas786:
What do you mean my negatives ?

narfbg:
As I said, it's locale-specific ... it's not universal. When was the last time you saw a "Date" field in a form and you *knew*, without additional notes, exactly what format to use when typing it? And it gets worse if the preferred format uses worded month names instead of numbers.

shanavas786:
Now a days, most of the websites employ date pickers for date inputs. So the user has little to do with the format. Even without date pickers there will be at least a placeholder which describes the format.

narfbg:
That's exactly what I'm saying - you have to add more stuff for date fields to work sanely. In fact, chances are you'd want to do even more checks, like is the date within a certain period.

So what problem does this solve?

shanavas786:
Of course yes. There could be need for other validations as well. We may include those in future :)

But, anyway isn't it a primary need to check whether it is a valid date string in the expected format ?

narfbg:
It's "most likely", not just "could be"; and you can't just add them later, because there are endless use cases.

That's an issue because if it only solves a fraction of the problem and it obviously only requires 1-2 lines of code if you do it manually, then you're hardly adding any value with it - more of a maintenance burden than a practical solution.

The other problem is that you're defaulting to 'Y-m-d' ... Because dates are locale-specific, I'd argue that more than half of the people using it would have to use the second parameter, making the default quite meaningless - defaults are supposed to work for most people.

PR #4347, opened 2015-12-29
Check if input value already exists in the specified database field

shanavas786:
Signed-off-by: Shanavas M <shanavas.m2@gmail.com>
Just reverse of is_unique method


narfbg:
No ... there are many other ways to achieve this, including "in_list".

PR #4349, opened 2015-12-30
allow custom mysqli option before invoking $mysqli->real_connect

nickshek:
In some situations, we may want to disable mysql strict mode without changing the sql_mode value globally.If we add the "options" key in the DATABASE CONNECTIVITY SETTINGS, we may able to inject some custom mysqli option before $mysqli->real_connect.For example:

	$db['default'] = array(
		'dsn'	=> '',
		'hostname' => 'localhost',
		'username' => '',
		'password' => '',
		'database' => '',
		'dbdriver' => 'mysqli',
		'dbprefix' => '',
		'pconnect' => FALSE,
		'db_debug' => (ENVIRONMENT !== 'production'),
		'cache_on' => FALSE,
		'cachedir' => '',
		'char_set' => 'utf8',
		'dbcollat' => 'utf8_general_ci',
		'swap_pre' => '',
		'encrypt' => FALSE,
		'compress' => FALSE,
		'stricton' => FALSE,
		'failover' => array(),
		'save_queries' => TRUE,
		'options' => array(
			MYSQLI_INIT_COMMAND => "SET SESSION sql_mode = 'NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION'"
		)
	);

### DIFF
--- a/system/database/drivers/mysqli/mysqli_driver.php
+++ b/system/database/drivers/mysqli/mysqli_driver.php
@@ -158,6 +158,12 @@ class CI_DB_mysqli_driver extends CI_DB {
 			}
 		}
 
+		if(isset($this->options) && is_array($this->options) && $this->options){
+			foreach($this->options as $mysqli_option => $mysqli_value){
+				$mysqli->options($mysqli_option, $mysqli_value);
+			}
+		}
+
 		if ($mysqli->real_connect($hostname, $this->username, $this->password, $this->database, $port, $socket, $client_flags))
 		{
 			// Prior to version 5.7.3, MySQL silently downgrades to an unencrypted connection if SSL setup fails


narfbg:
Do you want to preserve other mode flags (which seems to be your motivation) or do you really want the 'options' functionality?

I can imagine that the former could be achieved in another way, while the latter has caveats (like your patch being mysqli-specific).

nickshek:
Thank you your reply.Actually, if we are using MySQL as our database, we may just want to prevent the sql errors without the modification of CodeIgniter when we upgrade the MySQL to 5.7 (strict mode by default).

narfbg:
I don't quite understand what you mean by that.

narfbg:
... and I can't find references to 5.7 setting strict mode by default.

TrentForshaw:
> SQL mode changes. Strict SQL mode for transactional storage engines (STRICT_TRANS_TABLES) is now enabled by default.

> Implementation for the ONLY_FULL_GROUP_BY SQL mode has been made more sophisticated, to no longer reject deterministic queries that previously were rejected. In consequence, this mode is now enabled by default, to prohibit only nondeterministic queries containing expressions not guaranteed to be uniquely determined within a group.

> The ERROR_FOR_DIVISION_BY_ZERO, NO_ZERO_DATE, and NO_ZERO_IN_DATE SQL modes are now deprecated but enabled by default. The long term plan is to have them included in strict SQL mode and to remove them as explicit modes in a future MySQL release. See SQL Mode Changes in MySQL 5.7.

> The changes to the default SQL mode result in a default sql_mode system variable value with these modes enabled: ONLY_FULL_GROUP_BY, STRICT_TRANS_TABLES, NO_ZERO_IN_DATE, NO_ZERO_DATE, ERROR_FOR_DIVISION_BY_ZERO, NO_AUTO_CREATE_USER, and NO_ENGINE_SUBSTITUTION.



https://dev.mysql.com/doc/refman/5.7/en/mysql-nutshell.html

narfbg:
I see ... I was only looking for `STRICT_ALL_TABLES`.

narfbg:
b98c657c100d3220a5d7ed1b6ff3ec78e227b406
c83e894fe8b3c85ff40f00954df0033ad14940b0
679e94853451192410ac138cb592ac5ab021ea67

nickshek:
Thank you!

PR #4352, opened 2016-01-01
Allow route callback to return FALSE

yzou:
If route callback returns FALSE, then try next route in routes. This is useful if we have to map the top level path to different controller.


narfbg:
More details please ...

PR #4363, opened 2016-01-08
update attributes in the form helper

Rodrigo54:
if, in one of the key attributes of the array is null form_help returns only the key

example:

```php
$attributes = array(
      'type'     => 'text',
      'disabled' => '',
      'name'     => 'text',
      'id'       => 'text',
      'class'    => 'text',
      'value'    => 'text',
      );
echo form_input($attributes);
```
return 
```html
<input type="text" name="text" value="text" disabled id="text" class="text" />
```


narfbg:
... and that's perfectly valid.

Issue #4003, opened 2015-07-27
Session_database_driver.php detecting wrong IP address if using proxy server


Jammyman:
I think there might be a bug with Database session driver since it doesn't use the config set in $config['proxy_ips']; and stores the proxy IP in the database.

After some digging, seems to me that right way to get correct ip is to use $input->ip_address() so 
how about new variable in Session_database_driver.php:
```php
/**
 * Ip address
 * @var
*/
protected $ip_address;
```

And in constructor
```php
$this->ip_address =  $CI->input->ip_address();
```

Then just replace all
```php
$_SERVER['REMOTE_ADDR'] 
```
with
```php
$this->ip_address 
```

That seems to work with my setup.

narfbg:
Not a bug ... I don't want the session library to depend on the input class *and* I wouldn't trust anybody's choice about 'proxy_ips' to be used in a security context.

In fact, if the 'proxy_ips' setting was proposed today, I'd say those who need it and know what they're doing, would work around the issue with just a single line in index.php.

PR #4367, opened 2016-01-08
change it to use HTTP_X_REAL_IP when I do proxy

bootjp:


bootjp:
Was "array_key_exists" better?

narfbg:
Not happening.

Relevant: #4003

bootjp:
oh... thanks!

PR #4368, opened 2016-01-08
change year in license file

ceaniket:


jim-parry:
Thank you for thinking of this, but we don't need a PR for it :)
Narf usually does a bulk replace of the copyright notices throughout the framework (not just the files in the project root) and updates the repository, without having to go through a PR.

Issue #4277, opened 2015-11-28
is_supported with memcached error


Kaabi:
using is_supported with memcached driver is getting fatal error since CI 3.0.1 or above
currently testing on CI 3.0.3 is only return fatal error due to new addServer function

Code
```
if (!$this->cache->memcached->is_supported())
        {
            echo "Not Supported";
        }
```

Result

```
Fatal error: Call to a member function addServer() on a non-object in /var/www/box/system/libraries/Cache/drivers/Cache_memcached.php on line 130
A PHP Error was encountered

Severity: Error

Message: Call to a member function addServer() on a non-object

Filename: drivers/Cache_memcached.php

Line Number: 130
```

Filisko:
$this->load->driver('cache');
if ($this->cache->memcached->is_supported())
{
    echo "Ok";
}

Does that work to you?

Kaabi:
@narfbg if you punishing @bjjay why you close this issue while you didn't merge #4376, do I have to create new Issue ?

narfbg:
How is this issue closed?

Kaabi:
@narfbg sorry mybad I get confused between this and other issue though both closed

cmeenatchisundaram:
@narfbg
Do we require to load the driver like below before using the is_supported() function...
$this->load->driver('cache', array('adapter' => 'memcached'));

narfbg:
No.

cmeenatchisundaram:
@narfbg 
   I looked on the properties of $this obj. But there is no cache object on it when i am trying to use in controller. So i felt that, need to load the cache object. 

I too got the error when i am trying to use it directly like below : 

if(!$this->cache->memcached->is_supported()){
}

Could you tell me, What to do if we don't require to load the cache obj?

narfbg:
What do you mean what to do? It's a bug report, obviously something doesn't work right.

cmeenatchisundaram:
This means, Is the $this obj suppose to have the cache obj? I am trying to fix this issue.

narfbg:
There's a pull request referenced above that contained a fix, object instances have nothing to do with it.

narfbg:
#4066 is relevant.

PR #4376, opened 2016-01-12
Fixed PR #4277

bjjay:


narfbg:
- "PR" stands for "Pull Request"; #4277 is not a pull request.
- You're missing a changelog entry.
- You're missing a basic description here, similarly to half of your pull requests in the past.
- You've opened a separate PR in #4377 for what is essentially the same thing as here (and also lacks a changelog entry and doesn't follow our styleguide).

This may be a dick move, but I'm closing both of these - make an effort please.

PR #4382, opened 2016-01-16
Name correction

TheDigitalOrchard:
and removal of unnecessary blank line


application/config/user_agents.php
```diff
@@ -34,8 +34,8 @@
 	'iphone'			=> 'iOS',
 	'ipad'				=> 'iOS',
 	'ipod'				=> 'iOS',
-	'os x'				=> 'Mac OS X',
```
narfbg: This change breaks a unit test.

PR #4390, opened 2016-01-20
HTML 5 input tags

sebastienadam:
Creating helpers to generate HTML 5 input tags (just as "password"). Documentation and tests adapted.


narfbg:
Sorry, but no ... adding 10+ identical functions just bloats our codebase, especially given how useless these particular helpers are (the existing ones).

PR #4396, opened 2016-01-25
Some small fixes

kenjis:
It seems the style like `_application/config/mimes.php*` does not work.
See http://www.codeigniter.com/user_guide/installation/upgrade_300.html#step-3-replace-config-mimes-php

### DIFF
--- a/user_guide_src/source/installation/upgrade_300.rst
+++ b/user_guide_src/source/installation/upgrade_300.rst
@@ -222,7 +222,7 @@ Routes containing :any
 ======================
 
 Historically, CodeIgniter has always provided the **:any** wildcard in
-routing, with the intention of providing a way to match any characters
+routing, with the intention of providing a way to match any character
 **within** an URI segment.
 
 However, the **:any** wildcard is actually just an alias for a regular
@@ -865,7 +865,7 @@ tag ``<mark>``.
 Unless you've used your own highlighting tags, this might cause trouble
 for your visitors who use older web browsers such as Internet Explorer 8.
 We therefore suggest that you add the following code to your CSS files
-for backward compatibility with old browsers::
+for backwards compatibility with old browsers::
 
 	mark {
 		background: #ff0;
--- a/user_guide_src/source/installation/upgrade_300.rst
+++ b/user_guide_src/source/installation/upgrade_300.rst
@@ -65,7 +65,7 @@ Step 3: Replace config/mimes.php
 ********************************
 
 This config file has been updated to contain more user mime-types, please copy
-it to **application/config/mimes.php**.
+it to *application/config/mimes.php*.
 
 **************************************************************
 Step 4: Remove $autoload['core'] from your config/autoload.php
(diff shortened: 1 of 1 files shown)


narfbg:
It's supposed to be italic, not bold ... Please change that to a single asterisk and revert the rest of the changes.

kenjis:
Oh, sorry. I changed to a single asterisk.

And I found that a line break is needed between two file paths:
http://www.codeigniter.com/user_guide/installation/upgrade_300.html#step-5-move-your-log-class-overrides-or-extensions

>  revert the rest of the changes.

Where is wrong?

narfbg:
What's wrong is that you're fixing things that aren't broken ... This is far from your first PR, you should know, at the very least, not to batch multiple changes in a single patch. Here's why:

- While an acceptable change, the lack of a full stop at the end of one sentence isn't necessarily wrong.
- Changing "in order to avoid backwards compatibility" to "for backward compatibility" is good, but I insist on the "s". :)
- Changing "any character" to "any characters" is wrong - "any" for singular, "all" for plural.

If you've found that a linebreak is missing, why are you converting that thing to a list instead of adding the linebreak? While it is obviously a list of a kind, we use different indentation level for bullet-point lists (which this was not intended to be in the first place) and there are at least two similar instances that are now inconsistent with this one (but admittedly also broken, due to missing linebreaks).

It's not that big of a deal - I'm sure you'll now fix these issues. However, your description only says that you're addressing the ``_foo/bar*`` thing and if you really only did that, it would've been merged by now and we wouldn't have to now go back and forth in a conversation that neither of us enjoys.

kenjis:
Okay, I posted another PR #4411 only containing changes formatting of paths.

I close this PR, but I still have something to fix.

> If you've found that a linebreak is missing, why are you converting that thing to a list instead of adding the linebreak?

Because I don't know how to do so. I searched it but I was not able to find how to put linebreak in RST.

> we use different indentation level for bullet-point lists (which this was not intended to be in the first place) and there are at least two similar instances that are now inconsistent with this one

I don't understand the format you want. Do you mean to add one more tab?

narfbg:
>> If you've found that a linebreak is missing, why are you converting that thing to a list instead of adding the linebreak?
>
> Because I don't know how to do so. I searched it but I was not able to find how to put linebreak in RST.

I just found this: http://docutils.sourceforge.net/FAQ.html#how-to-indicate-a-line-break-or-a-significant-newline

But as a general rule, if you don't know how to fix something - just note that it is broken and leave it for someone else to fix. Treat it as real programming code - completely changing it is likely not welcome.

In this particular case, I don't mind switching it to a bullet-point list, but make it consistent and do that for other lists in the document as well. :) Although, changing it to a literal code block may be more readable.

>> we use different indentation level for bullet-point lists (which this was not intended to be in the first place) and there are at least two similar instances that are now inconsistent with this one
>
> I don't understand the format you want. Do you mean to add one more tab?

No ... See "Step 6", which uses a bullet-point list already - the dashes are indented by just 2 spaces further.

Issue #4362, opened 2016-01-06
Codeigniter session error in Redis using PHP 7.0


degp:
When I store session data in the Database, there are no errors. But if I switch to Redis, I get an error every time session tries to regenerate via session_regenerate_id, in my case every 300 sec. But if I refresh the page again, Redis get updated and page loads without errors.

    A PHP Error was encountered
    Severity: 4096
    Message: session_regenerate_id(): Failed to create(read) session ID: user (path: tcp://localhost:6379?auth=xxxx)
    Filename: Session/Session.php
    Line Number: 625

narfbg:
Related to #4179.

@degp Can you confirm if the commit referenced above fixes the issue?

degp:
@narfbg yes! it is working.

skys215:
I'm facing same problem under PHP7.1.4, CI3.1.6-dev on macOS Sierra.

victory7:
I have the same problem and also error "Unable to obtain lock for [sessionKey] after 30 attempts, aborting.
I'm on CI 3.1.9 (latest now), PHP 7.0.0 , php-redis 4.2.0 , redis 3.2.12 .
After acquiring this error the request waits for 30 seconds and after that my session is destroyed and the user is logged out.
This happening on different URLs.

kaamoss:
I have also run into this exact same issue on PHP 7.2 and CI 3.1.9 , this is only an issue for me on concurrent ajax requests.

ostrolucky:
Same here on PHP 7.0 and CI 3.1.10, also with concurrent ajax requests

schmoove:
Has anyone found a resolution for the `Session: Unable to obtain lock for ci_session:xxx` issue?

We are also seeing this on PHP 7.4.20 and CI 3.1.11, and it is always related to concurrent ajax requests.

victory7:
The best solution for me was to close the session (session_write_close) at the beginning of the method or wherever updating the session ended. It releases session lock for other concurrent requests.

Issue #3919, opened 2015-06-18
memcached driver for session and cache does not work


ghost:
Hi,

I am trying to use memcached as session and cache storage and running into troubles doing so.

For session setting userdata works for the 2 first requests, after that there are huge delays, up 20 sec, and I'm not sure where this comes from. I tested the same server setup without using codeigniter and it works fine:

<pre>
ini_set('session.save_handler', 'memcached');
ini_set('session.save_path', 'tcp://10.126.1.17:11211');

session_start();

$_SESSION['foo'] = 'bar';
$_SESSION['time'] = time();

var_dump($_SESSION);
</pre>

For cache there are no delays but the data is never stored. In the code below both var_dumps output bool(false):

<pre>
public function test() {
$this->load->driver('cache', array(
	'adapter' => 'memcached', 
	'backup' => 'file'
));

var_dump($this->cache->get('foo'));
$this->cache->save('foo', 'test test', 300);
var_dump($this->cache->get('foo'));
}
</pre>

shankarbala33:
Should know the memory space for storing session datas. Core PHP provides
more memory than the Codeigniter. So you couldn't store more datas and
retrieve itself quickly.
That's why you get the different time delays !

narfbg:
The session delays are caused by emulated locks that serve for avoiding concurrency issues ... don't worry about that - the delays will disappear once the writes start working again.  
That being said however, I'm afraid you'd have to debug this one on your own.

sneakyimp:
I'm not sure this comment belongs here, but the CI 3 documentation sample code for memcached doesn't work. The docs need to be updated to reflect the more complex driver loading that actually works.

http://www.codeigniter.com/userguide3/libraries/caching.html#memcached-caching

cmeenatchisundaram:
@sneakyimp & @narfbg 
         One very important and easy to miss thing... If you are using memcache then save_path must have the "tcp://" prefix, if you are using memcached then the save_path should not have the "tcp://" prefix.

yahyaerturan:
I am using memcached for session handling and cache purposes. With the upgrade to CI 3.0.4, it starts to throw the following error (while it was working as expected on CI 3.0.3:

```
Unknown: Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/var/lib/php5)
```

If I replace CI 3.0.4's `Session_memcached_driver.php` file with 3.0.3's `Session_memcached_driver.php` it is working again.

[...]

PR #4424, opened 2016-02-03
Fix PHP session_write_close warning when writing empty session to Redis

jonty-comp:
I'm working on an application that will eventually save session data, but doesn't yet (other than for user logins).  Therefore, the session driver is loaded on each pageview but nothing is written to it.  I noticed that I was getting a `PHP warning: session_write_close(): Failed to write session data (user). Please verify that the current setting of session.save_path is correct ()`, and spent some time debugging.

It seems that if writing an empty session (that is, where `$session_data` is a blank string) to redis, the key isn't actually created in the redis store (expected redis behaviour, I guess).  Therefore, when the driver then runs the `setTimeout` function on that key, it returns 0, which triggers a failure in the `session_write_close` handler.

Therefore, my fix is to simply not set a timeout when the session fingerprint is equal to that of a blank session - as soon as some session data is added, the code runs as it originally did.

### DIFF
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -240,15 +240,9 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 				return $this->_failure;
 			}
 
-			if($this->_fingerprint === md5(''))
-			{
-				// A blank session will not be written to redis, so a timeout cannot be set on it
-				return $this->_success;
-			} else {
-				return ($this->_redis->setTimeout($this->_key_prefix.$session_id, $this->_config['expiration']))
-					? $this->_success
-					: $this->_failure;
-			}
+			return ($this->_redis->setTimeout($this->_key_prefix.$session_id, $this->_config['expiration']))
+				? $this->_success
+				: $this->_failure;
 		}
 
 		return $this->_failure;
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -69,6 +69,13 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 	 */
 	protected $_lock_key;
 
+	/**
+	 * Key exists flag
+	 *
+	 * @var boolean
+	 */
+	protected $_key_exists = FALSE;
+
 	// ------------------------------------------------------------------------
 
 	/**
@@ -189,7 +196,15 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 			// Needed by write() to detect session_regenerate_id() calls
 			$this->_session_id = $session_id;
 
-			$session_data = (string) $this->_redis->get($this->_key_prefix.$session_id);
+			$session_data = $this->_redis->get($this->_key_prefix.$session_id);
+
+			if ($session_data === FALSE)
+			{
+				// The session ID does not exist in redis yet, so set a flag to create it
+				$this->_key_exists = FALSE;
+				$session_data = '';
+			}
+
 			$this->_fingerprint = md5($session_data);
 			return $session_data;
 		}
@@ -229,7 +244,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 		if (isset($this->_lock_key))
 		{
 			$this->_redis->setTimeout($this->_lock_key, 300);
-			if ($this->_fingerprint !== ($fingerprint = md5($session_data)))
+			if ($this->_fingerprint !== ($fingerprint = md5($session_data)) OR $this->_key_exists === FALSE)
 			{
 				if ($this->_redis->set($this->_key_prefix.$session_id, $session_data, $this->_config['expiration']))
 				{
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -240,9 +240,15 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 				return $this->_failure;
 			}
 
-			return ($this->_redis->setTimeout($this->_key_prefix.$session_id, $this->_config['expiration']))
-				? $this->_success
-				: $this->_failure;
+			if($this->_fingerprint === md5(''))
+			{
+				// A blank session will not be written to redis, so a timeout cannot be set on it
+				return $this->_success;
+			} else {
+				return ($this->_redis->setTimeout($this->_key_prefix.$session_id, $this->_config['expiration']))
+					? $this->_success
+					: $this->_failure;
+			}
 		}
 
 		return $this->_failure;
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -198,16 +198,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 
 			$session_data = $this->_redis->get($this->_key_prefix.$session_id);
 
-			if ($session_data === FALSE)
-			{
-				// The session ID does not exist in redis yet, so set a flag to create it
-				$this->_key_exists = FALSE;
-				$session_data = '';
-			}
-			else
-			{
-				$this->_key_exists = TRUE;
-			}
+			is_string($session_data) && $this->_key_exists = TRUE;
 
 			$this->_fingerprint = md5($session_data);
 			return $session_data;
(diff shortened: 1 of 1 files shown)


narfbg:
What is your sess_save_path?

jonty-comp:
`$config['sess_save_path'] = 'tcp://localhost:6379?prefix=myapp_session:';`

I confirmed the redis behaviour in redis-cli: if you try to set EXPIRE on a key that doesn't exist, it returns 0.

narfbg:
Then how come the error message says it is empty? (it goes between those parentheses at the end)

Also, if you've verified that `setTimeout()` fails for a non-existent key, did you also verify that a key won't be created if its value is an empty string? I'm pretty sure an empty string is a perfectly valid value, while your logic is based on a guess.

jonty-comp:
I don't know why the error says it's empty - I mean, with this fix in place the whole thing works as expected! Shouldn't it also say (redis) instead of (user)? :/

Okay, perhaps the fix is in the wrong place.  Here's what happens in the redis-cli MONITOR when I load a page (without my fix applied, and with a brand new session in a private browsing window):
```
1454502868.822712 [0 127.0.0.1:48224] "TTL" "myapp_session:421d8f3c80b3a683275874ae81312af4417d7938:lock"
1454502868.822810 [0 127.0.0.1:48224] "SETEX" "myapp_session:421d8f3c80b3a683275874ae81312af4417d7938:lock" "300" "1454502868"
1454502868.822931 [0 127.0.0.1:48224] "GET" "myapp_session:421d8f3c80b3a683275874ae81312af4417d7938"
1454502868.823370 [0 127.0.0.1:48228] "SMEMBERS" "_ci_redis_serialized"
1454502868.864365 [0 127.0.0.1:48224] "EXPIRE" "myapp_session:421d8f3c80b3a683275874ae81312af4417d7938:lock" "300"
1454502868.864462 [0 127.0.0.1:48224] "EXPIRE" "myapp_session:421d8f3c80b3a683275874ae81312af4417d7938" "7200"
1454502868.864670 [0 127.0.0.1:48224] "PING"
1454502868.864769 [0 127.0.0.1:48224] "DEL" "myapp_session:421d8f3c80b3a683275874ae81312af4417d7938:lock"
```

I guess it isn't trying to set the key in the first place? Either that, or it is, and phpredis is returning success from the set function while not actually creating the key.

narfbg:
> I don't know why the error says it's empty - I mean, with this fix in place the whole thing works as expected!

Well, it shouldn't be empty ... I'd try debugging that first.
And I can assure you it's not working as expected. :)

> Shouldn't it also say (redis) instead of (user)? :/

No, that's the name of the session handler module, which is always 'user' for us.

> I guess it isn't trying to set the key in the first place?

You're on the right track with that. Look here: https://github.com/bcit-ci/CodeIgniter/blob/e54c461d77ffce9f10999ca51bb2e3932304639b/system/libraries/Session/drivers/Session_redis_driver.php#L232

But before you modify your patch, please read our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html), as your current code is incompatible with it.

jonty-comp:
Okay, i've been working on this a bit more.  I've worked out the following:

- When the session library is loaded, it calls `read` to get the current session data
- If that key doesn't exist, because the session is new, then phpredis returns FALSE, but [system/libraries/Session/drivers/Session_redis_driver.php#L192](https://github.com/bcit-ci/CodeIgniter/blob/e54c461d77ffce9f10999ca51bb2e3932304639b/system/libraries/Session/drivers/Session_redis_driver.php#L192) casts this to string, which results in an empty string
- The session fingerprint is then set to `md5($session_data)`, which equates to the md5 of an empty string
- If the session is not modified during the page execution, then when `session_write_close` is called at the end, the fingerprint has not changed and the `set` function is not called, so the key still doesn't exist in redis
- However, the `setTimeout` function is still called regardless to update the expiration, resulting in that function returning FALSE (on the non-existent key) and a PHP warning is generated.

I'm not sure of the best way to fix this.  I guess one way would be to check for FALSE in the read function, and then call a write of an empty string there, but that seems messy.  Hopefully you can understand what's going on from my terrible explanation.

narfbg:
Yea, I know what's going on. That's why I pointed you to that line in `write()` - that's what needs to be changed. :)

jonty-comp:
Okay, so my options are to either always update the redis regardless of whether the fingerprint has changed or not, or set a flag in the class at read-time telling it that the key doesn't exist, so it knows to force a write at write-time...unless i'm still missing something obvious? Sorry, this should be trivial, but I don't want to go down the wrong route. :P

narfbg:
If you ignore the fingerprint, you're making it useless, so that leaves you with the flag option. The 'database' driver does something similar.

jonty-comp:
Hopefully these changes are alright.  I've already backported them into my 3.0.4 project!

narfbg:
Nope ... you're never setting the flag to TRUE.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -69,6 +69,13 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 	 */
 	protected $_lock_key;
 
+	/**
+	 * Key exists flag
+	 *
+	 * @var boolean
```
narfbg: ... and we use 'bool' instead of 'boolean'.

jonty-comp:
...I'll get there eventually!

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -189,7 +196,15 @@ public function read($session_id)
 			// Needed by write() to detect session_regenerate_id() calls
 			$this->_session_id = $session_id;
 
-			$session_data = (string) $this->_redis->get($this->_key_prefix.$session_id);
+			$session_data = $this->_redis->get($this->_key_prefix.$session_id);
+
+			if ($session_data === FALSE)
```
narfbg: This whole block is unnecessary.

FALSE is already the default, you don't need to set it again.  
`md5(FALSE)` will produce the same result as `md5('')` because of an implicit typecast to string, so you don't need that either.

Just replace the whole thing with `is_string($session_data) && $this->_key_exists = TRUE;`

yahyaerturan:
Similar problem with memcached. It was working as expected on v3.0.3. After updating to 3.0.4, it throws exception:

```
Unknown: Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/var/lib/php5)
```

*config.php*

```php
$config['sess_driver'] = 'memcached';
$config['sess_cookie_name'] = 'cisession';
$config['sess_expiration'] = 7200;
$config['sess_save_path'] = 'localhost:11211';
$config['sess_match_ip'] = false;
$config['sess_time_to_update'] = 300;
$config['sess_regenerate_destroy'] = false;
```

*config/memcached.php* is untouched. I already told that it was working nicely on 3.0.3.

jonty-comp:
Sigh, it seems like something i've changed has caused a regression that triggers #4362 again!

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -222,18 +232,19 @@ public function write($session_id, $session_data)
 				return $this->_failure;
 			}
 
-			$this->_fingerprint = md5('');
```
narfbg: Why this? Haven't checked, but it could be what's causing the regression you mentioned.

Also, you don't need to manually set the flag to FALSE anywhere at all.

narfbg:
@yahyaerturan Your problem is nothing similar to this, and anyway I don't see how it could've been working in 3.0.3 if it gives you this error now.

Either way, please post on the [forums](http://forum.codeigniter.com/) or #3919.

system/libraries/Session/drivers/Session_redis_driver.php
jonty-comp: It was my understanding that the session is read with the old ID (so the flag is set to TRUE in the `read` function), then PHP regenerates the ID, and when it comes to `write` it the flag should be set back to FALSE in this if-block as a key with that new ID doesn't exist.  I think the fingerprint line I removed resulted in the same behaviour, but resetting the flag seemed more semantic.

Either way, I'm still getting the regression when I revert back to the fingerprint line. :(

narfbg: Hmm, makes sense.

And I just checked the database driver, which does the same thing, so ... ignore my last comment. :)

narfbg:
Oh, I see it ... the regression is kind of my fault, sorry.

`$this->_failure` is FALSE under PHP7, so that line I gave you kind of broke it. It should be this instead:

    is_string($session_data)
        ? $this->_key_exists = TRUE
        : $session_data = '';

jonty-comp:
That makes sense.  It's a happy coincidence that my dev environment is running PHP7!

narfbg:
Hopefully more people would have that. :)

Congrats on your first PR merged.

PR #4432, opened 2016-02-04
Add is_error function in form helper

vichenze:
This function helps to know if there is an error on a specific field, so you can an eventually add a class to it :

`<input type="email" name="email" id="email" value="<?php echo set_value('email'); ?>" <?php echo is_error('email') ? 'class="error"' : ''; ?> />`


narfbg:
empty(form_error('email')) ? 'class="error"' : '';

PR #4433, opened 2016-02-04
Avoid warnings on IDE

sierrafayad:
This is not a bug itself. but i was getting warnings on PHPstorm. if it
happens to be non callable, returns in_progress false


narfbg:
Duplicate of #4246.

PR #4437, opened 2016-02-07
hunanize() helper: Escaping the $separator argument.

ivantcholakov:
Since the separator could be arbitrary, it needs to be preg-escaped.


narfbg:
Just FYI, you've made a typo in the function name.

ivantcholakov:
Oh, sorry.

PR #4438, opened 2016-02-07
Create Add Foreign Key Function

EpicKris:
Created `CI_DB_forge->add_foreign_key()` function and added documentation to user guide.

### DIFF
--- a/system/database/DB_forge.php
+++ b/system/database/DB_forge.php
@@ -75,6 +75,13 @@ abstract class CI_DB_forge {
 	public $primary_keys	= array();
 
 	/**
+	 * Foreign keys.
+	 *
+	 * @var array Foreign keys.
+	 */
+	public $foreign_keys = array();
+
+	/**
 	 * Database character set
 	 *
 	 * @var	string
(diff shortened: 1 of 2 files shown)


EpicKris:
Slightly different implementation of #4275 to keep inline with existing `CI_DB_forge->add_key()` functionality.

steve-goodwin:
Good work on this @EpicKris , hopefully it'll be merged soon :)

EpicKris:
@narfbg Could you review this pull request and let me know if I should make any changes? Thanks!

narfbg:
Did you not see my comment on #4275?

EpicKris:
@narfbg Sorry, but I don't understand why this would only work for MySQL and not other database drivers?

narfbg:
Because I'm yet to see a query other than very, very basic SELECT or DELETE to work across all databases. You just can't assume what works on MySQL will work for everything else ... asking "why wouldn't it" isn't serious.

Obviously, nobody has all of them in order to test, but you can at least consult different documentations to confirm it.

EpicKris:
I'll test this with different database drivers and work out the differences, almost certain this will work for MySQL, PostgreSQL, and Microsoft SQL though.

narfbg:
Those are only 3 out of ~10 ... and I'm pretty sure it doesn't work *out of the box* on SQLite.

EpicKris:
@narfbg What if a database doesn't support foreign keys?

narfbg:
Well then it obviously will fail, but it has to fail gracefully, not by virtue of the query just refusing to execute. :)

steve-goodwin:
@EpicKris Yeah I can put some time aside to review documentation on this to see where it will fail, although re-factoring bits of code may take me longer as i'm also a little stacked at the moment though. This is a feature I need in pretty much every project though so willing to invest time into getting this into an acceptable state for a release candidate

EpicKris:
@steve-goodwin Likewise, I'd like to get this in as soon as possible as I'd like to use it for a school project.

EpicKris:
@steve-goodwin I'm going to start having more time on my hands, so will look into this, I also want a way of creating and deleting foreign keys independently of creating a table, deleting foreign keys may be required for downgrading migrations.

weblogics:
@EpicKris - sorry I've been flat out on other projects, had a look through your commits and looking good. What's left to get this feature complete and passing?

steve-goodwin:
@EpicKris - sorry I've been flat out on other projects, had a look through your commits and looking good. What's left to get this feature complete and passing?

EpicKris:
@steve-goodwin, I've also been flat out busy with my new job, obviously the build is failing at the moment, so that needs checking out, and I'd like a way to be able to add and drop foreign keys independently of creating a table (adding and dropping foreign keys may be required for migrations, etc…), and then each database driver needs to be checked for compatibility and updated accordingly.

PR #4440, opened 2016-02-07
Adding Enum Properties

lukmannhakim:
Sorry if my english not so good
Code igniter  just provides field metadata like this
$fields = $this->db->field_data('table_name');

foreach ($fields as $field)
{
   echo $field->name;
   echo $field->type;
   echo $field->max_length;
   echo $field->primary_key;
}

Then i added extra properties called enum, that contain  values of enum data type
if($type=='enum')
			{
				$enum=$field->Type;
				$values="";
				$stack="";
    			$off = strpos($enum,"(");
    			$enum = substr($enum, $off+1, strlen($enum)-$off-2);
    			$values = explode(",",$enum);
				$stack[0]=Count($values);
	
    			for( $n = 1; $n <= Count($values); $n++) 
				{
    			$val = substr( $values[$n-1], 1,strlen($values[$n-1])-2);
    			$val = str_replace("''","'",$val);
    			$stack[$n]=$val;
     			}
			}

////////

			$F				= new stdClass();
			$F->name		= $field->Field;
			$F->type		= $type;
			$F->default		= $field->Default;
			$F->max_length	= $length;
			$F->primary_key = ( $field->Key == 'PRI' ? 1 : 0 );
		 $F->enum= $stack;

### DIFF
--- a/system/database/drivers/mysql/mysql_result.php
+++ b/system/database/drivers/mysql/mysql_result.php
@@ -119,14 +119,47 @@ class CI_DB_mysql_result extends CI_DB_result {
 	 */
 	public function field_data()
 	{
-		$retval = array();
-		for ($i = 0, $c = $this->num_fields(); $i < $c; $i++)
+	$retval = array();
+		while ($field = mysql_fetch_object($this->result_id))
 		{
-			$retval[$i]			= new stdClass();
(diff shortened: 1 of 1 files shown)


narfbg:
This breaks A LOT of things.

lukmannhakim:
why

Issue #805, opened 2011-12-21
cascading views doesn't work correctly for added application packages


somomomo:
I wrote a more detailed post on StackOverflow about this (http://stackoverflow.com/questions/7759723/code-igniter-2-0-3-loader-add-package-path-view-cascading) but here's a quick rundown:

When adding an application package, if there is a naming collision in view files, setting the 2nd parameter in add_package_path to FALSE does not load the view file from the original environment and will still load the view file from the added package.

The bug is in the _ci_load function in the core Loader class.  It starts at line 764 in the version 2.1.0 file where there's a check for the view file and the cascade value in the foreach loop.  Since there's a check if the file exists first, the cascade value is irrelevant in determining whether the view file of the added package should be loaded.  Thus, the cascade check does nothing if there is an actual conflict and the view file will always be loaded from the added package.

This bug has existed since at least 2.0.3 and I extended the_ci_load method in order to get around this.  What I did was basically switch the order of the check for the view file and the cascade value so the cascade check is done first in the loop.  Also, in the cascade check, I changed the break to be a continue.  I'm actually thinking it may be better to put the check for cascade inside the beginning of the check for the view file but putting it before the view file check has worked for me.

ViVers:
mwhitneysdsu code is correct. I use this for my multiple application in one instance solution and it is working. Please add it to CI for everyone  ;)

Xirt:
Noticed bug is still there after six years; fix in pull request #4444 is still working in current version though, but only requires modification of one lines and addition of another line (the other modifications in #4444 are merely lay-out changes to the code). 

Only line to add is `$_rev_view_paths = array_reverse($this->_ci_view_paths, true);` just above the foreach-loop and refer the new variable `$_rev_view_paths` in the loop. Keep in mind that you need to do the same in MX_Loader in case you use the popular extension HMVC.

PR #4444, opened 2016-02-09
Fix Cascade View

ViVers:
Fixed cascade view for loading packages by (mwhitneysdsu)

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -887,18 +887,16 @@ class CI_Loader {
 			$_ci_ext = pathinfo($_ci_view, PATHINFO_EXTENSION);
 			$_ci_file = ($_ci_ext === '') ? $_ci_view.'.php' : $_ci_view;
 
-			foreach ($this->_ci_view_paths as $_ci_view_file => $cascade)
+			foreach ($this->_ci_view_paths as $view_file => $cascade)
 			{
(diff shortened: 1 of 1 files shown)


mwhitneysdsu:
If I remember correctly (and only because I received a notification on #805 ), I also mentioned that the order of `$this->_ci_view_paths` may be the reverse of what is expected, which would result in this change causing a different view to load in cases where the view is found in more than one location in `$this->_ci_view_paths`. Preserving the existing load order for the views would require the array to be reversed before looping through it. Something along the lines of:

```php
			$_rev_view_paths = array_reverse($this->_ci_view_paths, true);
			foreach ($_rev_view_paths as $_ci_view_file => $cascade)
			{
				if (file_exists($_ci_view_file.$_ci_file))
				{
					$_ci_path = $_ci_view_file.$_ci_file;
					$file_exists = TRUE;
					if ( ! $cascade)
					{
						break;
					}
				}
			}
```

Also, if you're going to change the name of `$_ci_view_file` inside the loop, it would probably be better to use a more accurate name, like `$view_path`, though I'd probably use `$_ci_view_path` to remain consistent with the existing code. There's not much point in changing it if it becomes less consistent with the rest of the code *and* continues to misrepresent the data stored in that variable.

PR #4445, opened 2016-02-09
Update DB_query_builder.php - phpdoc (some incorrect return type)

damiengrandi:
Edit of some return type in phpdoc. This can also help to fix some autocompletion troubles in some IDE.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1138,7 +1138,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 * @param	string	$key
 	 * @param	string	$value
 	 * @param	bool	$escape
-	 * @return	object
+	 * @return	CI_DB_query_builder
 	 */
 	public function having($key, $value = NULL, $escape = NULL)
 	{
@@ -1155,7 +1155,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 * @param	string	$key
 	 * @param	string	$value
 	 * @param	bool	$escape
-	 * @return	object
+	 * @return	CI_DB_query_builder
 	 */
 	public function or_having($key, $value = NULL, $escape = NULL)
 	{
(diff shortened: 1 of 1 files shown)


mwhitneysdsu:
Most of the return types you've marked as `CI_DB_query_builder` would probably return `CI_DB` at runtime. The `delete()` method would return `string|bool|void`. There may be some additional variations based on the driver in use, as well.

system/database/DB_query_builder.php
```diff
@@ -1456,7 +1456,7 @@ public function get_where($table = '', $where = NULL, $limit = NULL, $offset = N
 	 * @param	string	$table	Table to insert into
 	 * @param	array	$set 	An associative array of insert values
 	 * @param	bool	$escape	Whether to escape values and identifiers
-	 * @return	int	Number of rows inserted or FALSE on failure
+	 * @return	int|bool	 Number of rows inserted or FALSE on failure
```
narfbg: We don't really use this format, and the rare case of boolean FALSE returns has been discussed before - it's practically irrelevant... please revert this.

system/database/DB_query_builder.php
```diff
@@ -1863,7 +1863,7 @@ protected function _validate_update($table)
 	 * @param	string	the table to retrieve the results from
 	 * @param	array	an associative array of update values
 	 * @param	string	the where key
-	 * @return	int	number of rows affected or FALSE on failure
+	 * @return	int|bool	 Number of rows affected or FALSE on failure
```
narfbg: Same as for `insert_batch()`.

system/database/DB_query_builder.php
```diff
@@ -2114,7 +2114,7 @@ public function get_compiled_delete($table = '', $reset = TRUE)
 	 * @param	mixed	the where clause
 	 * @param	mixed	the limit clause
 	 * @param	bool
-	 * @return	mixed
+	 * @return	CI_DB_query_builder|bool    CI_DB_query_builder instance or FALSE on failure
```
narfbg: It can also return void ... mixed is what should stay here.

narfbg:
`CI_DB` would still be a valid `CI_DB_query_builder` instance in all of these cases - don't worry about that.

damiengrandi:
Ok, I updated

PR #4448, opened 2016-02-10
Fix flash data not being cleared when using native sessions

coladict:
Using the file driver (why that exists at all is a mystery) we got file permission problems we could not resolve, even with `chmod -R 777` to give everyone rights to write in the designated directory, so we switched to native drivers using `session.auto_start` in php.ini, after taking a look at the Session.php implementation to learn that will stop it. This is how we discovered that flash data persisted when using PHP's native sessions.


narfbg:
Why you expect "flashdata" to work without using the drivers provided by CI is a mystery.

PR #4453, opened 2016-02-11
Autoload Driver Object Names

EpicKris:
Updated loader to autoload drivers with object names the same as autoloading libraries.

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -72,6 +72,11 @@ $autoload['libraries'] = array();
 | Prototype:
 |
 |	$autoload['drivers'] = array('cache');
+|
+| You can also supply an alternative library name to be assigned in
+| the controller:
+|
+|   $autoload['drivers'] = array('cache' => 'cch');
 */
 $autoload['drivers'] = array();
 
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -696,7 +696,7 @@ class CI_Loader {
 				}
 				else
 				{
-					$this->driver($key, $params, $value)
+					$this->driver($key, $params, $value);
 				}
 			}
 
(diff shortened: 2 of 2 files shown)


narfbg:
What's the point? Only cache uses this.

EpicKris:
I'm using a package that has a name conflict with one of my controllers which means I'm unable to use the driver from that package.

narfbg:
What package?

EpicKris:
It's an internal package used across several projects and clients.

narfbg:
You've actually developed a package based on CI_Driver? You're probably one of 5 people to ever want to do that. :)

Seriously, I'd rather discourage CI_Driver usage and this enhancement would be the opposite of that. If CI_Cache didn't use it, it would've been gone by now.

EpicKris:
Yes, I've actually developed 2 driver libraries; one for authentication, and one for storage which works with the local file system, Amazon Web Services, etc…

Why would you discourage `CI_Driver` usage? It seems like a good concept to me?

narfbg:
Libraries using adapters for different platforms is a good concept, `CI_Driver` is not that good of a concept. :)

It puts more limits on you than it helps and you don't really need a library called "Driver" to write libraries that have drivers/adapters. It's also largely undocumented and hard to understand - even I don't know exactly how it works, I'm impressed you've managed to put it into use.

EpicKris:
I suppose I always just thought of drivers similar to the database library and use it in much similar way too which is different to what I think drivers were intended for, but it works. I could show you an example if you'd like?

narfbg:
No, I don't really care to see it. You know what? I don't really care that much about this either, so I'll merge it.

narfbg:
Or rather, I'll merge it after you fix the syntax error on line 699. Please don't submit patches that you didn't test ...

narfbg:
Please leave a comment next time, commits don't issue notifications.

EpicKris:
Sorry, didn't realise, thanks for merging.

PR #4466, opened 2016-02-17
update to config/database to support port for mysqli

xims:
If you need to use alternative port number for mysqli connection, 'hostname:port' will not work.
Port needs to be specified separately.
Codeigniter's mysqli driver does supports separate port variable but it's not mentioned in the config file.


andersonunsonst:
https://github.com/bcit-ci/CodeIgniter/blob/develop/user_guide_src/source/database/configuration.rst

narfbg:
That's a nice catch, but it also needs to be added in the comment descriptions above the actual configuration array. And to the configuration docs linked by @andersonunsonst too.  
Also, ports are integers, not strings ... an empty string doesn't make sense as the default; please use NULL instead.

And one more note: almost all drivers support it, it's not just mysqli.

PR #4467, opened 2016-02-17
Redis performance optimization

gyyinzjut:
Bad performance when there are a great many of redis keys,so serialize all the keys when save keys.


narfbg:
No, this just moves the performance penalty from Redis to PHP and it impacts non-array values as well.

gyyinzjut:
You mean we should try to limit array values use ?

gyyinzjut:
You mean we should try to limit array values use?

narfbg:
Yes, that's the source of your problem - Redis doesn't have native support for that structure and therefore you cannot expect it to behave well with it.

PR #4470, opened 2016-02-18
update to config/database to support non default port for DB server

xims:
If you need to use non default port number for DB connection, 'hostname:port' may not work.
Port may need to be specified separately.
Codeigniter's mysqli and other drivers do supports separate port variable but it's not mentioned in the config file.

### DIFF
--- a/application/config/database.php
+++ b/application/config/database.php
@@ -16,6 +16,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 |
 |	['dsn']      The full DSN string describe a connection to the database.
 |	['hostname'] The hostname of your database server.
+|	['port']     The port of your database server (if not default)
 |	['username'] The username used to connect to the database
 |	['password'] The password used to connect to the database
 |	['database'] The name of the database you want to connect to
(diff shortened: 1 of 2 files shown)


user_guide_src/source/database/configuration.rst
```diff
@@ -194,6 +198,8 @@ Explanation of Values:
 **stricton**		TRUE/FALSE (boolean) - Whether to force "Strict Mode" connections, good for ensuring strict SQL
 			while developing an application.
 **port**		The database port number. To use this value you have to add a line to the database config array.
+
+			Some drivers (like mysql) allow specifying port as a part of hostname (localhost:3306) but others (like mysqli) require separate port value if it's not default.
```
narfbg: Remove this.

application/config/database.php
```diff
@@ -76,6 +77,7 @@
 $db['default'] = array(
 	'dsn'	=> '',
 	'hostname' => 'localhost',
+	'port' => null,
```
narfbg: We use upper-case - NULL, not null. You have to change every other occurence of that too.

Also, please align the assigments so that they start from the same column, like you've done in the comment description above.

narfbg:
And a nitpicky note: you're not updating the config to support a port - it supports it, just doesn't have a ready to go placeholder for it.

You can make your changes to this PR, you don't have to open a new one for every change. Pull requests can have many changes before they are accepted ... if contributors open a new one every time we'd have thousands of obsolete ones like #4466.

PR #4471, opened 2016-02-18
Use real IP in session driver if you're behind a proxy

remiheens:
Hi, 

I'm working on a login library using memcache with CI, but I've discovered that each session driver use "$_SERVER['REMOTE_ADDR']" to prefix the session key. 
But if you're running Codeigniter behind a proxy, $_SERVER['REMOTE_ADDR'] is always 127.0.0.1, and the driver doesn't use configured proxy_ips to get real client ip and the sess_match_ip option doesn't work. 

So, I submit my little contribution to Codeigniter :)

### DIFF
--- a/system/libraries/Session/Session_driver.php
+++ b/system/libraries/Session/Session_driver.php
@@ -86,6 +86,15 @@ abstract class CI_Session_driver implements SessionHandlerInterface {
 	 */
 	protected $_success, $_failure;
 
+	/**
+	 * Client IP if sess_match_ip used
+	 *
+	 * Necessary if you're running codeigniter behind proxy
+	 *
+	 * @var string
+	 */
+	protected $_client_ip;
+
 	// ------------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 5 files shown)


narfbg:
Duplicate of #4003.

remiheens:
Ok, but what is your solution ?

narfbg:
I don't have a specific solution, but that's irrelevant - as far as I and CodeIgniter are concerned, trusting a client-provided IP address is a security flaw and not an option.

It's up to you to come up with another solution for whatever you're trying to do.

remiheens:
It is my nginx configuration who forward query on apache/php by proxy_pass and client ip was added to header with this configuration : 

proxy_set_header   X-Real-IP        $remote_addr;
proxy_set_header   X-Forwarded-For  $proxy_add_x_forwarded_for;

Do you think it's a security flaw ?

narfbg:
*You* know that it's your nginx configuration and that you could trust it. The code doesn't.

remiheens:
So, in my case. I can create a subdriver of Session_memcached_driver that use real ip. Do you think it's a good alternative ?

narfbg:
No, but please note that this is a bug tracker ... we don't answer general help requests here.
Please ask your questions on our [forums](http://forum.codeigniter.com/) instead.

PR #4472, opened 2016-02-20
Update CodeIgniter.php

vibbow:
Declare get_instance() return type.

So IDE can detect return type of use get_instance().
Therefore show the correct autocomplete hint.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -359,7 +359,7 @@ if ( ! is_php('5.4'))
 	 *
 	 * Returns current CI instance object
 	 *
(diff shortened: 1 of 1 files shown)


narfbg:
You're supposed to replace "object" with "CI_Controller" ...
Also, we use tabs to separate the docblock columns.

vibbow:
Oh, sorry. I suppose to replace object with CI_Controller

narfbg:
So ... are you going to fix that?

vibbow:
Fixed now.

FIrst time learned actually I can modify a pull request :)

PR #4474, opened 2016-02-21
Check to see whether the cache driver(redis/memcache(d)) is supported when it is initialized

bjjay:
When setting a cache driver(redis/memcache(d)) but it isn't supported on the web server, the server will return "fatal error". So we need check whether the cache driver is supported when it is initialized.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_apc.php
+++ b/system/libraries/Cache/drivers/Cache_apc.php
@@ -198,13 +198,13 @@ class CI_Cache_apc extends CI_Driver {
 	 */
 	public function is_supported()
 	{
-        if (extension_loaded('apc') AND ini_get('apc.enabled'))
-        {
-            return TRUE;
-        }
+		if (extension_loaded('apc') AND ini_get('apc.enabled'))
+		{
+			return TRUE;
+		}
 
-        log_message('debug', 'The APC PHP extension must be loaded and enabled to use APC Cache.');
-        return FALSE;
+		log_message('debug', 'The APC PHP extension must be loaded and enabled to use APC Cache.');
+		return FALSE;
 	}
 
 }
--- a/system/libraries/Cache/drivers/Cache_apc.php
+++ b/system/libraries/Cache/drivers/Cache_apc.php
@@ -198,13 +198,13 @@ class CI_Cache_apc extends CI_Driver {
 	 */
 	public function is_supported()
 	{
-		if ( ! extension_loaded('apc') OR ! ini_get('apc.enabled'))
-		{
-			log_message('debug', 'The APC PHP extension must be loaded to use APC Cache.');
-			return FALSE;
-		}
+        if (extension_loaded('apc') AND ini_get('apc.enabled'))
+        {
+            return TRUE;
+        }
 
-		return TRUE;
+        log_message('debug', 'The APC PHP extension must be loaded and enabled to use APC Cache.');
+        return FALSE;
 	}
 
 }
--- a/system/libraries/Cache/drivers/Cache_file.php
+++ b/system/libraries/Cache/drivers/Cache_file.php
@@ -67,10 +67,10 @@ class CI_Cache_file extends CI_Driver {
 		$path = $CI->config->item('cache_path');
 		$this->_cache_path = ($path === '') ? APPPATH.'cache/' : $path;
 
-        if (!$this->is_supported())
-        {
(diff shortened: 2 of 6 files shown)


system/libraries/Cache/drivers/Cache_apc.php
```diff
@@ -198,13 +198,13 @@ public function get_metadata($id)
 	 */
 	public function is_supported()
 	{
-		if ( ! extension_loaded('apc') OR ! ini_get('apc.enabled'))
+		if (extension_loaded('apc') AND ini_get('apc.enabled'))
```
narfbg: Why are you reversing the logic? You don't need to touch this file at all.

Also, we use `&&` instead of `AND` - [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html)

system/libraries/Cache/drivers/Cache_file.php
```diff
@@ -66,6 +66,11 @@ public function __construct()
 		$CI->load->helper('file');
 		$path = $CI->config->item('cache_path');
 		$this->_cache_path = ($path === '') ? APPPATH.'cache/' : $path;
+
+		if (!$this->is_supported())
```
narfbg: This *again* doesn't comply with our [styleguide](http://www.codeigniter.com/userguide3/general/styleguide.html).

system/libraries/Cache/drivers/Cache_file.php
```diff
@@ -66,6 +66,11 @@ public function __construct()
...
+		{
+			log_message('error', 'The cache directory is indeed writable ：' . $this->_cache_path);
```
narfbg: The word "indeed" doesn't mean what you think it means.

Also, we don't put spaces before colons, nor around the concatenation operator.

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -79,6 +79,12 @@ class CI_Cache_memcached extends CI_Driver {
 	 */
 	public function __construct()
 	{
+		if (!$this->is_supported())
```
narfbg: http://www.codeigniter.com/userguide3/general/styleguide.html

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -100,13 +106,9 @@ public function __construct()
 		{
 			$this->_memcached = new Memcached();
 		}
-		elseif (class_exists('Memcache', FALSE))
```
narfbg: Why are you removing this?

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -91,6 +91,12 @@ class CI_Cache_redis extends CI_Driver
 	 */
 	public function __construct()
 	{
+		if (!$this->is_supported())
```
narfbg: http://www.codeigniter.com/userguide3/general/styleguide.html

system/libraries/Cache/drivers/Cache_wincache.php
```diff
@@ -194,13 +194,13 @@ public function get_metadata($id)
 	 */
 	public function is_supported()
 	{
-		if ( ! extension_loaded('wincache') OR ! ini_get('wincache.ucenabled'))
+		if (extension_loaded('wincache') AND ini_get('wincache.ucenabled'))
```
narfbg: Same as APC

user_guide_src/source/changelog.rst
```diff
@@ -38,7 +38,7 @@ Release Date: Not Released
 
 Bug fixes for 3.0.5
 -------------------
-
```
narfbg: Don't remove this

user_guide_src/source/changelog.rst
```diff
@@ -38,7 +38,7 @@ Release Date: Not Released
...
+-  Fixed a bug (#4277) - :doc:`Cache Library <libraries/caching>` Check to see whether the cache driver(file/redis/memcache(d)) is supported when it is initialized.
```
narfbg: We put new changelog messages at the *bottom*

narfbg:
Your previous PR (#4376) had problems with its presentation, not the patch. Why did you turn a one-liner into a whole bulk of changes that are mostly unnecessary?

Also, please reference issues from the PR title/description, not from commits in your fork.  
Relevant: #4277

PR #4475, opened 2016-02-21
Update core/Security.php to fix strip_image_tags() URL displaying bug

medhavini:
The earlier version of strip_image_tags() displays only the first letter of image URL when neither single quotes or double quotes are used to enclose the image URL.

This is fixed by making the regex greedy again, and it can be non-greedy if URL is followed by at least 1 whitespace.

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -762,7 +762,7 @@ class CI_Security {
 	 */
 	public function strip_image_tags($str)
 	{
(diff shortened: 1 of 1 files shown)


rajatsharma94:
Yeah, if you use : 
or : 
you get : www.example.com/smiley.gif (the complete URL) as output.

But if you use :  (which is a valid image tag)
then the output is : w

narfbg:
Applied a slightly different patch, thanks for reporting the issue.

PR #4491, opened 2016-02-28
clean current lock key in redis session driver

roastduck:
set $this->_lock_key to NULL after close

When calling session_start() again to re-use session after calling session_write_close(), the driver will call _get_lock(). Inside _get_lock, it will check whether the lock has been already gained by checking _lock_key variable. But before this commit, when calling session_write_close(), the close() method did not clear the _lock_key. Next, when calling _get_lock(), it found _lock_key already exist but the lock had been released, so _get_lock failed.

Signed-off-by: roastduck <beantang.tang@gmail.com>

### DIFF
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -277,7 +277,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 			try {
 				if ($this->_redis->ping() === '+PONG')
 				{
-					$this->_release_lock($this->_lock_key);
+					$this->_release_lock();
 					if ($this->_redis->close() === $this->_failure)
 					{
 						return $this->_failure;
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -277,8 +277,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 			try {
 				if ($this->_redis->ping() === '+PONG')
 				{
-					isset($this->_lock_key) && $this->_redis->delete($this->_lock_key);
-					$this->_lock_key = NULL;
+					$this->_release_lock($this->_lock_key);
 					if ($this->_redis->close() === $this->_failure)
 					{
 						return $this->_failure;
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -278,6 +278,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 				if ($this->_redis->ping() === '+PONG')
 				{
 					isset($this->_lock_key) && $this->_redis->delete($this->_lock_key);
+					$this->_lock_key = NULL;
 					if ($this->_redis->close() === $this->_failure)
 					{
 						return $this->_failure;


system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -278,6 +278,7 @@ public function close()
 				if ($this->_redis->ping() === '+PONG')
 				{
 					isset($this->_lock_key) && $this->_redis->delete($this->_lock_key);
```
narfbg: You could just replace the `delete()` call here with one to `_release_lock()`.

narfbg:
And not that it matters for the patch itself, but why do you need to do this at all?

roastduck:
I do this because I found unable to start a session again after calling `session_write_close()`. When there is already a `$this->_lock_key` created when using the same session object before (before `session_write_close()`), but there is actually no lock (deleted by calling `session_write_close()`), `_get_lock()` will not create a new lock.

I need to start a session again just because I have a long poll, which checks the session every few seconds.

roastduck:
Should I put line 280 into the `if` in line 281? The original code doesn't take this as a failure.

narfbg:
I meant why do you need to call `session_write_close()` and then `session_start()` again for the same session. You shouldn't do that.

> Should I put line 280 into the `if` in line 281? The original code doesn't take this as a failure.

No, if it fails it's either because the lock is already released (in which case it doesn't matter) or because the connection is broken, in which case you can't do anything about it and `close()` will fail anyway.

However, please put back the `isset()` check, all you had to do is change:

> isset($this->_lock_key) && $this->_redis->delete($this->_lock_key);

to:

> isset($this->_lock_key) && $this->_release_lock();

roastduck:
> I meant why do you need to call session_write_close() and then session_start() again for the same
> session. You shouldn't do that.

In the long poll, I check the session and then sleep, for several times. When the thread sleeps, it need to release the lock and let other thread to use it. What should I do if I don't do this? I didn't find a public method just to release the lock rather than release the lock and close.

> However, please put back the isset() check, all you had to do is change:

I think `isset($this->_lock_key)` is already checked in `$this->_release_lock()` and I don't need to do it again.

narfbg:
> In the long poll, I check the session and then sleep, for several times. When the thread sleeps, it need to release the lock and let other thread to use it. What should I do if I don't do this? I didn't find a public method just to release the lock rather than release the lock and close.

I understand why you need the `session_write_close()` call and that's completely natural, but starting the session again after that doesn't make sense ... Whatever response you get for subsequent polls, it's not going to be from that same request. `session_start()` does nothing useful for you, it can only cause problems.

> I think isset($this->_lock_key) is already checked in $this->_release_lock() and I don't need to do it again.

True.

roastduck:
In my project, the long poll is used for checking if something in the server is updated and need to remind the user, such as a new mail. Every time I check if the current data the same as I store in the session, if not, I push it back to the user. Because every page open in the browser will call the same long poll with AJAX, I choose to use session to share the newest data among the threads. I'm wondering what's the better practice.

narfbg:
I understand the use-case. What I'm saying is that `session_start()` doesn't help you in any way and you can remove it.

That being said, if you want to receive help about this and/or further discuss it, please open a thread on our [forums](http://forum.codeigniter.com/). We only deal with bugs here.

danielzev:
Hey guys, is there a planned date on the v3.0.5 release for this fix?

PR #4494, opened 2016-02-28
Support for using regex in the $field parameter of Form_validation::set_rules(), issue#1812

abude:
Made some changes to have this "Support for using regex in the $field parameter of Form_validation::set_rules() , issue#1812" work, hope it's good to go.

Regards,

### DIFF
--- a/TestReadMe
+++ b/TestReadMe
@@ -1,0 +1,1 @@
+Testing stuff around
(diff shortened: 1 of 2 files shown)


abude:
Please let me know if it's fine or if i need to change something, and also should we close the issue#1812?

Kind regards

Issue #4503, opened 2016-03-04
PDO Driver always returns empty results


Murtnowski:
When using the pdo driver with dblib the result sets from SELECT queries are always empty.

I traced the error down and here's what's happening:

In system/database/drivers/pdo/pdo_driver.php fetchAll() is used to count the affected rows, however the resource pointer is never reset before fetching the result set.

In system/database/DB_result.php _data_seek(0) is used to reset the resource pointer, but when you go to the actual implementation in system/database/drivers/pdo/pdo_result.php the _data_seek() function simply returns FALSE and no other work is done.

Possible solutions:
Either implement _data_seek() in the pdo_result.php file, or don't have pdo_driver.php use fetchAll() on SELECT statements.

Honestly I think pdo_driver.php is the problem.  Currently it looks at the number of rows in a SELECT statement and records them as the affected rows, but that feels wrong to me because no rows change in a SELECT statement.  Instead it should calculate the affected rows only when a non-SELECT statement is executed.  It at least shouldn't mess up the resource pointer.

Here is the offending code:

			if (is_numeric(stripos($sql, 'SELECT')))
			{
				$this->affect_rows = count($result_id->fetchAll());
			}
			else
			{
				$this->affect_rows = $result_id->rowCount();
			}

narfbg:
Duplicate of #4138.

PR #4504, opened 2016-03-04
Fix for #4503

Murtnowski:
#4503 notes that results are coming back empty because fetchAll() is being used on SELECT statements and the resource pointer isn't being reset later.

This change simply makes it so SELECT statements aren't counted as affected rows.


narfbg:
CI 2.x is no longer supported as of October 31st, 2015 and we no longer accept patches for it.

Also, please don't open an issue just to immediately reference it with a pull request; the PR itself is enough.

PR #4515, opened 2016-03-10
Changes in Memcached and Redis classes (drivers of the Cache library)

masterklavi:
1. Changed the mechanism of detecting the serialized values  
    The old mechanism does unnecessary requests to Redis server, and it may have some mistakes in unserializing of value because it doesn't check the type of value that can be different. For example, in the asynchronous requests to Redis with different types of data. The best way is determine the type of value (serialized or not) at the get method.
2. Added the destructor (disconnect from memcache(d))  
    Closing the all connections to cache server is a good practice
3. UNIX socket for connection  
    Many of people use the unix socket for connecting to Memcached because it is the fastest method for data exchange between application and the memcached server.
4. Docblock corrections  
    Some comments don't contain the variable name

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -359,7 +359,7 @@ if ( ! is_php('5.4'))
 	 *
 	 * Returns current CI instance object
 	 *
-	 * @return object
+	 * @return CI_Controller
 	 */
 	function &get_instance()
 	{


system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -60,11 +60,13 @@ class CI_Cache_memcached extends CI_Driver {
...
-	protected $_memcache_conf = array(
+	protected $_config = array(
 		'default' => array(
-			'host'		=> '127.0.0.1',
-			'port'		=> 11211,
-			'weight'	=> 1
+			'socket_type'	=> 'tcp',
+			'host'			=> '127.0.0.1',
+			'port'			=> 11211,
+			'socket'		=> '/var/run/memcached/memcached.sock',
```
narfbg: Why is this the default?

Also, if you'll be changing the *alignment*, use spaces for it.

narfbg: And on another thought, why do we need a separate 'socket' element instead of just using 'host'?
We also don't need 'socket_type' ... it's very easy to determine whether the 'host' value is a hostname or a path on the filesystem.

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -104,24 +97,35 @@ public function __construct()
 		{
 			$this->_memcached = new Memcache();
 		}
-		else
-		{
-			log_message('error', 'Cache: Failed to create Memcache(d) object; extension not loaded?');
```
narfbg: Why is this removed?

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -104,24 +97,35 @@ public function __construct()
...
+			}
+			
+			if ($this->_memcached instanceof Memcache)
 			{
-				// Third parameter is persistance and defaults to TRUE.
 				$this->_memcached->addServer(
 					$cache_server['hostname'],
 					$cache_server['port'],
-					TRUE,
+					TRUE,						// persistance
```
narfbg: I don't quite like that you're just moving comments around in the first place - we don't do changes for the sake of change - but if you do, prepend with a single space, not multiple tabs.

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -169,16 +173,14 @@ public function save($id, $data, $ttl = 60, $raw = FALSE)
 			$data = array($data, time(), $ttl);
 		}
 
-		if (get_class($this->_memcached) === 'Memcached')
+		if ($this->_memcached instanceof Memcache)
 		{
-			return $this->_memcached->set($id, $data, $ttl);
+			return $this->_memcached->set($id, $data, 0, $ttl);
```
narfbg: You wouldn't need to do this if you didn't invert the condition sequence ...

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -286,4 +288,28 @@ public function is_supported()
...
+	/**
+	 * Class destructor
+	 *
+	 * Closes the connection to Memcached if present.
+	 *
+	 * @return	void
+	 */
+	public function __destruct()
+	{
+		if ($this->_memcached)
```
narfbg: We don't really need this ... `$this->_memcached instanceof Memcache` won't trigger any errors, though you will have to do an `elseif ($this->_memcached instanceof Memcached)`.

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -55,11 +55,12 @@ class CI_Cache_redis extends CI_Driver
...
 	 */
 	protected static $_default_config = array(
-		'socket_type' => 'tcp',
-		'host' => '127.0.0.1',
-		'password' => NULL,
-		'port' => 6379,
-		'timeout' => 0
+		'socket_type'	=> 'tcp',
+		'host'			=> '127.0.0.1',
+		'socket'		=> '/var/run/redis/redis.sock',
```
narfbg: Why are you introducing this and changing all the alignments?

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -138,18 +130,24 @@ public function __construct()
 	/**
 	 * Get cache
 	 *
-	 * @param	string	Cache ID
+	 * @param	string	$id	Cache ID
 	 * @return	mixed
 	 */
-	public function get($key)
+	public function get($id)
```
narfbg: Do you really need to do this?

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -138,18 +130,24 @@ public function __construct()
...
 	{
-		$value = $this->_redis->get($key);
-
-		if ($value !== FALSE && isset($this->_serialized[$key]))
+		$value = $this->_redis->get($id);
+		
+		$unserialized = @unserialize($value);
```
narfbg: No! We don't do blind `unserialize()` calls.

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -138,18 +130,24 @@ public function __construct()
...
+		if ($unserialized !== FALSE)
 		{
-			return unserialize($value);
+				return $unserialized;
```
narfbg: Why change indentation?

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -138,18 +130,24 @@ public function __construct()
...
 		}
 
+		if ($value === serialize(FALSE))
+		{
+				return FALSE;
```
narfbg: This too is one level too far in indentation.

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -166,21 +164,10 @@ public function get($key)
 	 */
 	public function save($id, $data, $ttl = 60, $raw = FALSE)
 	{
-		if (is_array($data) OR is_object($data))
+		if ( ! is_string($data))
```
narfbg: What if it's an integer, float, boolean or null?

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -190,23 +177,12 @@ public function save($id, $data, $ttl = 60, $raw = FALSE)
...
-		}
-
-		if (isset($this->_serialized[$key]))
-		{
-			$this->_serialized[$key] = NULL;
-			$this->_redis->sRemove('_ci_redis_serialized', $key);
-		}
-
-		return TRUE;
+		return $this->_redis->delete($id) === 1;
```
narfbg: We always put parentheses around such conditions.

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -271,17 +247,17 @@ public function cache_info($type = NULL)
 	/**
 	 * Get cache metadata
 	 *
-	 * @param	string	Cache key
+	 * @param	string	$id	Cache ID
 	 * @return	array
 	 */
-	public function get_metadata($key)
+	public function get_metadata($id)
```
narfbg: Again a completely unnecessary change.

narfbg:
Please don't submit such large changesets at once.

As it stands - I'd merge the Memcache(d) modifications after the patch is refactored to meed our standards, but I can't because you're submitting it together with Redis changes that I don't want.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: > And on another thought, why do we need a separate 'socket' element instead of just using 'host'?

Because it is more correct. UNIX socket is not host+port. The socket parameter is also specified in Redis (driver in cache library).

> it's very easy to determine whether the 'host' value is a hostname or a path on the filesystem.

I think it is a bad idea. Host can't be a _file path_, it can be a _hostname_ only.

> Also, if you'll be changing the alignment, use spaces for it.

Ok, I'll fix them.

> Why is this the default?

Unfortunately there is no exact default value for the socket so we can use an empty string.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: Because we have the "is_supported()" method

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: > if you do, prepend with a single space, not multiple tabs.

Ok, I got it

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: I think it is not a good practice to use different conditions for similar things so I invert them.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: > We don't really need this ... $this->_memcached instanceof Memcache won't trigger any errors, 

I don't understand what errors was it trigger? In version 5.2 and more it won't trigger any errors.

> though you will have to do an elseif ($this->_memcached instanceof Memcached).

I think it doesn't need. The "_memcached" property contains object of Memcache or Memcached _only_.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: 1. There was no the socket parameter so `$success = $this->_redis->connect($config['socket']);` will trigger an error.
2. The default value for UNIX socket is `/var/run/redis/redis.sock` (it may be `/tmp/redis.sock` also).

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Yes. Why does it used different names of one entity?

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Look at the code again. I didn't do blind `unserialize()` calls. I check it efficiently. There are no any errors. Rather the old mechanism has some mistakes.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: If it is an integer, float, boolean or null we store data as serialized. Next if we'll get this data the `get()` method will return the correct value (not a string, because the string is not a correct value for integer, float, boolean and null).

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok, I got it

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Unnecessary? Why does it used different names of one entity?

system/libraries/Cache/drivers/Cache_memcached.php
narfbg: >> And on another thought, why do we need a separate 'socket' element instead of just using 'host'?
>
> Because it is more correct. UNIX socket is not host+port. The socket parameter is also specified in Redis (driver in cache library).

There's no such thing as "more correct" in this situation ... we just don't need 2 variables if we could do just fine with only one. Especially given that the Memcache(d) classes too use the same parameter to accept either a hostname or a unix socket path.

>> it's very easy to determine whether the 'host' value is a hostname or a path on the filesystem.
>
> I think it is a bad idea. Host can't be a file path, it can be a hostname only.

Read above.

>> Why is this the default?
>
> Unfortunately there is no exact default value for the socket so we can use an empty string.

`NULL` instead of an empty string, actually. But that's not the point ... that the underlying methods don't have defaults doesn't mean that we should invent them.
Erroring because of a non-existing value is way better (read "user-friendlier") than erroring because of a default value that the user didn't want to use.

system/libraries/Cache/drivers/Cache_memcached.php
narfbg: `is_supported()` neither logs this condition, nor is it supposed to.
This log message isn't here by accident.

system/libraries/Cache/drivers/Cache_memcached.php
narfbg: Your reply here doesn't make any sense ... the order inversion provides absolutely nothing in any way imaginable.

system/libraries/Cache/drivers/Cache_memcached.php
narfbg: >> We don't really need this ... $this->_memcached instanceof Memcache won't trigger any errors,
>
> I don't understand what errors was it trigger? In version 5.2 and more it won't trigger any errors.

That's what I'm saying - you're not avoiding any errors with this `if` condition and therefore it is unnecessary.

... and that gives context to the other part of my comment:

>> though you will have to do an elseif ($this->_memcached instanceof Memcached).
>
> I think it doesn't need. The $this->_memcached property contains object of Memcache or Memcached only.

You'll need it when `if ($this->_memcached)` is removed.

system/libraries/Cache/drivers/Cache_redis.php
narfbg: OK, we should fix the non-existing 'socket' problem, but my opinion about socket path defaults is the same here as for Memcache(d).

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: Well, then it is needed to _remove_ the socket parameter from Redis driver (cache library).

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: Ok, I got It. I'll return `log_message()`.

system/libraries/Cache/drivers/Cache_redis.php
narfbg: Somebody wrote it that way. The name is completely irrelevant, wouldn't matter if it was called `$fubar`.

Sure, it may be inconsistent compared to other classes, but that doesn't make it incorrect, not does it make this change necessary.

If you submitted a separate pull request targeted *only* at cleaning up such inconsistencies, I'd probably merge it. But this patch is already quite large and hard to review enough, we don't need stylistic chances on top of it to make the diff even longer and harder to review.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: I don't deny this. But it will be great if we will use a nice code.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: Ok

system/libraries/Cache/drivers/Cache_redis.php
narfbg: I'm looking at it and you *are* calling it blindly *at this point* right here and there's nothing efficient about it.
The check that follows it is a band-aid, not a solution - you need to *know* that the value was serialized before calling `unserialize()` on it.

Also, there are errors. You're just suppressing them - we don't do that unless there's absolutely no other way to make something work and that's not the case here.

If you think that the current mechanism has "some mistakes" - please explain them first.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok, we can remove the `socket` and `socket_type` parameters.

system/libraries/Cache/drivers/Cache_redis.php
narfbg: That's the thing I don't like about it - we *don't* need to serialize scalar types, only arrays and objects.

system/libraries/Cache/drivers/Cache_redis.php
narfbg: Same as here: https://github.com/bcit-ci/CodeIgniter/pull/4515#discussion_r55832527

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok, I'll do it in the next pill request.

system/libraries/Cache/drivers/Cache_memcached.php
narfbg: While I agree that it should be removed from Redis, it's not necessarily "needed" and that should happen in a *separate* pull request.

Nothing will break if this driver is patched and the other one is not. You're just making the changes harder to reason about by mixing so much things at once.

You can't "fix" everything at once ... bug fixes, new features, stylistic changes don't belong together, even less so on separate pieces of code that are completely independent from each other.

Please refer to what I said here: https://github.com/bcit-ci/CodeIgniter/pull/4515#issuecomment-195288759

system/libraries/Cache/drivers/Cache_memcached.php
narfbg: How is checking for Memcache before Memcache**d** make this "nice"?

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: > The check that follows it is a band-aid, not a solution - you need to know that the value was serialized before calling unserialize() on it.

It isn't a band-aid and it isn't my idea. Look at http://php.net/manual/en/function.unserialize.php#85097

> If you think that the current mechanism has "some mistakes" - please explain them first.

I wrote them in my first message:

> The old mechanism does unnecessary requests to Redis server, and it may have some mistakes in unserializing of value because it doesn't check the type of value that can be different. For example, in the asynchronous requests to Redis with different types of data.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok. What would we _get_ if we _set_ a boolean value like `FALSE` or a `NULL` value? We will _get_ the same value `""`. 
Worth mentioning that `"" !== FALSE` and `"" !== NULL`.

system/libraries/Cache/drivers/Cache_redis.php
narfbg: >> The check that follows it is a band-aid, not a solution - you need to know that the value was serialized before calling unserialize() on it.
>
> It isn't a band-aid and it isn't my idea. Look at http://php.net/manual/en/function.unserialize.php#85097

It is a band-aid. What I mean by "band-aid" is that it is masking a failure (to unserialize) that has already happened.

I don't undestand why someone writing a comment on php.net should make something a better idea, if that's what you're implying.
`serialize()` and `unserialize()` are not as harmless as you may think they are, you shouldn't be using them just to make quick checks (ironically, they are also not quick).

>> If you think that the current mechanism has "some mistakes" - please explain them first.
>
> I wrote them in my first message:
>

See, this is what wouldn't happen if this wasn't one huge patch - we're discussing so much things at once that I've completely lost your initial message. :)

>> The old mechanism does unnecessary requests to Redis server, and it may have some mistakes in unserializing of value because it doesn't check the type of value that can be different. For example, in the asynchronous requests to Redis with different types of data.

I disagree with the first part of this and don't really understand what you mean with the second part. Fair enough - you may be trying to solve a problem that I don't see - but please do it in a separate pull request where we can focus on it alone.

masterklavi:
> Please don't submit such large changesets at once.
> As it stands - I'd merge the Memcache(d) modifications after the patch is refactored to meed our standards, but I can't because you're submitting it together with Redis changes that I don't want.

Ok

system/libraries/Cache/drivers/Cache_redis.php
narfbg: Would we? I'm quite certain that Redis stores NULLs as NULLs and booleans as booleans.

masterklavi: Yeah, sorry, I'm confused with memcache.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok, I got it.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: Ok

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: > It is a band-aid. What I mean by "band-aid" is that it is masking a failure (to unserialize) that has already happened.

Ok, then I agree. I'll do another solution in a separate pull request.

> I don't undestand why someone writing a comment on php.net should make something a better idea, if that's what you're implying.

Ok.

> serialize() and unserialize() are not as harmless as you may think they are, you shouldn't be using them just to make quick checks (ironically, they are also not quick).

Yes, but it faster than an additional request to Redis server `$serialized = $this->_redis->sMembers('_ci_redis_serialized')`.

> See, this is what wouldn't happen if this wasn't one huge patch - we're discussing so much things at once that I've completely lost your initial message. :)

Yeah, I'm sorry. It is my first pull request.

> I disagree with the first part of this and don't really understand what you mean with the second part. Fair enough - you may be trying to solve a problem that I don't see - but please do it in a separate pull request where we can focus on it alone.

Ok.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: Ok.

system/libraries/Cache/drivers/Cache_memcached.php
masterklavi: It is nice if we use `if ($this->_memcached instanceof Memcache) ... else ...` in `_construct()`, `save()`, and etc.
It isn't nice if we use `if ($this->_memcached instanceof Memcache) ... else ...` in `_construct()`, and  `if ($this->_memcached instanceof MemcacheD) ... else ...` in `save()`.

PR #4521, opened 2016-03-12
Added variables to Docblocks where they are skipped

masterklavi:


system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -255,7 +255,7 @@ public function clean()
 	/**
 	 * Get cache driver info
 	 *
-	 * @param	string	Not supported in Redis.
+	 * @param	string	$type	Not supported in Redis.
 	 *			Only included in order to offer a
```
narfbg: This and the following line need to be further indented, so that they are aligned with the start of the description.

masterklavi: > This and the following line need to be further indented, so that they are aligned with the start of the description.
Ok, fixed.

PR #4522, opened 2016-03-12
Added the destructor (disconnect from memcache(d))

masterklavi:
`Memcache` and `Memcached` have different methods to closing connection.


janul:
just to be aware that destructor is not working in php 5.3.9 ("quit" method is not defined in Memcached class). I propose to change php ver 5.4 from recommended to required in documentation

narfbg:
`Memcached::quit()` requires version 2.0.0 of the PECL memcached extension, not a PHP version.

janul:
sorry, you're right. I guess that most distributions which have php < 5.4 have php5-memcached < v2 - but it's just guess.

PR #4524, opened 2016-03-12
Memcached refactoring

masterklavi:


system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -81,19 +81,11 @@ public function __construct()
 	{
 		// Try to load memcached server info from the config file.
 		$CI =& get_instance();
-		$defaults = $this->_memcache_conf['default'];
+		$defaults = $this->_config['default'];
 
-		if ($CI->config->load('memcached', TRUE, TRUE))
+		if ($CI->config->load('memcached', TRUE, TRUE) && is_array($CI->config->config['memcached']))
```
narfbg: We don't need the `is_array()` check - if `CI_Config::load()` returns TRUE, then we're guaranteed to have an array.

masterklavi: Ok, fixed.

PR #4526, opened 2016-03-12
Code refactoring (redis). The "$config" variable is moved to "if" condition.

masterklavi:


narfbg:
There's a conflict between your branch and develop, preventing this from being merged.

On another note, please use the description field instead of using the title for everything. Here would've been a good opportunity to explain what the actual benefit from the change is - it's not obvious by looking at the patch.

masterklavi:
> There's a conflict between your branch and develop, preventing this from being merged.

Ok, fixed.

> On another note, please use the description field instead of using the title for everything. Here would've been a good opportunity to explain what the actual benefit from the change is - it's not obvious by looking at the patch.

Ok

PR #4528, opened 2016-03-12
Setup the data defining for stored data (redis)

masterklavi:
It is an improved version.
See previous at #4527 .

By the way, @narfbg  **you are wrong!**
> > Ok. What would we get if we set a boolean value like FALSE or a NULL value? We will get the same value "". 
> > Worth mentioning that "" !== FALSE and "" !== NULL.

> Would we? I'm quite certain that Redis stores NULLs as NULLs and booleans as booleans.

Redis stores NULLs as NULLs and booleans as booleans but PHPRedis returns they as `string`. Check it yourself.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -143,19 +143,20 @@ class CI_Cache_redis extends CI_Driver
 	 */
 	public function get($key)
 	{
-		$value = $this->_redis->hMGet($key, array('serialized', 'data'));
+		$value = $this->_redis->hMGet($key, array('type', 'data'));
 
-		if ( ! is_array($value) OR  ! isset($value['serialized'], $value['data']))
+		if ( ! is_array($value) OR  ! isset($value['type'], $value['data']))
 		{
 			return FALSE;
 		}
 
-		if ($value['serialized'])
+		if ($value['type'] === 'array' OR $value['type'] === 'object')
 		{
 			return unserialize($value['data']);
 		}
 		else
 		{
+			settype($value['data'], $value['type']);
 			return $value['data'];
 		}
 	}
@@ -175,11 +176,11 @@ class CI_Cache_redis extends CI_Driver
 	{
 		if (is_array($data) OR is_object($data))
 		{
-			return $this->_redis->hMSet($id, array('serialized' => TRUE, 'data' => serialize($data)));
+			return $this->_redis->hMSet($id, array('type' => gettype($data), 'data' => serialize($data)));
 		}
 		else
 		{
-			return $this->_redis->hMSet($id, array('serialized' => FALSE, 'data' => $data));
+			return $this->_redis->hMSet($id, array('type' => gettype($data), 'data' => $data));
 		}
 	}
 
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -143,14 +143,21 @@ class CI_Cache_redis extends CI_Driver
 	 */
 	public function get($key)
 	{
-		$value = $this->_redis->get($key);
+		$value = $this->_redis->hMGet($key, array('serialized', 'data'));
 
-		if ($value !== FALSE && isset($this->_serialized[$key]))
+		if ( ! is_array($value))
 		{
-			return unserialize($value);
+			return FALSE;
 		}
 
-		return $value;
+		if ( ! empty($value['serialized']))
+		{
+			return unserialize($value['data']);
+		}
+		else
+		{
+			return $value['data'];
+		}
 	}
 
 	// ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


narfbg:
Bolding "you are wrong!" in your PR description? Some way to establish a good communication tone ...

I'm not wrong - you just confirmed what I've stated (how Redis stores data), and answered what I *asked* (what we would receive).
And I asked that because I don't have a Redis instance available, nor phpRedis installed, so I can't really test. I can't possibly have access to all the technologies that CI supports.

------------------------

If this is an improved version of #4527, then why didn't you just push your changes there instead of opening a new PR and why is it still open? We can't have both ...

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -143,14 +143,22 @@ public function __construct()
 	 */
 	public function get($key)
 	{
-		$value = $this->_redis->get($key);
+		$value = $this->_redis->hMGet($key, array('type', 'data'));
 
-		if ($value !== FALSE && isset($this->_serialized[$key]))
+		if ( ! is_array($value) OR  ! isset($value['type'], $value['data']))
```
narfbg: Can `hMGet()` return anything else but an array or FALSE?
Unless that's the case, `$value === FALSE` would be better than `! is_array($value)`.

masterklavi:
> Bolding "you are wrong!" in your PR description? Some way to establish a good communication tone ...

I'm sorry.

> If this is an improved version of #4527, then why didn't you just push your changes there instead of opening a new PR and why is it still open? We can't have both ...

Ok, I have closed it.

system/libraries/Cache/drivers/Cache_redis.php
masterklavi: > Can hMGet() return anything else but an array or FALSE?

Good note. There are three possible values:
1. `array` like `array('type' => 'integer', 'value' => 123)` if key exists and contains a hashtable
2. `array` like `array('type' => false, 'value' => false)` if key doesn't exist
3. `FALSE` if key exists but contain a simple string or other type
Fixed in new commit.

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -143,14 +143,22 @@ public function __construct()
 	 */
 	public function get($key)
 	{
-		$value = $this->_redis->get($key);
+		$value = $this->_redis->hMGet($key, array('type', 'data'));
 
-		if ($value !== FALSE && isset($this->_serialized[$key]))
+		if ($value === FALSE OR  $value['type'] === FALSE)
```
narfbg: The `$value['type'] === FALSE` condition used to be `! isset($value['type'], $value['data'])` in your previous commit ...
Checking if the key exists or if it's FALSE - wWhich one is correct? If it's the latter, then we should be checking `$value['data']` for FALSE too.

Also, you have an extra space character after the `OR`.

masterklavi: > Checking if the key exists or if it's FALSE - wWhich one is correct?

Both are correct but we can use `if (empty($value['type']))` only. See my next commit.

> If it's the latter, then we should be checking $value['data'] for FALSE too.

No, we shouldn't because there are three variants only and `if (empty($value['type']))` is worked for all of them. If `$value['data']` doesn't exist then `$value['type']` doesn't exist too.

narfbg: They can't be both correct ... if `$value['type']` is FALSE, `isset($value['type'])` will return TRUE.

And while in theory `$value['data']` SHOULD exist if `$value['type']` does, that depends on the assumption that we control all of Redis, which can never be guaranteed.
In fact, we're missing other checks related to this, but I'll add those later ... `empty($value['type'])` is good enough for your patch.

masterklavi: > > > Checking if the key exists or if it's FALSE - wWhich one is correct?

> > Both are correct but we can use if (empty($value['type'])) only. See my next commit.

> They can't be both correct ... if $value['type'] is FALSE, isset($value['type']) will return TRUE.

I mean if `$value === FALSE` then `isset($value['type'])` will return `FALSE`.

narfbg:
Still, in relation to [this inline comment](https://github.com/bcit-ci/CodeIgniter/pull/4528#discussion_r55920377) - what would be the `$value['type']` value if it doesn't exist in Redis?

masterklavi:
If key doesn't exist, `hMGet()` will return `array('type' => false, 'value' => false)`.
If key exists but the value has another type (like a simple string), `hMGet()` will return `FALSE`.
If key exists and the value contains a hash table without `type` index, `hMGet()` will return `array('type' => false, 'value' => ..., 'other_index' => ...)`.

narfbg:
OK, thanks.

One last question - does `Redis::ttl()` work on hashes? `get_metadata()` uses it ...

narfbg:
And now that I said that, I also notice that `hMSet()` doesn't have a `$ttl` parameter ...

This is bad, unless you come up with a quick solution to this, I'll have to revert your patch.

masterklavi:
One minute. I have a quick solution.

masterklavi:
Fixed in #4530

narfbg:
I've made some changes to your patch (see reference above this comment), adding some sanity checks, renaming some keys/variables and cleaning it up a bit.

Also this: 16436e3c037ba4688954cffab2316d44115b2eff

Thanks for your contributions!

masterklavi:
Ok, great!

PR #4531, opened 2016-03-12
Removed the second condition for unix sockets (redis)

masterklavi:
Removed the second condition of connectivity for unix sockets in `Redis::__construct()` method.
It is unnecessary because we can use `host` and `port` parameters.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -55,7 +55,6 @@ class CI_Cache_redis extends CI_Driver
 	 * @var	array
 	 */
 	protected static $_default_config = array(
-		'socket_type' => 'tcp',
 		'host' => '127.0.0.1',
 		'password' => NULL,
 		'port' => 6379,
@@ -105,16 +104,7 @@ class CI_Cache_redis extends CI_Driver
 
 		try
 		{
-			if ($config['socket_type'] === 'unix')
-			{
-				$success = $this->_redis->connect($config['socket']);
-			}
-			else // tcp socket
-			{
-				$success = $this->_redis->connect($config['host'], $config['port'], $config['timeout']);
-			}
-
-			if ( ! $success)
+			if ( ! $this->_redis->connect($config['host'], $config['port'], $config['timeout']))
 			{
 				log_message('error', 'Cache: Redis connection failed. Check your configuration.');
 			}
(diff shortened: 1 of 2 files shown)


masterklavi:
If you would approve it, I will remove the #4520 .

narfbg:
We still need different connect calls depending on the socket type ... you have to check if the first character of `$config['host']` is a slash for that.
And 'socket_type' needs to be removed from the documentation as well.

Also, please don't open 2 pull requests and wait to see which one (if any) is accepted. Pick only one route ... I'm closing #4520.

masterklavi:
> you have to check if the first character of $config['host'] is a slash for that.

I think it's a bad idea. Why do we need to check it?

> And 'socket_type' needs to be removed from the documentation as well.

Ok, removed.

> Also, please don't open 2 pull requests and wait to see which one (if any) is accepted. Pick only one route 

Ok.

narfbg:
What do you mean why? You shouldn't pass port and timeout values for a unix socket.

masterklavi:
> What do you mean why? You shouldn't pass port and timeout values for a unix socket.

Ok. We have only one restriction for using it: the port must to be zero. See my commit.

narfbg:
Why do you insist on using the timeout value? The phpredis manual shows no such usage for unix sockets.

Also, you shouldn't be using curly braces - i.e. `$config['host'][0]` instead of `$config['host']{0}`.

masterklavi:
> Why do you insist on using the timeout value? The phpredis manual shows no such usage for unix sockets.

Yes, but the manual doesn't deny the timeout value. There is only one restriction for the timeout value: it must be greater than (or equal) to zero.

masterklavi:
From https://github.com/phpredis/phpredis 
> Also more than one persistent connection can be made identified by either host + port + timeout or host + persistent_id or unix socket + timeout.

Maybe it would be better if we will save the timeout value for unix socket?

narfbg:
I don't get it ... you're still insisting on (and giving arguments in favor of) using the timeout for UNIX sockets, yet you've set it to 0 in the latest commits. Do you stand by your suggestions or not?

Also, I don't care about the timeout itself ... I want to avoid the unnecessary variable assignments. That is, you could just do this:

    $this->_redis->connect($config['host'], ($config['host'][0] === '/' ? 0 : $config['port']), $config['timeout'])

It's less code and we're both happy.

masterklavi:
> It's less code and we're both happy.
Ok.

PR #4534, opened 2016-03-14
3.0 stable

hammad-mustafa:


narfbg:
I do these merges after releases and/or when necessary for immediate patches.

Does an impending PR from you depend on this? If not, while there's no technical reason not to merge it, anybody could do this for every tiny commit we push to 3.0-stable and I don't want to encourage that.

Issue #3649, opened 2015-03-06
CI_Session and session ID validation


jmccauli:
#3073 added an extra bit to session cookie validation. Namely, it added making sure the session cookie's  value matches the following regex: /^[0-9a-f]{40}$/. While I believe this was a good addition, I believe it needs to be configurable.

Case in point:

I'm currently trying to, very slowly, move an old system to Codeigniter. Unfortunately, that system is using  a session hash function and session hash_bits_per_character that is incompatible with the regex pattern referenced above.

I believe the regex pattern above should be able to be configurable OR ignored. The change I'm referring to was done to add additional security to session IDs in the session cookie. However, it forces a server administrator to use SHA-1 and the fewest hash bits per character. This should be left up to the developers and server administrators, not the framework in my opinion.

Additionally, there are several calls to ini_set at the bottom of CI_Session::_configure that again force the above on developers and server administrators. I believe these should be configurable as well.

vahidov:
Vote for this issue.

In php 7.1 are new session configuration values:
session.sid_length
session.sid_bits_per_character
http://php.net/manual/en/session.configuration.php#ini.session.sid-length

and by default values  regex: /^[0-9a-f]{40}$/ is not working.

narfbg:
md5 as the hash_function, or sid_length < 40 will remain disallowed.
But we do allow stronger INI values now.

narfbg:
2f760877c313871e5066b93b0b1aa76428c09fb6

ahmed00medhat:
still i don't understand how to fix this problem. can you please provide some examples? i found that the cookie value is 32 characters. i tried everything and still same issue. i tried to use encryption and not to encrypt. i tried files and database drivers. i tried setting the expiry time to 0 and update time to 0. 
my site is hosted as subdomain. whatever the value i set for the coockie_domain, i find 3 cookies created in the session . one for .subdomain.maindomain.com and subdomain.maindomain.com and .maindomain.com. i'm not sure why.

i'm using nginx webserver, php 7.1 and php-pfm, and CentOS 6.
on localhost i don't face this issue.

i spent all the day and couldn't reach anything. please help.

narfbg:
@ahmed00medhat This is a bug tracker, please post your questions on our [forums](https://forum.codeigniter.com/) instead.

It also sounds a lot like you're talking about CI2, which was abandoned a long time ago and has nothing to do with what we're talking about here.

technoknol:
[...]

PR #4542, opened 2016-03-17
Feature request #3649 fix

pacobernabe:
I have prepared a solution for the feature requested in https://github.com/bcit-ci/CodeIgniter/issues/3649.
It includes two new configuration variables for the Sessions, allowing CodeIgniter ci_session value to be coherent with whatever values we have set in our PHP.ini configuration:
- session.hash_function
- session.hash_bits_per_character

Also this change is backward compatible as default values in config.php are the same as the hard coded values in the existing Session.php.

It does not interfere with ini_set values set in index.php around line 285.

For testing it I created a test controller: https://github.com/bcit-ci/CodeIgniter/commit/02fe02d584f98761ff5a0665a13f5bd799c5cfdb

Please consider if is meaningful for the project.

Thank you,

Paco Bernabe


narfbg:
Is it meaningful? Sure.

But is it what we want? I can't confirm at this time.

narfbg:
Already fixed; read through #3649.

PR #4543, opened 2016-03-17
Database throws exception on error

rodolfodn:
Option to throw exception insted print the error


narfbg:
In addition to #4189, `display_error()` is only used when 'db_debug' is enabled.

PR #4568, opened 2016-04-05
more correct validation of session id

fiblan:
Signed-off-by: Fabio Sfuncia <fiblan@tooware.com>

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -130,7 +130,7 @@ class CI_Session {
 		if (isset($_COOKIE[$this->_config['cookie_name']])
 			&& (
 				! is_string($_COOKIE[$this->_config['cookie_name']])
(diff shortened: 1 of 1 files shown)


fiblan:
View internal php validation:
https://github.com/php/php-src/blob/master/ext/session/session.c#L456

narfbg:
How's that more correct?

fiblan:
Actual regexp
```
'/^[0-9a-f]{40}$/'
``` 
validate only a possible format, 
the proposed:
```
'/^[-,a-zA-Z0-9]{1,128}$/'
```
is in accordance with https://github.com/php/php-src/blob/master/ext/session/session.c#L456 
and considering the configuration ```session.hash_bits_per_character=6```
http://php.net/manual/en/session.configuration.php#ini.session.hash-bits-per-character

narfbg:
https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Session/Session.php#L317

PR #4585, opened 2016-04-17
Fixes #4583 by separating attachments to related and mixed multiparts

### LINKED ISSUE #4583: Attachments are not displayed on iOS stock Email client
### The issue happens in the following scenario

1. Construct an HTML email
2. Attach image(s) to the email
3. Make image(s) attached in the previous step to be shown as the part of the email's body (for example a logo on the top)
4. Add other attachments, which are not intended to be shown as a part of the email's body. (for example report in pdf format)

### The Result

Inline images are displayed correctly, but attachments added in step#4 are hidden in iOS's email client.

### The Problem with the current implementation

To make an attachment to be shown inline in the email body, the following code is needed:

```php
$filename = 'images/inline_image.png';
$this->email->attach($filename);
$cid = $this->email->attachment_cid($filename);
// $cid is used in the html
// as the value of an src attribute of an <image> tag for example
```

It goes wrong when `attachment_cid` function is called, because it sets the root multipart to related.
The structure of the message will be the following:

```
multipart/related
  inline_image.png
  report1.pdf
  report2.pdf
  ...
```

According to https://tools.ietf.org/html/rfc2387 the `multipart/related` is 

> used to indicate that each message part is a component of an aggregate whole. It is for compound objects consisting of several inter-related components - **proper display cannot be achieved by individually displaying the constituent parts**.

 (source: https://en.wikipedia.org/wiki/MIME)

As all of the attachments will be the direct child of `multipart/related`, even the "normal" attachments won't be displayed by the stricter email clients (for example iOS's stock email client).

r4ngy4l:
The issue has been fixed by separating attachments into 2 different multiparts for html emails

- Embedded attachments are generated into a multipart/related part
- Non-embedded attachments are generated into a multipart/mixed part

Please note that this solution also works for Thunderbird.

### DIFF
--- a/system/libraries/Email.php
+++ b/system/libraries/Email.php
@@ -150,13 +150,6 @@ class CI_Email {
 	public $charset		= 'utf-8';
 
 	/**
-	 * Multipart message
-	 *
-	 * @var	string	'mixed' (in the body) or 'related' (separate)
-	 */
-	public $multipart	= 'mixed';		// "mixed" (in the body) or "related" (separate)
-
-	/**
 	 * Alternative message (for HTML messages only)
 	 *
 	 * @var	string
@@ -273,6 +266,13 @@ class CI_Email {
 	 * @var	string
 	 */
 	protected $_atc_boundary	= '';
+
+	/**
+	 * Related Attachment boundary
+	 *
+	 * @var	string
+	 */
+	protected $_rel_boundary	= '';
 
 	/**
 	 * Final headers to send
@@ -766,7 +766,8 @@ class CI_Email {
 			'name'		=> array($file, $newname),
 			'disposition'	=> empty($disposition) ? 'attachment' : $disposition,  // Can also be 'inline'  Not sure if it matters
 			'type'		=> $mime,
-			'content'	=> chunk_split(base64_encode($file_content))
+			'content'	=> chunk_split(base64_encode($file_content)),
+			'multipart' => 'mixed'
 		);
 
 		return $this;
@@ -784,15 +785,11 @@ class CI_Email {
 	 */
 	public function attachment_cid($filename)
 	{
-		if ($this->multipart !== 'related')
-		{
-			$this->multipart = 'related'; // Thunderbird need this for inline images
-		}
-
 		for ($i = 0, $c = count($this->_attachments); $i < $c; $i++)
 		{
 			if ($this->_attachments[$i]['name'][0] === $filename)
 			{
+				$this->_attachments[$i]['multipart'] = 'related';
 				$this->_attachments[$i]['cid'] = uniqid(basename($this->_attachments[$i]['name'][0]).'@');
 				return $this->_attachments[$i]['cid'];
 			}
@@ -945,6 +942,7 @@ class CI_Email {
 	{
 		$this->_alt_boundary = 'B_ALT_'.uniqid(''); // multipart/alternative
 		$this->_atc_boundary = 'B_ATC_'.uniqid(''); // attachment boundary
+		$this->_rel_boundary = 'B_REL_'.uniqid(''); // related attachment boundary
 	}
 
 	// --------------------------------------------------------------------
@@ -1391,7 +1389,7 @@ class CI_Email {
 
 			case 'plain-attach' :
 
-				$hdr .= 'Content-Type: multipart/'.$this->multipart.'; boundary="'.$this->_atc_boundary.'"';
+				$hdr .= 'Content-Type: multipart/mixed; boundary="'.$this->_atc_boundary.'"';
 
 				if ($this->_get_protocol() === 'mail')
 				{
@@ -1406,18 +1404,45 @@ class CI_Email {
 					.$this->newline
 					.$this->_body.$this->newline.$this->newline;
 
+				$attachment_prepared = $this->_prep_attachments($this->_attachments, $this->_atc_boundary);
+
 			break;
 			case 'html-attach' :
 
-				$hdr .= 'Content-Type: multipart/'.$this->multipart.'; boundary="'.$this->_atc_boundary.'"';
+				$attachments_indexed_by_multipart = $this->_attachments_indexed_by_multipart();
+				$attachments_related = $attachments_indexed_by_multipart['related'];
+				$attachments_mixed = $attachments_indexed_by_multipart['mixed'];
+				$prepared_attachment_parts = array();
+
+				if (isset($attachments_mixed) && count($attachments_mixed) > 0)
+				{
+					$hdr .= 'Content-Type: multipart/mixed; boundary="'.$this->_atc_boundary.'"';
+
+					array_push($prepared_attachment_parts, $this->_prep_attachments($attachments_mixed, $this->_atc_boundary));
+					$last_boundary = $this->_atc_boundary;
+				}
+
+				if (isset($attachments_related) && count($attachments_related) > 0)
+				{
+					$target_container =& $hdr;
+					if (isset($last_boundary)) {
+						$target_container =& $body;
+						$target_container .= '--' . $last_boundary . $this->newline;
+					}
+					$target_container .= 'Content-Type: multipart/related; boundary="'.$this->_rel_boundary.'"';
+
+					array_unshift($prepared_attachment_parts, $this->_prep_attachments($attachments_related, $this->_rel_boundary));
+					$last_boundary = $this->_rel_boundary;
+				}
 
 				if ($this->_get_protocol() === 'mail')
 				{
 					$this->_header_str .= $hdr;
 				}
 
+				if (strlen(body) > 0) $body .= $this->newline.$this->newline;
 				$body .= $this->_get_mime_message().$this->newline.$this->newline
-					.'--'.$this->_atc_boundary.$this->newline
+					.'--'.$last_boundary.$this->newline
 
 					.'Content-Type: multipart/alternative; boundary="'.$this->_alt_boundary.'"'.$this->newline.$this->newline
 					.'--'.$this->_alt_boundary.$this->newline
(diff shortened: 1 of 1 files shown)


narfbg:
Have you tested this with email clients other than whatever the iOS default one is?

narfbg:
This still needs work though ... for starters, it doesn't comply with our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).

Please read it and correct accordingly before I can leave comments for other stuff.

robertonline:
Code style issues are fixed. Please review.

system/libraries/Email.php
```diff
@@ -784,15 +785,11 @@ public function attach($file, $disposition = '', $newname = NULL, $mime = '')
...
-		if ($this->multipart !== 'related')
-		{
-			$this->multipart = 'related'; // Thunderbird need this for inline images
-		}
-
 		for ($i = 0, $c = count($this->_attachments); $i < $c; $i++)
 		{
 			if ($this->_attachments[$i]['name'][0] === $filename)
 			{
+				$this->_attachments[$i]['multipart'] = 'related';
```
narfbg: If this is always 'related', why is it set for every element here?

system/libraries/Email.php
```diff
@@ -1406,18 +1404,50 @@ protected function _build_message()
...
 			case 'html-attach' :
 
-				$hdr .= 'Content-Type: multipart/'.$this->multipart.'; boundary="'.$this->_atc_boundary.'"';
+				$attachments_indexed_by_multipart = $this->_attachments_indexed_by_multipart();
+				$attachments_related = $attachments_indexed_by_multipart['related'];
+				$attachments_mixed = $attachments_indexed_by_multipart['mixed'];
+				$prepared_attachment_parts = array();
+				$last_boundary = NULL;
+
+				if (isset($attachments_mixed) && count($attachments_mixed) > 0)
```
narfbg: just `! empty($attachments_mixed)`

system/libraries/Email.php
```diff
@@ -1406,18 +1404,50 @@ protected function _build_message()
...
+				{
+					$hdr .= 'Content-Type: multipart/mixed; boundary="'.$this->_atc_boundary.'"';
+
+					array_push($prepared_attachment_parts, $this->_prep_attachments($attachments_mixed, $this->_atc_boundary));
```
narfbg: `array_push()` is unnecessary overhead here, just do `$prepared_attachment_parts[] = ...`

system/libraries/Email.php
```diff
@@ -1406,18 +1404,50 @@ protected function _build_message()
...
+					$last_boundary = $this->_atc_boundary;
+				}
+
+				if (isset($attachments_related) && count($attachments_related) > 0)
```
narfbg: Same as on [line 1418](https://github.com/bcit-ci/CodeIgniter/pull/4585/files#r60362682).

system/libraries/Email.php
```diff
@@ -1406,18 +1404,50 @@ protected function _build_message()
...
+				{
+					$target_container =& $hdr;
```
narfbg: Why do you have to use a reference instead of directly appending to `$hdr`?

system/libraries/Email.php
```diff
@@ -1406,18 +1404,50 @@ protected function _build_message()
...
+					array_unshift($prepared_attachment_parts, $this->_prep_attachments($attachments_related, $this->_rel_boundary));
+					$last_boundary = $this->_rel_boundary;
+				}
 
 				if ($this->_get_protocol() === 'mail')
 				{
 					$this->_header_str .= $hdr;
 				}
 
+				if (strlen(body) > 0)
```
narfbg: You're missing the dolar sign here ...
Also, here's a cleaner way to do such short conditionals:

    strlen($body) && $body .= $this->newline.$this->newline;

system/libraries/Email.php
```diff
@@ -1406,18 +1404,50 @@ protected function _build_message()
...
 					.$this->_body.$this->newline.$this->newline;
 
+				$attachment_prepared = $this->_prep_attachments($this->_attachments, $this->_atc_boundary);
+
 			break;
 			case 'html-attach' :
 
-				$hdr .= 'Content-Type: multipart/'.$this->multipart.'; boundary="'.$this->_atc_boundary.'"';
+				$attachments_indexed_by_multipart = $this->_attachments_indexed_by_multipart();
+				$attachments_related = $attachments_indexed_by_multipart['related'];
```
narfbg: Do these really have to be re-assigned to separate variables? Why are you not simply accessing the respective array elements?

system/libraries/Email.php
```diff
@@ -1432,32 +1462,77 @@ protected function _build_message()
...
+	 */
+	protected function _attachments_indexed_by_multipart()
+	{
+		foreach ($this->_attachments as $attachment)
 		{
-			$filename = $this->_attachments[$i]['name'][0];
-			$basename = ($this->_attachments[$i]['name'][1] === NULL)
-				? basename($filename) : $this->_attachments[$i]['name'][1];
+			$multipart = $attachment['multipart'];
+			if (!isset($attachments_indexed[$multipart]))
```
narfbg: Missing spaces around the `!` operator.

system/libraries/Email.php
```diff
@@ -1432,32 +1462,77 @@ protected function _build_message()
...
+			{
+				$attachments_indexed[$multipart] = array();
+			}
+			array_push($attachments_indexed[$multipart], $attachment);
```
narfbg: Again avoid `array_push()` please.

system/libraries/Email.php
```diff
@@ -1432,32 +1462,77 @@ protected function _build_message()
...
+	/**
+	 * Prepares attachment string
+	 *
+	 * @param	array	$attachments
+	 * @param	string	$boundary	Multipart boundary string
+	 * @return	string
+	 */
+	protected function _prep_attachments($attachments, $boundary)
+	{
+		if (!isset($attachments) OR count($attachments) === 0)
```
narfbg: Again missing spaces around `!`, though this should be just `empty($attachments)`.

system/libraries/Email.php
```diff
@@ -1432,32 +1462,77 @@ protected function _build_message()
...
+		{
+			return '';
 		}
 
-		$body .= implode($this->newline, $attachment).$this->newline.'--'.$this->_atc_boundary.'--';
-		$this->_finalbody = ($this->_get_protocol() === 'mail')
-			? $body
-			: $hdr.$this->newline.$this->newline.$body;
+		$attachment_parts = array();
+		foreach ($attachments as $attachment)
```
narfbg: The previous version used a `for` loop very intentionally, for lower memory usage.

By using a `foreach` here, you're copying each element once and by assigning to the newly-created `$attachment_parts`, you're copying them a second time. The end result is a peak memory usage three times higher than what it needs to be.

system/libraries/Email.php
```diff
@@ -1432,32 +1462,77 @@ protected function _build_message()
...
+		{
+			$filename = $attachment['name'][0];
+			$basename = ($attachment['name'][1] === NULL)
+				? basename($filename) : $attachment['name'][1];
 
-		return TRUE;
+			array_push(
```
narfbg: Again, no `array_push()` please. :)

system/libraries/Email.php
robertonline: It is 'mixed' for every element by default. See the related code from 'attach' function:

```php
$this->_attachments[] = array(
    'name'		=> array($file, $newname),
    'disposition'	=> empty($disposition) ? 'attachment' : $disposition,
    'type'		=> $mime,
    'content'	=> chunk_split(base64_encode($file_content)),
    'multipart'	=> 'mixed' // <--- the default value is set HERE
);
```

'related' is not set for every element, only for those, which gets a CID ( these attachments are supposed to be used as embedded in the email's body )

system/libraries/Email.php
```diff
@@ -1497,13 +1502,16 @@ protected function _attachments_indexed_by_multipart()
...
-	 * @param   string  $boundary	  Multipart boundary string
-	 * @return  string
+	 * @param	array	$attachments
+	 * @param	string	$boundary	Multipart boundary string
+	 * @return	string
 	 */
 	protected function _prep_attachments($attachments, $boundary)
 	{
-		if (!isset($attachments) || count($attachments) === 0) return '';
+		if (!isset($attachments) OR count($attachments) === 0)
```
vkeranov: Should be `if ( ! isset($attachments)`... With spaces around "!".

robertonline: This has already been commented by @narfbg

system/libraries/Email.php
robertonline: Fixed

system/libraries/Email.php
robertonline: Refactored.

system/libraries/Email.php
robertonline: Refactored.

system/libraries/Email.php
robertonline: Refactored.

system/libraries/Email.php
robertonline: Refactored.

system/libraries/Email.php
robertonline: Refactored.

system/libraries/Email.php
robertonline: Refactored

system/libraries/Email.php
robertonline: Refactored

system/libraries/Email.php
robertonline: Refactored

system/libraries/Email.php
robertonline: `for` is restored. Thanks for the explanation.

system/libraries/Email.php
robertonline: Refactored

robertonline:
All fixes/suggestions commented are implemented in ed458fb.

system/libraries/Email.php
```diff
@@ -1406,18 +1404,47 @@ protected function _build_message()
...
+					$prepared_attachment_parts[] = $this->_prep_attachments($attachments_indexed_by_multipart['mixed'], $this->_atc_boundary);
+					$last_boundary = $this->_atc_boundary;
+				}
+
+				if ( ! empty($attachments_indexed_by_multipart['related']))
+				{
+					$rel_boundary_header = 'Content-Type: multipart/related; boundary="'.$this->_rel_boundary.'"';
+					if (isset($last_boundary))
+					{
+						$body .= '--' . $last_boundary . $this->newline . $rel_boundary_header;
```
narfbg: We don't put spaces around concatenation.

system/libraries/Email.php
```diff
@@ -1406,18 +1404,47 @@ protected function _build_message()
...
+					}
+					else
+					{
+						$hdr .= $rel_boundary_header;
+					}
+
+					array_unshift($prepared_attachment_parts, $this->_prep_attachments($attachments_indexed_by_multipart['related'], $this->_rel_boundary));
```
narfbg: Is there a particular reason why `array_unshift()` is used here as opposed to just processing 'related' attachments first? i.e. Does that relate to the `$last_boundary` value?

system/libraries/Email.php
```diff
@@ -1406,18 +1404,47 @@ protected function _build_message()
 					.$this->newline
 					.$this->_body.$this->newline.$this->newline;
 
-			break;
+				$attachment_prepared = $this->_prep_attachments($this->_attachments, $this->_atc_boundary);
```
narfbg: Aside from the method name implying that multiple attachments are being handled, but the variable being in a singular form, does the variable really need to exist? Why not just append to `$body`?

system/libraries/Email.php
r4ngy4l: Fixed

system/libraries/Email.php
r4ngy4l: Refactored

system/libraries/Email.php
r4ngy4l: Basically the order of the multipart headers should be 'mixed' -> 'related', but in the content the 'related' should come first and then the 'mixed'.

Refactored: Processing of the attachment related multipart headers and mulipart contents are separated now.

r4ngy4l:
All further issues commented are fixed in 5277583.

system/libraries/Email.php
```diff
@@ -1432,32 +1454,75 @@ protected function _build_message()
 					.$this->_prep_quoted_printable($this->_body).$this->newline.$this->newline
 					.'--'.$this->_alt_boundary.'--'.$this->newline.$this->newline;
 
-			break;
+				( ! empty($attachments_indexed_by_multipart['related'])) && $body .= $this->newline.$this->newline
```
narfbg: I know I suggested using a similar style, but we only do that for shorter lines (and we replace `&&` with `OR` instead of using negation :)).

Anyway, I'll merge and correct this myself as I want to do other adjustments too and there's no point in just telling you what to do at this point.

r4ngy4l: Thank you :smile:

narfbg:
@r4ngy4l Can you test with 21a41ba37b9248291939cf4c801ac0f27982668e applied on top of your patches? (or just pull the feature/email-attachments branch)

r4ngy4l:
@narfbg I have tested with feature/email-attachments branch and have found 2 issues. Please see them commented on 21a41ba37b9248291939cf4c801ac0f27982668e.
I have fixed those issues on my local machine and tested with Gmail, Thunderbird and iOS Mail and it worked fine.

narfbg:
Great, thank you.

PR #4591, opened 2016-04-20
add Input Class json method

ytetsuro:
application/json request get parameter


narfbg:
If we do this, we also have to do it for every single format that exists - not happening.

PR #4601, opened 2016-04-27
library config split file for each Controller class Name

ytetsuro:
The problem of config file fat for "Form Validation Class", "Upload Class" etc 
I want slim config file for split by Controller class Name


narfbg:
No thanks.

PR #4602, opened 2016-04-27
Update Lang.php

pedrocosta:
Added a boolean to allow the return of the line key (between **) if that line is not found


narfbg:
No, this is introducing parameter creep (and the `TRUE` default is a BC break).

PR #4606, opened 2016-04-28
Exceptions xmlrpc

jim-parry:
Exceptions refactoring for xmlrpc & xmlrpcs libraries.

(Getting back to working on the exceptions feature branch started on in the fall :-/)

### DIFF
--- a/system/libraries/Xmlrpc.php
+++ b/system/libraries/Xmlrpc.php
@@ -39,7 +39,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 
 if ( ! function_exists('xml_parser_create'))
 {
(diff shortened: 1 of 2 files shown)


jim-parry:
It looks like the unit tests are failing because openSSL is not enabled. Am I doing something wrong? The unit tests run fine on my dev system.

narfbg:
You're not doing anything wrong ... Travis broke it for PHP 5.3, which is irrelevant to 3.1.0 anyway - I'll look into that.

On the patch itself though ... I'm thinking we could just remove the `if ( ! function_exists('xml_parser_create'))` checks. PHP's own fatal error really is clearer than our message.

PR #4607, opened 2016-04-29
Fix notice in cases where method happens to be empty

animalillo:
In some cases while routing some urls, I'm not too sure exactly what causes this, but it happens that method gets to be empty and a php notice "Uninitialized string offset: 0" is shown on the page or ajax requests.


narfbg:
Please find the root cause for the issue first.

PR #4610, opened 2016-05-03
Exceptions pagination

jim-parry:
Convert error-handling to exceptions in Pagination

### DIFF
--- a/system/libraries/Pagination.php
+++ b/system/libraries/Pagination.php
@@ -394,7 +394,7 @@ class CI_Pagination {
 	/**
 	 * Generate the pagination links
 	 *
-	 * @throws	RuntimeException	If the number of links was incorrectly initialized
+	 * @throws	UnexpectedValueException	If the number of links was incorrectly initialized
 	 * @return	string
 	 */
 	public function create_links()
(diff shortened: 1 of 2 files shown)


system/libraries/Pagination.php
```diff
@@ -419,7 +420,7 @@ public function create_links()
 
 		if ($this->num_links < 0)
 		{
-			show_error('Your number of links must be a non-negative number.');
+			throw new RuntimeException('Your number of links must be a non-negative number.');
```
narfbg: Should be:

> Pagination: Number of links can't be negative

Clarity being the goal, we prefix with `Library_name:` to indicate where the exception comes from, and simplify the message itself as much as possible. :)

narfbg: On another note, it should also be one of `UnexpectedValueException` or `RangeException` ... I'm voting for the former.

system/libraries/Pagination.php
```diff
@@ -394,6 +394,7 @@ public function initialize(array $params = array())
 	/**
 	 * Generate the pagination links
 	 *
+	 * @throws	RuntimeException	If the number of links was incorrectly initialized
```
narfbg: This too is unnecessarily vague; something like "In case of invalid configuration" should be better.

(obviously, the same goes for the documentation)

jim-parry:

PR #4623, opened 2016-05-10
Fix a missing doc link [ci skip]

rochefort:


rochefort:
I gathered it into #4625

PR #4625, opened 2016-05-10
Fix docs [ci skip]

rochefort:


narfbg:
If you've got more of these up your sleeve, please group them in a single pull request and briefly list them in the description (which you haven't used so far).

Appreciate your effort of course, but it's three PRs for three one-liners in a row now ... kinda noisy.

rochefort:
I gathered them into this PR.
I got it, so from now on I group like this fixes.
I'm sorry, I'm translating docs into Japanese.

narfbg:
I meant that only for future PRs ... :)

PR #4626, opened 2016-05-12
Exceptions driver

jim-parry:
Exception handling for PDO driver.

### DIFF
--- a/system/database/drivers/pdo/pdo_driver.php
+++ b/system/database/drivers/pdo/pdo_driver.php
@@ -75,6 +75,7 @@ class CI_DB_pdo_driver extends CI_DB {
 	 *
 	 * @param	array	$params
 	 * @return	void
+	 * @throws	RuntimeException	In case of failure
 	 */
 	public function __construct($params)
 	{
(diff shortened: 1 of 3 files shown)


system/database/drivers/pdo/pdo_driver.php
```diff
@@ -109,7 +110,7 @@ public function __construct($params)
 
 			if ($this->db_debug)
 			{
-				show_error('Invalid or non-existent PDO subdriver');
+				throw new RuntimeException('PDO_driver: Invalid or non-existent PDO subdriver');
```
narfbg: 'Database' instead of 'PDO_driver'

system/libraries/Driver.php
```diff
@@ -108,7 +109,7 @@ public function load_driver($child)
 			// The requested driver isn't valid!
 			$msg = 'Invalid driver requested: '.$child_name;
```
narfbg: This needs to be prefixed with 'Database' too, and it should be hard-coded into the `RuntimeException` throw ...

Exceptions are logged, and therefore `log_message()` should be removed.

system/libraries/Driver.php
```diff
@@ -178,7 +179,7 @@ public function load_driver($child)
 			{
 				$msg = 'Unable to load the requested driver: '.$class_name;
 				log_message('error', $msg);
-				show_error($msg);
+				throw new RuntimeException($msg);
```
narfbg: Same as on line 110

jim-parry:

PR #4627, opened 2016-05-12
Exceptions ftp

jim-parry:
Add exception handling to the Ftp library.

### DIFF
--- a/system/libraries/Ftp.php
+++ b/system/libraries/Ftp.php
@@ -656,12 +656,13 @@ class CI_FTP {
 	 *
 	 * @param	string	$line
 	 * @return	void
+	 * @throws	RuntimeException	In case of failure
 	 */
 	protected function _error($line)
 	{
 		$CI =& get_instance();
 		$CI->lang->load('ftp');
-		show_error($CI->lang->line($line));
+		throw new RuntimeException($CI->lang->line($line));
 	}
 
 }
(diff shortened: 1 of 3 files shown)


system/libraries/Ftp.php
```diff
@@ -656,12 +656,13 @@ public function close()
...
 	 * @param	string	$line
 	 * @return	void
+	 * @throws	RuntimeException	In case of failure
 	 */
 	protected function _error($line)
 	{
 		$CI =& get_instance();
 		$CI->lang->load('ftp');
-		show_error($CI->lang->line($line));
+		throw new RuntimeException('Ftp: ' . $CI->lang->line($line));
```
narfbg: We don't put spaces around concatenation.

I'm also not sure about this whole thing here ... translated exceptions smell, we may remove the translations.

jim-parry:
Translated exception/error messages ... if these are meant to be recoverable, I don't see the issue. How is this different from the other libraries (eg email & form_validation) that have similar provisions?

narfbg:
Form validation errors are displayed to the end user, exceptions shouldn't be.

Don't know about Email, haven't looked into it really, but it wouldn't surprise me if somebody made error translations there for "consistency" with everything else - similar to what we have here, bad decisions have been made in the past.
Unless the application itself is primarily an FTP client, there's no reason to trigger fatal errors and/or to show these errors to the end user at all.

jim-parry:

PR #4628, opened 2016-05-12
Exceptions migration

jim-parry:
Add exception handling to the Migration library.

### DIFF
--- a/system/libraries/Migration.php
+++ b/system/libraries/Migration.php
@@ -112,6 +112,7 @@ class CI_Migration {
 	 *
 	 * @param	array	$config
 	 * @return	void
+	 * @throws	RuntimeException	In case of failure
 	 */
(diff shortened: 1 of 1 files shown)


system/libraries/Migration.php
```diff
@@ -131,7 +133,7 @@ public function __construct($config = array())
 		// Are they trying to use migrations while it is disabled?
 		if ($this->_migration_enabled !== TRUE)
 		{
-			show_error('Migrations has been loaded but is disabled or set up incorrectly.');
+			throw new InvalidArgumentException('Migration: has been loaded but is disabled or set up incorrectly.');
```
narfbg: `InvalidArgumentException` doesn't seem right here.
Also, the text probably needs more adjustment than that.

jim-parry:

PR #4633, opened 2016-05-16
allow multiple validation callbacks for uploads

zburke:
The current implementation of _execute discards all but the first validation rule for fields with empty post data. When a field corresponds to a file-upload field, this means it is not possible to, for example, handle the file upload in the first callback and perform validation of the uploaded file in subsequent callbacks.

This PR provides support for cascading callbacks for fields with empty post data.

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -581,6 +581,7 @@ class CI_Form_validation {
 		}
 
 		// If the field is blank, but NOT required, no further tests are necessary
+		$tmp_rules = [];
 		$callback = FALSE;
 		if ( ! in_array('required', $rules) && ($postdata === NULL OR $postdata === ''))
 		{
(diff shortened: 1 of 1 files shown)


narfbg:
I don't understand your description. Can you elaborate more, and possibly with a code example?

zburke:
In short, this brings the handling of callback functions for file-upload fields into line with how callbacks are handled for regular post fields, allowing multiple callbacks to be chained like this: 
````php
$rules = array(
    array(
        'field' => 'poster_file', 
        'label' => 'Poster File', 
        'rules' => 'callback_handle_upload|callback_validate_page_count|callback_validate_page_size'
    ),
);
````
Here is a use case example: When uploading files, if you do the upload in a callback function via the Form Validation library then you can use the form validation methods like `validation_errors()` to handle any errors: 
````php
public function do_upload()
{
        $rules = array(
            array('field' => 'poster_file', 'label' => 'Poster File', 'rules' => 'callback_handle_upload'),
        );
        $this->form_validation->set_rules($rules);
        if ($this->form_validation->run()) {
            ...
        }
}

public function handle_upload($str)
{
    $this->load->library('upload', $config);
    if ($this->upload->do_upload()) {
        ...
        return TRUE;
    }

    $this->form_validation->set_message('handle_upload', $this->upload->display_errors());
    return FALSE;
}
````
This works fine if all you want to do is upload the file. But if you want to do any post-processing, you would have to do it all in the `handle_upload` callback because when you chain together multiple callbacks as in the first example above, everything after `callback_handle_upload` is silently discarded by the if-block starting near line 583. 

You can work around this by performing the upload and the validation together in that first callback, but it's much cleaner to use separate callback functions for each distinct validation task. Additionally, since regular post fields support chained callbacks, it is more consistent if file-upload fields support it too.

narfbg:
I get what you're trying to do now.

You're making this so hard though ... this has nothing to do with file uploads, the library doesn't care for them at all. You're using a logical *hack* to somehow inject your uploads into the form validation process, and have then built your whole argumentation around that.

It's a very simple problem - `CI_Form_validation::_execute()` doesn't check for more than one callback rule (for non-required, non-submitted/empty fields), regardless of the field type or if it is the first rule or not. And that's a bug.

I'll work on a patch of my own, as I want to also make other adjustments to the relevant code, but please read our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html) if you want to contribute in the future - your patch currently doesn't comply with it.

zburke:
You're sharp. Sorry about my lousy report and formatting. Let me know if you want me to redo the PR with a better explanation and correct style, just in case you decide not to tweak the surrounding code at this time.

narfbg:
The code in question being a mess is why such bugs exist ... the commit linked about should improve that. :)

PR #4637, opened 2016-05-19
Not returning array when there's no DB error

khaosdoctor:
In Oracle Drivers I've noticed that when we call `$this->db->error()` and there's no error, then we get a `false` operator, which allows us to compare in our applications.

When I tested the same result with PostgreSQL I came across a possible bug. Even when there's no error in the query (or even when there's no query executed at all) the function `$this->db->error()` always return a blank array with `code` and `message` keys.

I rewrote the function to check if there's actually an error and return `false` or the array depending on its eval.

### DIFF
--- a/system/database/drivers/postgre/postgre_driver.php
+++ b/system/database/drivers/postgre/postgre_driver.php
@@ -471,7 +471,25 @@ class CI_DB_postgre_driver extends CI_DB {
 	 */
 	public function error()
 	{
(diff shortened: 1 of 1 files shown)


narfbg:
There is a bug here, but the intended result is the opposite of what you want ...

See linked commit above.

aanbar:
@narfbg I guess you have an error in your fix on line 584 'message' should be a key not a value.

narfbg:
Indeed, thanks.

PR #4638, opened 2016-05-19
Fixed PHPDoc parameter name and type discrepancies

kasimtan:


system/helpers/form_helper.php
```diff
@@ -568,7 +568,7 @@ function form_button($data = '', $content = '', $extra = '')
 	 *
 	 * @param	string	The text to appear onscreen
 	 * @param	string	The id the label applies to
-	 * @param	string	Additional attributes
+	 * @param	array 	Additional attributes
```
narfbg: You're adding a space here.

kasimtan: Fixed.

PR #4643, opened 2016-05-24
fixed@4642:sometime, pasteing image no ext from clipbroad,default giv…

lqmcn:
…e .png


narfbg:
Issue and pull request IDs are referenced via the `#` character.
`@` is for usernames (and I do mean usernames, not people's real names).
Also, don't write your whole description in the title.

Closing ... #4643 has the details why.

Issue #3250, opened 2014-09-23
Moving logic from create_links() in pagination


aanbar:
I am thinking that create_links() method shouldn't do all the work, it should just create links in html form.

What if we add a protected method that runs at the end of initialize which generates the data like 'next_url', 'prev_url', 'page_numbers' ...etc;

then these data are available to access via getters which create_links() will use, this would make it easier to provide json response of pagination for APIs

sv3tli0:
Perhaps a refactorisation of this lib will be a good point in future 3.1 version ? 
I am volunteer to do it when there is an accepted idea for its next version :)

narfbg:
I don't imagine anyone still actively looking at this, or anything related to the shitty Pagination library for that matter, but ... pull requests are good and issues just sitting open like this one are useless.

PR #4650, opened 2016-05-26
#3250 create_links logic refactoring + JSON format added

eudgen-dev:
Moved the logic for next\previous URLs out of the create_links() method and added pagination_format option(can take values 'html'{default} or 'json') to return links in JSON format for API as users were requesting.


aanbar:
This isn't what I had in mind, create_links should always return html (a string) in your case it's a breaking change since you're returning arrays everywhere.

My thought the processing should be done at initialize() which creates a data structure you can consume in create_links(); And making the data structure accessible You're free to do whatever you want to do with it.

Issue #4477, opened 2016-02-22
Profiler only supports HTML output - Virtually impossible to use for JSON API and/or CLI


ashiina:
While the CodeIgniter profiler is useful, it is only able to output data via HTML. 
We use CodeIgniter for REST APIs (JSON output), and also use CLI outputs. 
The profiler outputs a ton of HTML tags, so it is almost impossible to use the profiler in any sort of practical way.

I propose that the profiler output should be done via a template file (like the DB errors being outputted via the `errors/error_db.php` template), instead of raw HTML code written in the core. 
- This way, people like me can customize the output so that it is more friendly to our own environment, while not hurting the current users of CI

If this is a valid proposal, I would be more than happy to work on this myself and submit a PR. 
I just want to know beforehand if I am not facing the wrong direction, since I will need to put some amount of work into this.

narfbg:
Sure, that'd be a welcome improvement, although I can't guarantee that I'd like the patch itself.

However, please note that CI itself was never designed for and is itself not suitable for building REST APIs, and its CLI appliances are limited. Switching the profiler to use templates won't change that.

ashiina:
Thanks for the reply.
Yep I understand that CI is primarily for web applications so APIs/CLIs are not its priority - though I believe there should be somewhat of a workaround for people with those use-cases (without hacking into the framework's source code). 

I will submit a PR once I get to this, so please tell me what you think then!

aanbar:
I think this is a great idea, but I would suggest converting profiling statistics into a data structure which means ``_compile_*`` methods will append data to an array that's defined in the class then the method that does the output would use the compiled data from that array.

Keep in mind that simply returning JSON as output won't get your API data correct as you'd end up with 2 json arrays unless you have a way to merge profiling data to your original API output.

The best way I would think off is generating a JSON encoded array of the profiling data & appending it as a commented out structure so it won't affect your API output

Don't forget to add these flags when you output the JSON profiling data ``JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES``

I would suggest modifying ``$this->output->enable_profiler()`` method after you finish so it accepts a second argument to set the output type e.g 'html', 'json'

ashiina:
@aanbar Thanks for the tips! Converting the profiling statistics is exactly the approach I am taking so glad to hear I am on the same page :)
[...]

PR #4653, opened 2016-05-29
Separate profiler view to a template file

ashiina:
A PR for this issue: https://github.com/bcit-ci/CodeIgniter/issues/4477

This PR simply removes the HTML from `libraries/Profiler.php` , and instead reads from a separate HTML template file `application/views/profiler/profiler_template.php` . 
Developers will be free to modify the template to suit their needs.

### DIFF
--- a/application/views/profiler/profiler_template.php
+++ b/application/views/profiler/profiler_template.php
@@ -1,0 +1,263 @@
+
+<div id="codeigniter_profiler" style="clear:both;background-color:#fff;padding:10px;">
+
+<?php if ($fields_displayed > 0): ?>
+	<fieldset id="ci_profiler_benchmarks" style="border:1px solid #900;padding:6px 10px 10px 10px;margin:20px 0 20px 0;background-color:#eee;">
+	<legend style="color:#900;">
+		&nbsp;&nbsp;<?php echo $this->CI->lang->line('profiler_benchmarks'); ?>&nbsp;&nbsp;
+	</legend>
+
+	<table style="width:100%;">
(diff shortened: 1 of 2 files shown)


narfbg:
Haven't looked into it yet, just referencing #3847 as relevant.

ashiina:
@narfbg Thanks for the reference; didn't know there was a similar PR before. Since my current commit requires a newly implemented view, it will apply to your past comment: 

> If 3.0.0 requires nothing for this to work, and then you change it in 3.0.1 to require a view, how is it not a BC break?

To keep BC I cannot introduce any new files, so frankly I will propose just sticking the content of `application/views/profiler/profiler_template.php` into my newly refactored `Profiler.php` as a separate method or something. 
Not pretty, I admit, but much better than the current state of `Profiler.php`.

Please tell me if that's an appropriate direction to be heading.

narfbg:
This is targeting 3.1.0, where we'll have a few minor BC breaks like this, so ... no, you don't need to do that.

ashiina:
@narfbg Great, thanks! I personally prefer a separate view for the profiler (even by breaking BC) so that's good news.
Please let me know if there are any poor code or design decisions which require fixing or explaining. Thanks!

Ema4rl:
+1 :+1: 
I support this.

narfbg:
Sorry all that I've left this one to rot. The Profiler has always been just an uninteresting chore for me.

@gxgpet I don't know if there are still people interested, but feel free to make any decision on this one.

PR #4654, opened 2016-05-29
add Form_validation new rule is_exists

ytetsuro:
add new Validation rule is_exists
Check if the input value already exist


narfbg:
Duplicate of #3574.

PR #4678, opened 2016-06-26
Fix oci8_forge

kenjis:
* Oralce does not have `create table if exists`
* Oracle has only NUMBER for integer

### DIFF
--- a/system/database/drivers/pdo/subdrivers/pdo_oci_forge.php
+++ b/system/database/drivers/pdo/subdrivers/pdo_oci_forge.php
@@ -54,6 +54,13 @@ class CI_DB_pdo_oci_forge extends CI_DB_pdo_forge {
 	protected $_create_database	= FALSE;
 
 	/**
+	 * CREATE TABLE IF statement
+	 *
+	 * @var	string
+	 */
+	protected $_create_table_if	= FALSE;
+
+	/**
 	 * DROP DATABASE statement
 	 *
 	 * @var	string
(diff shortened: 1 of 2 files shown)


narfbg:
Any changes to the oci8 driver should be applied to pdo_oci too.

Also, while there is no INTEGER type per se, Oracle does recognize the keyword and automatically creates NUMBER fields instead - we don't do aliases for aliases.

... and this is why you shouldn't submit patches that aggregate multiple changes.

kenjis:
> Also, while there is no INTEGER type per se, Oracle does recognize the keyword and automatically creates NUMBER fields instead - we don't do aliases for aliases.

When I run migration, if I remove the `_attr_type` function int this PR, I got the following error:
```
Database error: A Database Error Occurred

	Error Number: 907
	ORA-00907: 右カッコがありません。
	CREATE TABLE "migrations" (
	"version" BIGINT(20) NOT NULL
)
	Filename: libraries/Migration.php
	Line Number: 173
```
But the code with this PR, I can run migration.

narfbg:
Fine.

Please make the same changes for pdo/oci.

kenjis:
I pushed.

system/database/drivers/pdo/subdrivers/pdo_oci_forge.php
```diff
@@ -54,6 +54,13 @@ class CI_DB_pdo_oci_forge extends CI_DB_pdo_forge {
 	protected $_create_database	= FALSE;
 
 	/**
+	 * CREATE TABLE IF statement
+	 *
+	 * @var	string
+	 */
+	protected $_create_table_if	= FALSE;
```
narfbg: Look 15 lines below ... this property already exists.

kenjis:
I'm sorry. I fixed it.

PR #4680, opened 2016-06-30
fixes #4679 bug in handling of IPv6 reverse proxies

### LINKED ISSUE #4679: ip_address() hangs when reverse proxy is IPv6
This is a bug I originally posted here:

http://forum.codeigniter.com/thread-65587-post-334025.html#pid334025

I have done some more digging and found this is a bug in Code Igniter's core `ip_address()` function. This bug is in CI 3.0.6.

When a reverse proxy uses an IPv6 address that is in the `proxy_ip` list in the config file, the `ip_address()` function gets stuck in a loop. 

The PHP file below is a stripped out version of `/system/core/Input.php` that recreates the problem. I hard coded `$proxy_ips`, the `HTTP_X_FORWARDED_FOR` header, and the `REMOTE_ADDR` header to simulate the bug. 

Save the code below as `Input.php` and run it with 

    $ php Input.php

Comment out line 35 to see the function run properly with an IPv4 Address.

    <?php

    // Proxy IPs list. 
    // In production this would be set in config.php
    $proxy_ips = [
        '103.21.244.0/22',
        '103.22.200.0/22',
        '103.31.4.0/22',
        '104.16.0.0/12',
        '108.162.192.0/18',
        '141.101.64.0/18',
        '162.158.0.0/15',
        '172.64.0.0/13',
        '173.245.48.0/20',
        '188.114.96.0/20',
        '190.93.240.0/20',
        '197.234.240.0/22',
        '198.41.128.0/17',
        '199.27.128.0/21',
        '2400:cb00::/32',
        '2405:8100::/32',
        '2405:b500::/32',
        '2606:4700::/32',
        '2803:f800::/32',
    ];

    // Set fake forwarded for IP. In production this would be set by
    // cloudflare's reverse proxy. 
    $_SERVER['HTTP_X_FORWARDED_FOR'] = '8.8.8.8';

    // Set IP Address manually. 
    // In production this would come from $_SERVER['REMOTE_ADDR'];
    $ip_address = '103.31.4.1';  // When the reverse proxy is on an IPv4 address
    $ip_address = '2001:5b0:2b25:a4e0:1900:ef83:2fe9:5c88';  // If you uncomment this line we get stuck in a loop.

    if ( ! empty($proxy_ips) && ! is_array($proxy_ips))
    {
        $proxy_ips = explode(',', str_replace(' ', '', $proxy_ips));
    }

    if ($proxy_ips)
    {
        foreach (array('HTTP_X_FORWARDED_FOR', 'HTTP_CLIENT_IP', 'HTTP_X_CLIENT_IP', 'HTTP_X_CLUSTER_CLIENT_IP') as $header)
        {
            if (($spoof = server($header)) !== NULL)
            {
                // Some proxies typically list the whole chain of IP
                // addresses through which the client has reached us.
                // e.g. client_ip, proxy_ip1, proxy_ip2, etc.
                sscanf($spoof, '%[^,]', $spoof);

                if ( ! valid_ip($spoof))
                {
                    $spoof = NULL;
                }
                else
                {
                    break;
                }
            }
        }

        if ($spoof)
        {
            for ($i = 0, $c = count($proxy_ips); $i < $c; $i++)
            {
                var_dump($i);

                // Check if we have an IP address or a subnet
                if (strpos($proxy_ips[$i], '/') === FALSE)
                {
                    // An IP address (and not a subnet) is specified.
                    // We can compare right away.
                    if ($proxy_ips[$i] === $ip_address)
                    {
                        $ip_address = $spoof;
                        break;
                    }

                    continue;
                }

                // We have a subnet ... now the heavy lifting begins
                isset($separator) OR $separator = valid_ip($ip_address, 'ipv6') ? ':' : '.';
                
                // If the proxy entry doesn't match the IP protocol - skip it
                if (strpos($proxy_ips[$i], $separator) === FALSE)
                {
                    continue;
                }

                // Convert the REMOTE_ADDR IP address to binary, if needed
                if ( ! isset($ip, $sprintf))
                {
                    if ($separator === ':')
                    {
                        // Make sure we're have the "full" IPv6 format
                        $ip = explode(':',
                            str_replace('::',
                                str_repeat(':', 9 - substr_count($ip_address, ':')),
                                $ip_address
                            )
                        );

                        for ($j = 0; $j < 8; $j++)
                        {
                            $ip[$j] = intval($ip[$j], 16);
                        }

                        $sprintf = '%016b%016b%016b%016b%016b%016b%016b%016b';
                    }
                    else
                    {
                        $ip = explode('.', $ip_address);
                        $sprintf = '%08b%08b%08b%08b';
                    }

                    $ip = vsprintf($sprintf, $ip);
                }

                // Split the netmask length off the network address
                sscanf($proxy_ips[$i], '%[^/]/%d', $netaddr, $masklen);

                // Again, an IPv6 address is most likely in a compressed form
                if ($separator === ':')
                {
                    $netaddr = explode(':', str_replace('::', str_repeat(':', 9 - substr_count($netaddr, ':')), $netaddr));
                    for ($i = 0; $i < 8; $i++)
                    {
                        $netaddr[$i] = intval($netaddr[$i], 16);
                    }
                }
                else
                {
                    $netaddr = explode('.', $netaddr);
                }

                // Convert to binary and finally compare
                if (strncmp($ip, vsprintf($sprintf, $netaddr), $masklen) === 0)
                {
                    $ip_address = $spoof;
                    break;
                }
            }
        }
    }

    if ( ! valid_ip($ip_address))
    {
        return $ip_address = '0.0.0.0';
    }

    // die instead of returning
    die($ip_address);

    // --------------------------------------
    // --------------------------------------
    // --------------------------------------

    function valid_ip($ip, $which = '')
    {
        switch (strtolower($which))
        {
            case 'ipv4':
                $which = FILTER_FLAG_IPV4;
                break;
            case 'ipv6':
                $which = FILTER_FLAG_IPV6;
                break;
            default:
                $which = NULL;
                break;
        }

        return (bool) filter_var($ip, FILTER_VALIDATE_IP, $which);
    }

    /**
     * Fetch an item from the SERVER array
     *
     * @param    mixed    $index        Index for item to be fetched from $_SERVER
     * @param    bool    $xss_clean    Whether to apply XSS filtering
     * @return    mixed
     */
    function server($index, $xss_clean = NULL)
    {
        return _fetch_from_array($_SERVER, $index, $xss_clean);
    }

    /**
     * Fetch from array
     *
     * Internal method used to retrieve values from global arrays.
     *
     * @param    array    &$array        $_GET, $_POST, $_COOKIE, $_SERVER, etc.
     * @param    mixed    $index        Index for item to be fetched from $array
     * @param    bool    $xss_clean    Whether to apply XSS filtering
     * @return    mixed
     */
    function _fetch_from_array(&$array, $index = NULL, $xss_clean = NULL)
    {
        $xss_clean = true;

        // If $index is NULL, it means that the whole $array is requested
        isset($index) OR $index = array_keys($array);

        // allow fetching multiple keys at once
        if (is_array($index))
        {
            $output = array();
            foreach ($index as $key)
            {
                $output[$key] = _fetch_from_array($array, $key, $xss_clean);
            }

            return $output;
        }

        if (isset($array[$index]))
        {
            $value = $array[$index];
        }
        elseif (($count = preg_match_all('/(?:^[^\[]+)|\[[^]]*\]/', $index, $matches)) > 1) // Does the index contain array notation
        {
            $value = $array;
            for ($i = 0; $i < $count; $i++)
            {
                $key = trim($matches[0][$i], '[]');
                if ($key === '') // Empty notation will return the value as array
                {
                    break;
                }

                if (isset($value[$key]))
                {
                    $value = $value[$key];
                }
                else
                {
                    return NULL;
                }
            }
        }
        else
        {
            return NULL;
        }

        return $value;
    }

alan-reed:
Signed-off-by: Alan Reed <areed@prouddev.com>

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -519,9 +519,9 @@ class CI_Input {
 					if ($separator === ':')
 					{
 						$netaddr = explode(':', str_replace('::', str_repeat(':', 9 - substr_count($netaddr, ':')), $netaddr));
-						for ($i = 0; $i < 8; $i++)
+						for ($j = 0; $j < 8; $j++)
 						{
-							$netaddr[$i] = intval($netaddr[$i], 16);
+							$netaddr[$j] = intval($netaddr[$j], 16);
 						}
 					}
 					else


narfbg:
Sorry, I didn't see this before pushing 9fd9248a2d712d5ae95bf2e6c6cd036e6b522cbb

Issue #2717, opened 2013-11-07
Extend DB Drivers...?


a-h-abid:
Today I tried to extend a DB driver in hope that it will work on CI 3.0 but seems like it has not been implemented yet. Placed in `application/database/drivers/mysqli/MY_mysqli_driver.php`. Needed for adding "insert duplicate update" feature. We currently extend via this approach:

https://github.com/EllisLab/CodeIgniter/wiki/Extending-Database-Drivers

But to me this seems like extra work for extending a DB driver. Plus in case the database() method is updated in main Loader.php and forgot to update in extended class, then we'll be missing any fixes/changes that are done.

So would be great if made easier to extend any db driver just like extending core, helpers or libraries.

Nudasoft:
Just wondering if this feature is gonna implement in the near future?

narfbg:
No. I've slowly convinced myself out of it and with CI4 out I now have even less reason to think of big new features for CI3.

PR #4684, opened 2016-07-05
Make CI_DB_query_builder customisable like any core class

Growiel:
We needed to be able to customize the way DB's update(), insert() and delete() work.

We expected to be able to just create MY_DB_query_builder in application/core and that it would be used, just as easily as extending the Model or Controller.

Turns out it doesn't work, I look around a lot, some people also tried it without any luck, so I decided to fix it myself.
It's heavily inspired by what's happening inside Loader::model() for the checking for the file and class part.

I then use `eval()` to create the CI_DB class extending the correct class. I know a lot of people think that "eval() is evil", but in this case, it's 100% safe and the only way to make this work.

I do realize that the `application/core` folder is design to override only `system/core` files, and not `system/database`, but what I provide is just the base for you to build on. Maybe create an `application/database folder` ?

Also this only works for the CI_DB_query_builder as its the one I needed, but it can be extended to CI_DB_driver or any other DB class really.

### DIFF
--- a/system/database/DB.php
+++ b/system/database/DB.php
@@ -169,25 +169,35 @@ function &DB($params = '', $query_builder_override = NULL)
 	if ( ! isset($query_builder) OR $query_builder === TRUE)
 	{
 		require_once(BASEPATH.'database/DB_query_builder.php');
-		if ( ! class_exists('CI_DB', FALSE))
+		$app_path = APPPATH.'core/'.DIRECTORY_SEPARATOR;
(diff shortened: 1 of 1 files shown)


narfbg:
#2717.

Growiel:
So it's "not something your consider for 3.0", I get that (I don't approve, but I get it), so is that something that will be built in 4.0 from the start or will we have to hack the core again ? Just curious to know how I should approach this change.

lonnieezell:
It should be pretty simple to do in CI4.

PR #4688, opened 2016-07-05
Updated comment of config.php by adding more info.

harikrishnanp:
Fix of bug https://support.ellislab.com/bugs/detail/21898/getting-database-error-when-we-set-sess_match_ip-to-true-in-config.php-and-


narfbg:
Adding more comments is not "fix of bug" - sorry, I'm nitpicky like that.

This doesn't help because it's one-way (only considers switching from Off to On, but not the reverse), and I don't want users to have the false sense of confidence that not reading the [documentation](http://www.codeigniter.com/userguide3/libraries/sessions.html#database-driver) is ok.

PR #4689, opened 2016-07-05
Update Config.php

vibbow:
Not all code paths return a value, as PHPDoc declare there is a return value.


narfbg:
This line will never execute, it's dead code.

PR #4690, opened 2016-07-05
Update Security.php

vibbow:
Fix PHPDoc to declare correct return type.


narfbg:
We went through this with somebody else, but I can't be bothered to search for an obscure PR right now (which reminds me - please use PR titles appropriately).

Long story short: We don't do this.

Issue #4693, opened 2016-07-06
Language helper incorrect Return value


dkhalife:
The [doc](http://www.codeigniter.com/user_guide/helpers/language_helper.html) mentions that the function `lang` returns `HTML-formatted language line label` which is incorrect as it is not always true.

More precision should be added to say that if only the `$line` parameter is supplied, it actually returns just the `$lang[$line]` value from the language file. Here's an except of lines 68:71 

```
		if ($for !== '')
		{
			$line = '<label for="'.$for.'"'._stringify_attributes($attributes).'>'.$line.'</label>';
		}
```

However, as soon as the `$for` parameter is supplied, only then it will return a complete `<label>` tag with the `for` attribute set accordingly.

Issue #4692, opened 2016-07-06
Inflector helper doc specifications


dkhalife:
The [doc ](http://www.codeigniter.com/user_guide/helpers/inflector_helper.html) should explicitly mention that this Helper is for English language only and there is no supported way of directly extending it to other languages.

Perhaps future plans might include this as part of the language Library instead in order to help profit from localization?

jim-parry:
The current user guide *could* mention your suggestion, but it is not inconsistent with the other frameworks. Cake & Yii, for instance, have inflector helpers that are English only, without expressly stating that. 

As for future plans, this sounds like a can of worms :(
I did not find any frameworks that deal elegantly with multiple languages, hmmm.
The closest was https://github.com/ICanBoogie/Inflector, which handles a handful of languages.
In other words, something easier said than done!

dkhalife:
Yup you are right, I suppose for multilingual cases one could simply add a few extra language lines, it's not a huge deal anyway, thanks!

narfbg:
Nitpicking level: infinity.

PR #4699, opened 2016-07-08
Userguide/languages

jim-parry:
Incorporate two community suggestions for the helper writeups.


narfbg:
I've handled #4692, #4693 manually.

Issue #4197, opened 2015-10-28
composer_autoload config bug


dexterns88:
When you set $config['composer_autoload'] = TRUE; vendor autoload doesn't load into app you mast insert path to vendor folder then work.
Like:
$config['composer_autoload'] =  FCPATH . '/vendor/autoload.php';

CI version 3.0-stable

narfbg:
Quote from the config.php description of composer_autoload:

>| Enabling this setting will tell CodeIgniter to look for a Composer
>| package auto-loader script in application/vendor/autoload.php.

mytharcher:
I also met this. The first think of mine about `composer.json` should be used in web root path. But this configuration need to put that file into `application` folder. Really hard to understand.

feryardiant:
@dexterns88 yes, mine also use the same [config](https://github.com/projek-xyz/ci-startapp/blob/develop/app/config/config.php#L133) as yours and I don't think it's a bug anyway.

@mytharcher [my](https://github.com/projek-xyz/ci-startapp/tree/develop) `composer.json` is in the `FCPATH` and it's work fine :grin:

mytharcher:
Thanks @feryardiant . I now works with `$config['composer_autoload'] = FCPATH . 'vendor/autoload.php';` fine!

PR #4701, opened 2016-07-09
[BUGFIX] Codeigniter tries to get composer vendor in APPPATH

isantolin:
When installs codeigniter using this command:

composer create-project bcit-ci/codeigniter CI-Clean --prefer-dist

install the "vendor" directory on codeigniter root (FCPATH) including "mikey179/vfsstream" but $composer_autoload tries to get it on APPPATH.'vendor/autoload.php'


narfbg:
It's not a bug.
Use the search bar.

#3636
#4197

PR #4703, opened 2016-07-09
Update CodeIgniter.php

Everterstraat:
Added FCPATH to search for on $config['composer_autoload']

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -166,9 +166,18 @@ if ( ! is_php('5.4'))
 	{
 		if ($composer_autoload === TRUE)
 		{
-			file_exists(APPPATH.'vendor/autoload.php')
-				? require_once(APPPATH.'vendor/autoload.php')
-				: log_message('error', '$config[\'composer_autoload\'] is set to TRUE but '.APPPATH.'vendor/autoload.php was not found.');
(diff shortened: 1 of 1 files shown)


georgexsh:
agree with this change, but the commit message needs more detail IMO

mwhitneysdsu:
If you want it to look in `FCPATH.'vendor/autoload.php'`, then set `$config['composer_autoload'] = FCPATH.'vendor/autoload.php';` instead of `$config['composer_autoload'] = TRUE;`...

Everterstraat:
It's possible that way @mwhitneysdsu but why make it harder than needed?!

mwhitneysdsu:
It's just behaving as documented. If you want to change that behavior, you should also update the comment that describes the behavior in the config file.

I don't really care either way, because setting the variable to the path would probably still be faster than setting it to `TRUE` and having it check multiple directories for the file.

narfbg:
Because "easier" doesn't mean "correct".

[Stop](https://github.com/bcit-ci/CodeIgniter/pull/3636) [requesting](https://github.com/bcit-ci/CodeIgniter/issues/4197) [this](https://github.com/bcit-ci/CodeIgniter/pull/4701)! I don't care if a thousand PRs suggest it, it won't be merged.

PR #4691, opened 2016-07-06
Update Upload.php

chestnutprog:
特殊文件名可能会引起bug

PR #4705, opened 2016-07-11
Use negative length for speeding up.

tianhe1986:
I use the following code for benchmark test.

`
                /* just comment */

		$name = "test_benchmark.cpp.cpp";
		$ext = ".cpp";		
		$raw_name = "";
		
		$this->benchmark->mark("positive_start");
		
		$count = 1000000;
		for($i = 0; $i < $count; $i++){			
			$raw_name = substr($name, 0, strlen($name) - strlen($ext));			
		}
		$this->benchmark->mark("positive_end");
		echo "Positive length elapsed ".$this->benchmark->elapsed_time("positive_start", "positive_end")." seconds. "."<br>";
		
		$this->benchmark->mark("negative_start");
		for($i = 0; $i < $count; $i++){
			$raw_name = substr($name, 0, - strlen($ext));
		}
		$this->benchmark->mark("negative_end");
		echo "Negative length elapsed ".$this->benchmark->elapsed_time("negative_start", "negative_end")." seconds. "."<br>";`

The output is as follows:
Positive length elapsed 4.1075 seconds 
Negative length elapsed 2.8315 seconds

### DIFF
--- a/system/libraries/Upload.php
+++ b/system/libraries/Upload.php
@@ -601,7 +601,7 @@ class CI_Upload {
 				'file_type'		=> $this->file_type,
 				'file_path'		=> $this->upload_path,
 				'full_path'		=> $this->upload_path.$this->file_name,
-				'raw_name'		=> substr($this->file_name, 0, strlen($this->file_name) - strlen($this->file_ext)),
+				'raw_name'		=> substr($this->file_name, 0, -strlen($this->file_ext)),
 				'orig_name'		=> $this->orig_name,
 				'client_name'		=> $this->client_name,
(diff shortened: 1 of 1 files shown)


pfote:
So you successfully demonstrated that calling strlen() twice is more expensive than calling it once. Impressive. And that in the upload class where the most time consuming factor is ...

tianhe1986:
Oh I make a joke, just forget it ^^

narfbg:
I did notice this in #4691, but was too lazy to do something about it at the time.

I don't understand why you've closed this though.

PR #4706, opened 2016-07-12
Use real write access instead of "supposed" in Log_test.

tianhe1986:
I get the following info while running unittest.

1) Log_test::test_configuration
Failed asserting that false matches expected true.

C:\workspace\CodeIgniter\tests\codeigniter\core\Log_test.php:36
C:\workspace\CodeIgniter\tests\mocks\ci_testcase.php:351

I use WAMP in windows10 and find it created "c:/root" directory with write access.
It can also be reproduced under linux with root user.


narfbg:
Our tests are made for *our testing environment*, not yours.

PR #4713, opened 2016-07-19
update_batch returns -1 when execute update query failed.

huxinbang:
## Issue
Update batch could cause "Illegal mix of collations (utf8_general_ci,COERCIBLE) and (utf8mb4_general_ci,IMPLICIT) for operation 'case'" failure, when CI's collations and charset was not set correctly. Attach files can show how to make this happens:
[update_batch_return_num_issue.zip](https://github.com/bcit-ci/CodeIgniter/files/370779/update_batch_return_num_issue.zip)

When those kind of error happens, update_batch does not return false as described, but return "-1" instead,  not expected. 

## Patch
Check every return of query call, return FALSE immediately when query call returns FALSE.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1913,7 +1913,11 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		$affected_rows = 0;
 		for ($i = 0, $total = count($this->qb_set); $i < $total; $i += $batch_size)
 		{
-			$this->query($this->_update_batch($this->protect_identifiers($table, TRUE, NULL, FALSE), array_slice($this->qb_set, $i, $batch_size), $this->protect_identifiers($index)));
+			$ret = $this->query($this->_update_batch($this->protect_identifiers($table, TRUE, NULL, FALSE), array_slice($this->qb_set, $i, $batch_size), $this->protect_identifiers($index)));
+			if ($ret)
+			{
+				return ($this->db_debug) ? $this->display_error('db_invalid_query') : FALSE;
+			}
 			$affected_rows += $this->affected_rows();
 			$this->qb_where = array();
 		}


narfbg:
What if it *did* update 85 of the first 100 and then failed on the second batch of 100? Sure, it's a failure, but it still did update 85 records.

Either way, if the problem is caused by something else being incorrectly configured, it's obviously a configuration problem.

Sorry, I'm not saying the current functionality is perfect (it has similar flaws to what has been reported previously for `insert_batch()`), but there are too many question marks around this for me to merge it.

P.S.: I don't understand how `$ret` being positive is a failure check; should be negative.

huxinbang:
> What if it did update 85 of the first 100 and then failed on the second batch of 100? Sure, it's a failure, but it still did update 85 records.

In this situation, another wired thing would happen under current functionality of `update_batch`:  return **84** instead of **85**.  For **$affected_rows** would add (-1) by each failure of query execution.

> Either way, if the problem is caused by something else being incorrectly configured, it's obviously a configuration problem.

Incorrect configuration was not the only way that could trigger this problem. For example, update field(has unique constraint) with duplicate values.

## Suggestion
At least check return of previous query function before `$affected_rows += $this->affected_rows()`.
>	$ret = $this->query($this->_insert_batch($this->protect_identifiers($table, TRUE, $escape, FALSE), $this->qb_keys, array_slice   ($this->qb_set, $i, $batch_size)));
	if ($ret)
	{
		$affected_rows += $this->affected_rows();
	}

narfbg:
Which db driver are you using?

huxinbang:
mysqli

> MysqlI
 Support => enabled
Client API library version => 5.5.44
Active Persistent Links => 0
Inactive Persistent Links => 0
Active Links => 0
Client API header version => 5.5.46
MYSQLI_SOCKET => /var/run/mysqld/mysqld.sock
Directive => Local Value => Master Value
mysqli.allow_local_infile => On => On
mysqli.allow_persistent => On => On
mysqli.default_host => no value => no value
mysqli.default_port => 3306 => 3306
mysqli.default_pw => no value => no value
mysqli.default_socket => /var/run/mysqld/mysqld.sock => /var/run/mysqld/mysqld.sock
mysqli.default_user => no value => no value
mysqli.max_links => Unlimited => Unlimited
mysqli.max_persistent => Unlimited => Unlimited
mysqli.reconnect => Off => Off

sneakyimp:
I'm trying to work with the query builder's insert_batch function. According to the documentation, it should return FALSE on failure. That is not the case if you specify a nonexistent table.

Try this controller method:
```php
	public function barf() {
		$data = array(
				array(
						'title' => 'My title',
						'name' => 'My Name',
						'date' => 'My date'
				),
				array(
						'title' => 'Another title',
						'name' => 'Another Name',
						'date' => 'Another date'
				)
		);
		$retval = $this->db->insert_batch("i_do_not_exist", $data);
		var_dump($retval);
	}
```
Seems to me it should throw an exception or at the very least return false as described in the documentation:
http://www.codeigniter.com/userguide3/database/query_builder.html#CI_DB_query_builder::insert_batch

PR #4714, opened 2016-07-19
Session: use associative array key search instead of array value search in userdata()

tianhe1986:
The efficiency would be highly improved in case of calling userdata() with thousands of flashdata/tempdata.

I use the following code for benchmark test.

```
		$count = 20000;
		for($i = 0; $i < $count; $i++){			
			$this->session->set_tempdata('temp_key'.$i, $i);
			$this->session->set_userdata('key'.$i, $i);
		}
		
		$this->benchmark->mark("session_start");
		$userdata = $this->session->userdata();
		$this->benchmark->mark("session_end");
		echo "Session elapsed ".$this->benchmark->elapsed_time("session_start", "session_end")." seconds. ";
```

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -678,19 +678,16 @@ class CI_Session {
 			return array();
 		}
 
-		$userdata = array();
+		$userdata = $_SESSION;
+		
+		$_filter_keys = isset($_SESSION['__ci_vars']) ? array_keys($_SESSION['__ci_vars']) : array();
 		$_exclude = array_merge(
-			array('__ci_vars'),
-			$this->get_flash_keys(),
-			$this->get_temp_keys()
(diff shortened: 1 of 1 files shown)


narfbg:
If you do have *thousands*, you're doing something very wrong.

tianhe1986:
@narfbg I agree with you, too. However, optimization is necessary.

narfbg:
How can you agree and at the same time say this is necessary?

You can't say that something *pointless* is *necessary*. That's a huge contradiction.

tianhe1986:
@narfbg I mean, it is necessary in programming algorithm, not only with this case. I just recommended to use hash instead of  linear search.

jiaxinming:
@narfbg I think, In the long run it is stable and useful.

PR #4715, opened 2016-07-20
Form dropdown duplicates

tiennou:
I'm in the process of upstreaming a bunch of changes made against 2.1, so...
Per acafee9 :

The rationale is that I have a list of country names where the top few items are a few "top-most" countries, but those must still appear in the alphabetical list so users aren't wondering why they can't find their country.

It also adds nice "separators" (even though they're not in the HTML spec).

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -319,6 +319,63 @@ if ( ! function_exists('form_multiselect'))
 	}
 }
 
+if ( ! function_exists('_form_dropdown_options'))
+{
+	function _form_dropdown_options($options, &$selected = array(), $in_optgroup = FALSE)
+	{
+		$form = '';
+		foreach ($options as $key => $val)
+		{
+			$option = NULL;
+			$is_separator = FALSE;
+			if (is_string($key) && is_string($val))
+			{
+				// Arrayify our parameters for backward-compatibility
+				$option = array('key' => $key, 'value' => $val);
+			}
+			else if (is_array($val) && isset($val['key']))
+			{
+				// New-style construction
+				// If there are no keys to select, default to what's in the options settings
+				if (count($selected) == 0 && isset($val['selected']) && $val['selected'])
+				{
+					$selected[] = $key;
+				}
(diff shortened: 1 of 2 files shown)


narfbg:
No thanks.

tiennou:
Is that a "no thanks" we *really* don't want that, or a "no thanks" that code is a mess ? I mean, my use case is clear, the current function doesn't handle it, so it's an improvement (and I even added a test case). Any chance I could make some changes and it would become acceptable ?

narfbg:
Your use case is quite specific and it is far from clear why it would be useful for anybody else. It may an improvement to you, but that doesn't mean it's a valuable addition for the general public.

And yes, the patch is somewhat messy in a couple of ways:

- You're creating a new function, but only to call it from inside `form_dropdown()` and from nowhere else. Just moving code around between functions makes it hard to follow what the actual additions are.
- These separators are ... let's just say if something's not in the HTML spec, we don't want it.

I do value test cases, not least becase few people make that effort, but that's secondary to everything else.

I don't see how any changes would make this valuable enough to be part of the framework.

tiennou:
We're in agreement w.r.t the separator part, even though people like separators in menus 😉 . I do agree I'd have a hard time coming up with any other use case than my list of countries, but since it's in a way a limitation of the form helper (as there's nothing that prevents duplicate key-values in a `<select>`), I'd thought I'd fix it. The function is created because I can't use a simple array since I have duplicate keys, but `<optgroup>` will eat those, breaking backward-compatibility (hence the weird `array('key' => ..., 'value' => ...)` syntax.

If the duplicate keys are actually something you'd want, I might be able to rework that so as not to have that function. Else I'll keep my patch (or most likely, generate my own `<select>`s the way I want them to be 😉 ).

narfbg:
When there's convenient functionality, there's also limitations - it's all about trade-offs.

Current functionality still allows one key-value as a regular option and then the same pair in an optgroup. This *should* satisfy even your use case, which I would say goes to the extremes of UX.

PR #4718, opened 2016-07-21
Ftp : Close present connection before create new one and also in destructor method

tianhe1986:
Just freeing resource.

### DIFF
--- a/system/libraries/Ftp.php
+++ b/system/libraries/Ftp.php
@@ -151,6 +151,9 @@ class CI_FTP {
 		{
 			$this->initialize($config);
 		}
+		
+		// Close old connection before creating new one
+		$this->close();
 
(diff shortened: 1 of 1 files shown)


narfbg:
You know PHP does this by itself, no?

tianhe1986:
@narfbg Yes, I know PHP's garbage collector with reference-counting system, but I found __destruct in CI_Email and thought they may be the same case. Just forget it.

narfbg:
I'm not talking about the GC, I'm talking about it closing all (non-persistent) connections at the end of script execution.
There are sometimes reasons to close them manually, and that may be the case with `CI_Email`, but in general it's not necessary.

Anyway, if I'm to "forget about it", please close it. :)

PR #4722, opened 2016-07-22
Common: Use case-insensitive comparing/replacing

tianhe1986:
1 Call strtolower before comparing $_SERVER['HTTP_X_FORWARDED_PROTO']  since it is also treated case insensitively in Apache Tomcat([see here](http://tomcat.apache.org/tomcat-8.0-doc/config/filter.html#Basic_configuration_to_handle_'x-forwarded-for'_and_'x-forwarded-proto'))

2 Replace url_encoded invisible characters case-insensitively. For example, both '%0c' and '%0C' should be removed after calling remove_invisible_characters with url_encoded param enabled.

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -716,8 +716,8 @@ if ( ! function_exists('remove_invisible_characters'))
 		// carriage return (dec 13) and horizontal tab (dec 09)
 		if ($url_encoded)
 		{
-			$non_displayables[] = '/%0[0-8bcef]/';	// url encoded 00-08, 11, 12, 14, 15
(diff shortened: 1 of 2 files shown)


narfbg:
I *would* merge these, **but**:

1. Submit them separately; don't change different functionalities in a single PR.
2. Don't explain what you're changing - this is very clearly visible in the patch. Explain *why* you're changing it.
3. Don't use how Apache Tomcat (or any other software) works as argumentation. The HTTP protocol is what dictates how things should work, not a single HTTP server.
4. These test cases are badly written - inconsistent indentation, inverted parameters order.

tianhe1986:
@narfbg OK, I'll submit two separate commits and also with your advise.

PR #4723, opened 2016-07-22
fixed return FALSE when load database error

iyaozhen:
In my case:
```
// maybe can't connect database server
if ($db = $this->load->database('db_conf', true)) {
    // some query to get value
    return $row['field'];
}
else {
    return 'default_value';
}
```
But in code always return $DB(new $driver($params);).

In code note, `CI_Loader.database` return `object|bool`.

### DIFF
--- a/system/database/DB.php
+++ b/system/database/DB.php
@@ -216,6 +216,10 @@ function &DB($params = '', $query_builder_override = NULL)
 		}
 	}
 
-	$DB->initialize();
-	return $DB;
+	if ($DB->initialize() !== FALSE) {
+		return $DB;
+	}
+	else {
+		return FALSE;
+	}
 }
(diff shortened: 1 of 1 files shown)


tianhe1986:
@iyaozhen 

Maybe your case is caused by other reason.

You can find in CI_DB_driver::initialize that a RuntimeException will be thrown if can't connect to db server.

 CI_DB_driver::initialize has no return value, so your commit, which failed to pass continuous-integration check, is not proper.

iyaozhen:
Oh, My mistake. My local code is 3.0, I just diff DB.php file not diff DB_driver.php.

ps: I first use try/catch in my code, but It is not work. I find some old issue, and wrong to think `RuntimeException` only add in version 4.0.

PR #4726, opened 2016-07-22
Code completion

plinfut:
Added properties to Controller, Model and Migration to enable code completion in popular IDEs.

For example: when you type $this->db->, the IDE will suggest query, insert, etc.


narfbg:
It is IDE's job to provide such functionality, not ours.

plinfut:
Of course you are correct in that principle. However, I believe there is more to it than responsibility. 

This support has been suggested numerous times to the Eclipse and Netbeans teams, to no avail, possibly because it is incredibly complicated for an IDE and extremely simple for CI.

But if you stand by your opinion, I'll just repeat these changes myself after updating CI in my projects.

kfern-github:
+1

PR #4742, opened 2016-07-30
Changed the return value of CI_DB_result::_fetch_object(..) to object

masterklavi:


narfbg:
You know this code is **never** executed, right?
How did you even get to it?

masterklavi:
Yes, it's right and of course I know about it.
**But** there assumes that the function returns an object. So it **should** return object. Am I wrong?

masterklavi:
> How did you even get to it?

Through a code review.

PR #4748, opened 2016-08-05
Added a note for helpers alias

Ema4rl:
There is an undocumented feature - `helpers()` method.
I simply added a note to expose this feature.

### DIFF
--- a/user_guide_src/source/libraries/loader.rst
+++ b/user_guide_src/source/libraries/loader.rst
@@ -358,6 +358,8 @@ Class Reference
 		This method loads helper files, where file_name is the name of the
 		file, without the _helper.php extension.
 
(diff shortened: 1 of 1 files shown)


narfbg:
Some people make typos and that's the only thing this method is useful for.

I don't want it used or documented, it's just an old time relic that wouldn't be added to the framework today.

Ema4rl:
>Some people make typos and that's the only thing this method is useful for.

>I don't want it used or documented, it's just an old time relic that wouldn't be added to the framework today.
&mdash; [**@narfbg**](https://github.com/narfbg)

I think anything not documented should not exist. Imagine if there was an undocumented `$CI->load->controller()` method - of what use is it?

If a developer mistakenly uses `$CI->load->helpers()` it should return a **FATAL** error!

This is my standpoint and should be adopted as it prevents _CodeIgniter_ from being an over-bloated framework.

PR #4755, opened 2016-08-09
CI_Unit_test: Do not replace "is_double" with "is_float".

tianhe1986:
"is_double" is already an alias of "is_float".


narfbg:
This changes the resulting report.

tianhe1986:
@narfbg I wonder how it changes the resulting report since `$lang['ut_float']` and `$lang['ut_double']` are the same.

And,  the following code in [unit_test_lang.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/language/english/unit_test_lang.php#L52) is useless if replacing "is_double" with "is_float" because `$extype` would never be "double".
`$lang['ut_double'] = 'Float';`

narfbg:
Now that you said that, it's actually `$lang['ut_float']` that's useless as `gettype()` returns 'double'.

tianhe1986:
@narfbg Sorry, this is my mistake. I should just discuss with `is_*` functions.

Issue #4756, opened 2016-08-09
Working with index metadata for a given table


pkarunyu:
Yipee!! My first attempted contribution to the CI framework, after 7 years of active use!!

**How it started**
I found myself in a situation where I needed to drop and re-create an index on very many tables, but wanted to check if that index exists first.

I assumed since the function `field_exists` exists, then `index_exists` exists, but it did not. I checked the code and it didn't seem too hard to implement, so I tried it.

Basically, this feature would be closely modelled on `field_exists`. This is what I learnt:

**TODO**
* `index_exists` would depend on `list_indexes`
* `list_indexes` depends on `_list_indexes`
* For completeness, I would want to have `index_data`
    
**Questions**

1. The [contribution guide](https://github.com/bcit-ci/CodeIgniter/blob/develop/contributing.md) says `If you change anything that requires a change to documentation then you will need to add it.`, does this mean that, if accepted, the eventual pull request should also include changes to the following sub-sections of the database section of the user guide?
  * [metadata](https://github.com/pkarunyu/CodeIgniter/blob/develop/user_guide_src/source/database/metadata.rst)
  * [DB driver reference](https://www.codeigniter.com/userguide3/database/db_driver_reference.html)
2. The  contribution guide says `One thing at a time: A pull request should only contain one change`, does my suggested changes above constitute `one change`?
3. Do I need to make the change in all the 12 database drivers?

Now, I did attempt to make this change on the MySQL, MySQLi and Postgre database drivers, and it seems to work fine. I will initiate a PR for the change to the MySQLi driver, for review and further guidance.

narfbg:
> I found myself in a situation where I needed to drop and re-create an index on very many tables, but wanted to check if that index exists first.

Most databases support `IF EXISTS`. Also, I don't understand why you want to do this with PHP.

> I assumed since the function `field_exists` exists, then `index_exists` exists, but it did not. I checked the code and it didn't seem too hard to implement, so I tried it.
>
> Basically, this feature would be closely modelled on `field_exists`

`field_exists()` is way more likely to be called on a regular basis. For example, an application may dynamically determine whether it should `SELECT` a certain field or not, probably for BC maintenance purposes.

I'm not at all convinced that `index_exists()` would be beneficial in the same way to our users.

> - `index_exists` would depend on `list_indexes`

That would be sub-optimal.

> - `list_indexes` depends on `_list_indexes`

[...]

PR #4757, opened 2016-08-09
Review PR for issue #4756

pkarunyu:
Requesting for review of changes suggested in issue #4756

### DIFF
--- a/system/database/drivers/mysqli/mysqli_driver.php
+++ b/system/database/drivers/mysqli/mysqli_driver.php
@@ -448,6 +448,21 @@ class CI_DB_mysqli_driver extends CI_DB {
 	}
 
 	// --------------------------------------------------------------------
+	
+	/**
+	 * Show indexes query
+	 *
+	 * Generates a platform-specific query string so that the indexes can be fetched
+	 *
+	 * @param	string	$table
+	 * @return	string
+	 */
+	protected function _list_indexes($table = '')
+	{
+		return 'SHOW INDEXES FROM '.$this->protect_identifiers($table, TRUE, NULL, FALSE);
+	}
+
+	// --------------------------------------------------------------------
 
 	/**
 	 * Returns an object with field data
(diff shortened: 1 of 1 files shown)


system/database/drivers/mysqli/mysqli_driver.php
```diff
@@ -484,6 +499,91 @@ public function field_data($table)
...
+	 */
+	public function list_indexes($table)
+	{
+		// Is there a cached result?
+		if (isset($this->data_cache['table_indexes'][$table]))
+		{
+			return $this->data_cache['table_indexes'][$table];
+		}
+	
+		if (FALSE === ($sql = $this->_list_indexes($table)))
```
tianhe1986: How could it return FALSE since your code is "return 'SHOW INDEXES FROM '.$this->protect_identifiers($table, TRUE, NULL, FALSE)" in function _list_indexes?

system/database/drivers/mysqli/mysqli_driver.php
```diff
@@ -484,6 +499,91 @@ public function field_data($table)
...
+			{
+				if (isset($row['Key_name']))
+				{
+					$key = 'Key_name';
+				}
+				else
+				{
+					// We have no other choice but to just get the third element's key.
+					$temp_array = array_keys($row);
+					$key = $temp[2];
```
tianhe1986: Maybe "$key = $temp_array[2];" ?

pkarunyu: @tianhe1986, yes, `$key = $temp_array[2];` is correct! Thanks for that.

system/database/drivers/mysqli/mysqli_driver.php
pkarunyu: @tianhe1986, I presume you are referring to the line `if (FALSE === ($sql = $this->_list_indexes($table)))`?

If yes, you will note that I tried to follow the convention used for the function `list_fields` in [DB_Driver.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L1246)

tianhe1986: @pkarunyu 
Yes, you have followed the convention used for the function list_fields in DB_Driver.php, 
but there is still a difference.

In DB_Driver.php, `_list_columns` is undefined(maybe it is also a bug, it should be defined and return FALSE),  and the return value depends on the overridden function in subclass, which may return FALSE.

In your code,  you called `_list_indexes` defined in the same class, which does not return FALSE and would rarely be overridden.

In theory, you are right, but I suggest that you define  `_list_indexes` in DB_Driver.php and override it in each subclass.

pkarunyu: @tianhe1986, `list_fields` is defined in [DB_driver.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L1246), but it calls `_list_columns` defined in the individual sub-classes, e.g. [mysqli_driver.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/drivers/mysqli/mysqli_driver.php#L445). 

However, if you look at the function `list_fields` in DB_driver.php, there is this bit...

```php
// Do we know from where to get the column's name?
if ( ! isset($key))
{
	if (isset($row['column_name']))
	{
		$key = 'column_name';
	}
	elseif (isset($row['COLUMN_NAME']))
	{
		$key = 'COLUMN_NAME';
	}
	else
	{
		// We have no other choice but to just get the first element's key.
		$key = key($row);
	}
}
```

Extrapolating that logic to `list_indexes` would mean that we would have a new function in DB_driver called `list_indexes`, which would call a respective `_list_indexes` in the individual sub-classes. And herein is the dilemma.

In MySQL 5.5 which I tested on, the query `SHOW INDEXES FROM sakila.actor` gives the following:

mysql sakila actor show indexes

Note that the column with the definitive name of the index is the third column, named `Key_name`.

For PostgreSQL 9.5 which I also tested on, the query `SELECT * FROM pg_indexes WHERE tablename = 'actor'` gives the following:

postgresql sakila actor show indexes

Note that the column with the definitive name of the index is the third column, named `indexname`.

Based on these two drivers, then the code block I mentioned above could be re-written thus:

```php
// Do we know from where to get the column's name?
if ( ! isset($key))
{
	if (isset($row['Key_name']))
	{
		$key = 'Key_name';
	}
	elseif (isset($row['indexname']))
	{
		$key = 'indexname';
	}
	else
	{
		// We have no other choice but to just get the third element's key.
		$temp_array = array_keys($row);
		$key = $temp_array[2];
		unset($temp_array);
	}
}
```

In other words, we would have an `elseif` for each database driver, and thus far there are 12.

tianhe1986: @pkarunyu 

I understand it.

How about keeping `list_indexes` in DB_driver.php with calling a new function defined in each individual sub-classes to get the name of the index ?

pkarunyu: @tianhe1986, 
> How about keeping list_indexes in DB_driver.php with calling a new function defined in each individual sub-classes to get the name of the index?

That would probably work and seems to be the more elegant way of implementing it.

PR #4758, opened 2016-08-09
URI scheme case

butane:
URI schemes are not case sensitive.


narfbg:
Fair enough, but using `stripos()` is more effective than `strtolower()` and `strpos()`.

butane:
@narfbg Agreed.

PR #4759, opened 2016-08-10
Reloadconfig hook

nispa:
Hello, 
can you have a look to my patch... ;)
I extended the "hooks" so you can add every configuration path, and added reload_config (to call when you add a configuration path).
Thanks for your work and your time!
Nico.

### DIFF
--- a/system/core/Hooks.php
+++ b/system/core/Hooks.php
@@ -97,15 +97,17 @@ class CI_Hooks {
 			return;
 		}
 
-		// Grab the "hooks" definition file.
-		if (file_exists(APPPATH.'config/hooks.php'))
+		foreach ($CFG->_config_paths as $path)
 		{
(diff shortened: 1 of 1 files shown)


narfbg:
Hooks are feature-agnostic, i.e. what happens inside them is not our business - the user decides that.

Also, your second commit goes against our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html), and altering every single non-comment line makes the combined patch very hard to review - don't do that in the future.

nispa:
Ok, sorry... :) 
my IDE realigned every single line...

nispa:
...the correct code is here: https://github.com/bcit-ci/CodeIgniter/pull/4759/commits/d9ec8a7c99904b12d269da76c060b5b7b387fcaf
;)

narfbg:
I did look at the individual commits.

PR #4760, opened 2016-08-10
Accept SQL Functions on SELECT Fields

MarceloRGomes:
When query builder have SQL Functions on SELECT, the explode return error.  This solves the problem.


narfbg:
There's an `$escape` parameter that allows you to override that.

Also, `create_function()` is pretty much forbidden in our code.

MarceloRGomes:
The `$escape` don´t work with array fields.

narfbg:
If so, you're welcome to fix that instead.

PR #4762, opened 2016-08-11
Cache_file: use $data['time'] for calculating expired time.

tianhe1986:
In case of mtime changed and also keeping the same way implemented  in `_get()` , see [Cache_file.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Cache/drivers/Cache_file.php#L277)


narfbg:
Please change line 230 to also check for `$data['time']` then.

tianhe1986:
@narfbg Done.

narfbg:
Um ... `! isset($data['ttl'], $data['time'])`

PR #4764, opened 2016-08-11
URI schemes are not case-sensitive

butane:
#4758 
_Grepping_ the source revealed a few more similar bugs.

Signed-off-by: Vivek Dinesh <vivekdinesh5@gmail.com>

### DIFF
--- a/system/helpers/url_helper.php
+++ b/system/helpers/url_helper.php
@@ -443,7 +443,7 @@ if ( ! function_exists('prep_url'))
 	 */
 	function prep_url($str = '')
 	{
-		if ($str === 'http://' OR $str === '')
+		if (strtolower($str) === 'http://' OR $str === '')
 		{
 			return '';
 		}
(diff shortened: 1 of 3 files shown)


system/helpers/url_helper.php
```diff
@@ -443,7 +443,7 @@ function auto_link($str, $type = 'both', $popup = FALSE)
 	 */
 	function prep_url($str = '')
 	{
-		if ($str === 'http://' OR $str === '')
+		if (strtolower($str) === 'http://' OR $str === '')
```
narfbg: This 'http://' check is useless, if anything - it should be removed.
Same thing with the Form validation method.

butane: What is the expected return value for the function call,
prep_url('http://')

narfbg: Don't care for it - it's obviously bad usage.

butane:
Removed the unwanted checks.

PR #4765, opened 2016-08-12
Fix custom user agent not working

viitols:
Signed-off-by: Kristaps V kristaps@viitols.com

### DIFF
--- a/system/libraries/Email.php
+++ b/system/libraries/Email.php
@@ -1206,6 +1206,7 @@ class CI_Email {
 	 */
 	protected function _build_headers()
 	{
+		$this->set_header('User-Agent', $this->useragent);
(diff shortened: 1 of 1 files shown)


narfbg:
I'm not aware of this header field being part of any email-related protocol.

viitols:
Oh, then it can probably be removed from line 452 of system/libraries/Email.php

`$this->set_header('User-Agent', $this->useragent);`

narfbg:
Actually, will do both.

Neither of `User-Agent` and `X-Mailer` are standards-defined headers, and there's just one [expired IETF RFC draft](https://tools.ietf.org/html/draft-melnikov-email-user-agent-00) attempt to do that, but they do both appear to be de-facto standard headers for the purpose they serve.

Issue #1652, opened 2012-07-24
Add new function replace_batch()


Dentxinho:
We have the functions `insert_batch()` and `update_batch()`, but no `replace_batch()` is provided.
Can `replace_batch()` be created?

Issue #4452, opened 2016-02-11
Replace_batch() function? and Insert_batch() error on large data.


zocheyado:
I noticed that there is no replace_batch() function in the docs, do people not use this often? I pull data from an API regularly and I want to update any new entries that come through the API, OR update existing rows (as the values change quite often). Right now I have been just doing a simple ->replace() in a loop...which is fine except it takes about 20 minutes with the amount of data I have. 

Insert_batch() for example is incredibly fast at inserting rows and I am looking for something similar in replace_batch(). Maybe I am thinking about this wrong but it seems like replace_batch() would be the most common.

As for my issue with insert_batch() it seems to be only inserting the first 100 regardless of how many there really are.

narfbg:
Duplicate of #1652.

zocheyado:
I did read that post first, the reason why I made this post is that you dismissed the idea in the last issue

> ce you've requested it - I'm sure you could, from your point of view. But apparently, so far it's only you that needs such a functionality and your reasoning so far is "Since there's insert_batch() and update_batch(), why no replace_batch()". That doesn't sound like a popular feature request. :)

And I was giving a use case in my example since I knew you were looking for one.

narfbg:
There is a (stupidly presented) use case given there already, although yours does make more sense. Either way, you should've commented there instead of opening a new issue.

But still, you're only the second person in 3+ years to request that feature and demand for it is another point that I've touched in that quote. Ultimately, it was closed because nobody was going to implement it anyway.  
I know I'm not going to, and while I wasn't directly opposed to the idea back then, I kind of am now ... REPLACE should be used with caution; it has some low-level caveats, it's not just a handy alias for DELETE + INSERT.

zocheyado:
Do you care if I implement it? I have never made a pull request but I would not mind doing this. Reading this [article](https://thomashunter.name/blog/mysql-replace-vs-insert-on-duplicate-key-update/) the second option is more what I am looking for and in-fact the default replace() function uses replace into when it should use on duplicate key. I did a rough test by appending ON DUPLICATE KEY UPDATE to the end of an insert_batch() and it worked perfectly, but I wouldn't mind adding it as an actual part of the framework if you are open.

narfbg:
[...]

PR #4768, opened 2016-08-15
Implemented replace_batch() #1652. Definitely plenty of use cases for this.

woodcockjosh:
Instead of making people compile the query and then replace the "INSERT" string it's better to have this feature.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1525,6 +1525,76 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		return 'INSERT INTO '.$table.' ('.implode(', ', $keys).') VALUES '.implode(', ', $values);
 	}
 
+	/**
+	 * Replace_Batch
+	 *
+	 * Compiles batch insert strings replacing any existing rows and runs the queries
+	 *
+	 * @param	string	$table	Table to replace insert into
+	 * @param	array	$set 	An associative array of insert values
+	 * @param	bool	$escape	Whether to escape values and identifiers
+	 * @return	int	Number of rows inserted or FALSE on failure
+	 */
+	public function replace_batch($table, $set = NULL, $escape = NULL, $batch_size = 100)
+	{
+		if ($set === NULL)
+		{
+			if (empty($this->qb_set))
+			{
+				return ($this->db_debug) ? $this->display_error('db_must_use_set') : FALSE;
+			}
+		}
+		else
+		{
+			if (empty($set))
+			{
+				return ($this->db_debug) ? $this->display_error('replace_batch() called with no data') : FALSE;
+			}
+
+			$this->set_insert_batch($set, '', $escape);
+		}
+
+		if (strlen($table) === 0)
+		{
+			if ( ! isset($this->qb_from[0]))
+			{
+				return ($this->db_debug) ? $this->display_error('db_must_set_table') : FALSE;
+			}
+
+			$table = $this->qb_from[0];
+		}
+
+		// Batch this baby
+		$affected_rows = 0;
+		for ($i = 0, $total = count($this->qb_set); $i < $total; $i += $batch_size)
+		{
+			if ($this->query($this->_replace_batch($this->protect_identifiers($table, TRUE, $escape, FALSE), $this->qb_keys, array_slice($this->qb_set, $i, $batch_size))))
+			{
+				$affected_rows += $this->affected_rows();
+			}
+		}
+
+		$this->_reset_write();
+		return $affected_rows;
+	}
+
+	// --------------------------------------------------------------------
+
+	/**
+	 * Insert batch statement
+	 *
+	 * Generates a platform-specific insert string from the supplied data.
+	 *
+	 * @param	string	$table	Table name
+	 * @param	array	$keys	INSERT keys
+	 * @param	array	$values	INSERT values
+	 * @return	string
+	 */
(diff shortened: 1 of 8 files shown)


narfbg:
Already has been suggested and rejected - see not just #1652, but also #4452 - and your reasoning in support of it isn't any better.

woodcockjosh:
Just because you don't have any use case for it doesn't mean there isn't any. I wasn't the first person to open an issue and I won't be the last. Many developers will not even bother and just use a sql workaround which is not required if they have this code aND defeats the entire purpose of active record. I think this feature warrants a larger vote by the community.

One use cases involve inserting many rows of user profile settings into a table where it's inefficient to select the rows then update them one by one. There is no other efficient way to both insert and update multiple rows efficiently without replace, or on conflict update.

Consider also creating multiple shopping cart items that the user clicks save and updates multiple quantities at once, you don't want to have to check each one to see which ones changed or got added. What you want is to insert or update all of them in one query, aka batch replace.

If you're a domain registrar and want to update multiple domain owner data at once including adding ones that don't exist you need this.

I could literally think of thousands of use cases.

There's a reason all good ORM solutions in every language and platform include a batch replace. There's no valid reason to not support this, especially when the query builder class cannot be extended. If you allowed users to extend the query builder I would say there is possibly a reason.

We have insert, replace, batch insert, but no batch replace? It's obviously a missing feature. It doesn't make sense, especially when you have a pull request. Isn't the framework supposed to grow and get better based on the needs of the community? How could this possibily make it worse? If you're concerned about supporting it, I can volunteer to make any required changes. 

I think you're holding something back that the community actually needs, and I'm not really sure how that benefits the any of us.

narfbg:
> Just because you don't have any use case for it doesn't mean there isn't any.

You're basing this on my *very first* [comment on #1652](https://github.com/bcit-ci/CodeIgniter/issues/1652#issuecomment-7299057) 4 years ago. I didn't point you to #4452 for nothing.

And you're twisting my words on top of that.

> I wasn't the first person to open an issue and I won't be the last.

Yep, you're the 3rd in the 5 years since I've been involved with the project. This doesn't mean anything.

> Many developers will not even bother and just use a sql workaround which is not required if they have this code

I don't see a problem with that.

> aND defeats the entire purpose of active record.

We don't have Active Record.

> I think this feature warrants a larger vote by the community.

I don't.

> One use cases involve inserting many rows of user profile settings into a table where it's inefficient to select the rows then update them one by one.

I'd say it is highly likely that you have a poor database design if you need to do this.

> There is no other efficient way to both insert and update multiple rows efficiently without replace, or on conflict update.

You're saying "efficient", but really mean "easy".

> Consider also creating multiple shopping cart items that the user clicks save and updates multiple quantities at once, you don't want to have to check each one to see which ones changed or got added. What you want is to insert or update all of them in one query, aka batch replace.

Fair enough - one half-decent use case.

I'm saying half-decent for a reason though - cart contents are temporary, and as such I don't believe they should be stored in a persistent database in the first place.

> If you're a domain registrar and want to update multiple domain owner data at once including adding ones that don't exist you need this.

You'll have a very hard time convincing me that this ever happens, even more so with CodeIgniter being the engine underneath.

> I could literally think of thousands of use cases.

You're thinking of ways to use it, not of cases that require its usage or where it is overwhelmingly better than an alternative - that's a completely different thing.

> There's a reason all good ORM solutions in every language and platform include a batch replace.

ORMs are a different beast; we don't have an ORM.

> There's no valid reason to not support this, especially when the query builder class cannot be extended.
> If you allowed users to extend the query builder I would say there is possibly a reason.

By your logic so far, there's no valid reason to reject any feature ever. I call bullshit.

Again, please read through #4452, and more specifically [this comment](https://github.com/bcit-ci/CodeIgniter/issues/4452#issuecomment-183088543).

Besides, everything that adds tens (and in this case hundreds) lines of code just to replace a single keyword is bloat in my book.

> We have insert, replace, batch insert, but no batch replace? It's obviously a missing feature.

Features are not collection items.

> It doesn't make sense, especially when you have a pull request.

I appreciate that you did make the effort to actually write the code, which few people do. However, that doesn't mean I must merge it.

Honestly, this expectation that all contributions must be accepted is quite annoying.

> Isn't the framework supposed to grow and get better based on the needs of the community?

You know, I get similar pleas almost every time I reject a feature ...

Yes, it is supposed to do that. It's also supposed to consider the usefulness and safety of proposed features, whether they fit in within its philosophy, guard itself against from feature creep, bloat and technical debt, etc.

What you feel is a "need" for you personally may not be one for the community as a whole, and can very often be exaggerating what is actually a "nice to have sometimes", or poor judgement in the sense that you "need" it only because something else you do makes this seem like an ideal tool (see my "poor database" comment above for an example).

> How could this possibily make it worse?

Lack of (known) negatives doesn't mean a definite positive.

> If you're concerned about supporting it, I can volunteer to make any required changes.

I am also concerned about supporting it, and I would still be a part of my efforts regardless of whether you'd be helping with it or not.

> I think you're holding something back that the community actually needs, and I'm not really sure how that benefits the any of us.

Passive-aggressive remarks like this one don't help your case.

woodcockjosh:
Well, I guess if that's how you want to respond to new contributors I won't try to contribute anymore. I was trying to help because I really appreciate the work you have done on CI so far. I would be ok if you at least made the query builder class extendable. Could you at least consider that? I don't want just open an issue so you can just reject it.

PR #4772, opened 2016-08-16
Support for PHP7

inri13666:
I know that its a hack, but seems for many peoples it should be helpful to update existing applications to support PHP7


narfbg:
CI2 was discontinued last year, stop using it.

PR #4777, opened 2016-08-19
Common: Adding E_PARSE in error judgment.

tianhe1986:
Compile-time parse errors should also be treated in the same way as fatal errors.

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -598,7 +598,7 @@ if ( ! function_exists('_error_handler'))
 	 */
 	function _error_handler($severity, $message, $filepath, $line)
 	{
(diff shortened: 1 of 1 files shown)


narfbg:
Have you tested this? I'm pretty sure `E_PARSE` is not catchable by userspace error handles ... it's a *parser* error after all, not a runtime one.

tianhe1986:
I have tested it, but only with PHP 5.6. The http status code will be 500 after adding E_PARSE and is 200 before adding.

Userspace error handles could not catch this, however,  shutdown handles would do.

The shutdown handle [_shutdown_handler](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/CodeIgniter.php#L137) will call `_error_handler`([see here)](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Common.php#L692) and would finally deal with it.

narfbg:
Makes sense, thanks.

PR #4778, opened 2016-08-21
return fix

antydemant:
return fix


narfbg:
Thanks for the corrections, but please be more descriptive than 2 words next time - that's really vague and anybody searching through issues/pull requests in the future will hate it.

PR #4781, opened 2016-08-23
Hash: processing algorithm name case-insensitively in hash_pbkdf2()

tianhe1986:
Original `hash_pbkdf2()` allows upper-case algorithm name.

### DIFF
--- a/system/core/compat/hash.php
+++ b/system/core/compat/hash.php
@@ -119,8 +119,7 @@ if ( ! function_exists('hash_pbkdf2'))
 	 */
 	function hash_pbkdf2($algo, $password, $salt, $iterations, $length = 0, $raw_output = FALSE)
 	{
-		$algo = strtolower($algo);
-		if ( ! in_array($algo, hash_algos(), TRUE))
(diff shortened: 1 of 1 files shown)


narfbg:
Are you going through every single line of our code?

Not that I don't appreciate your contributions, but if you're going to submit 20 pull requests for one-liners that nobody would ever notice, you might as well submit them all at once.

tianhe1986:
@narfbg OK, I'll submit them at once next time if you accept this way.

I remember that you asked me to commit two changes separately, so I thought maybe I must do like this every time :)

narfbg:
That's the rule in general and I often note newcomers about it as a lot of people try to submit features, bug fixes and styling changes at once.

When you submit tens of patches like this one though, it becomes bureaucracy. :)

narfbg:
Please move the `strtolower()` call inside the `is_array()` check, we don't need that assignment.

tianhe1986:
Oh, I just copy your words and forget to modify it to `in_array()`  :)

PR #4782, opened 2016-08-24
Adding ability to change the language translations directory logic in the application with a function overload

jeremysells:
We need to be able to change the language files location logic in our application. I have split the load function into two parts so the language path logic can be overloaded by MY_Lang.php class

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -117,7 +117,7 @@ class CI_Lang {
 			return;
 		}
 
-		$language_file = $this->get_language_file($langfile, $langfile, $alt_path);
+		$language_file = $this->get_language_file($idiom, $langfile, $alt_path);
 		if ($language_file !== null) {
 			include($language_file);	//Defines $lang
 		} else {
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -117,39 +117,10 @@ class CI_Lang {
 			return;
 		}
 
-		// Load the base file, so any others found can override it
-		$basepath = BASEPATH.'language/'.$idiom.'/'.$langfile;
-		if (($found = file_exists($basepath)) === TRUE)
-		{
-			include($basepath);
-		}
-
-		// Do we have an alternative path to look in?
-		if ($alt_path !== '')
-		{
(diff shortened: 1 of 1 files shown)


narfbg:
Why would you need to do that?
And why would anybody else want it to be a part of the framework, as opposed to your own override?

Also, there are easier ways both to achieve what you want and to change the framework's logic to allow it.

jeremysells:
Hi @narfbg, Thanks for your reply, I want to load the CI language translations directly from the vendor directory as we do not use them for our own custom translations and just want CI to fall-back to using its/the defaults. I would like to avoid copy pasting the official translations into the application folder if that's possible and would like to make the updating easier (hence using composer)

I basically want something like the following to happen (below) I could put this in this pull request but it feels too application specific (in terms of where the translations are located) to go into the framework
```php
 switch ($idiom) {
            case "english":
                $baseLangFolder = ROOT . "vendor/bcit/codeigniter/system/language/";
                break;
            default:
                $baseLangFolder = ROOT . "vendor/bcit/codeigniter3-translations/language/";
                break;
 }

$basepath = $langFolder.$idiom.'/'.$langfile;
if (file_exists($basepath)) {
    return $basepath;
}
```

But that doesn’t really look the best approach to put into the pull request i.e.
* The vendor directory should really be a constant in CI but (from what I can tell CI does not really have much support for the vendor directory)
* "codeigniter3-translations" is a bit specific and I think it should be complimentary to the application name (no version) i.e. "codeigniter-translations" but that differs from the repo name. I could always pull in the additional translations repo as a "vendor/bcit/codeigniter-translations" and add that to this pull request as a file_exists() but that also does not seem like a good idea as its not very consistent.

What do you think?

jim-parry:
The translations need to be version specific. It is not clear exactly how they will be handled in CI4, but it is likely to be very different, more along the lines of PHP internationalization and standard language specifiers (eg EN-US vs english)

narfbg:
The method has an `$alt_path` parameter, you could use `add_package_path()`, you could create a symbolic link ... it's not like there aren't options available.

PR #4784, opened 2016-08-25
Input: sanitizing optimization

tianhe1986:
I did three things.
1. When `_clean_input_keys()` returns `FALSE`, just discard this key.
   If I use '`0=ok&1=gogogo&<*:887=akb48`' as query string and then call `var_dump($_GET)`, I'll get 

```
`array (size=3)
  0 => string 'akb48' (length=5)
  1 => string 'gogogo' (length=6)
  '<*:887' => string 'akb48' (length=5)`
```

However, `$_GET[0]` is expected to be 'ok'. And, I do not confirm whether or not right, but I prefer that uncleaned key be removed from `$_GET` and `$_POST`.
2. Calling `_clean_input_data()` to sanitize `$_GET` , `$_POST` and `$_COOKIE`, basing on the above modification.
3.  Just a one-liner change, keep consist with the meaning of 'fatal'.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -674,7 +674,10 @@ class CI_Input {
 			$new_array = array();
 			foreach (array_keys($str) as $key)
 			{
-				$new_array[$this->_clean_input_keys($key)] = $this->_clean_input_data($str[$key]);
+				if(($cleaned_key = $this->_clean_input_keys($key)) !== FALSE)
+				{
+					$new_array[$cleaned_key] = $this->_clean_input_data($str[$key]);
+				}
 			}
 			return $new_array;
 		}
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -706,11 +706,11 @@ class CI_Input {
 	 *				key is encountered
 	 * @return	string|bool
 	 */
-	protected function _clean_input_keys($str, $fatal = TRUE)
+	protected function _clean_input_keys($str, $fatal = FALSE)
 	{
 		if ( ! preg_match('/^[a-z0-9:_\/|-]+$/i', $str))
 		{
-			if ($fatal === TRUE)
+			if ($fatal === FALSE)
 			{
 				return FALSE;
 			}
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -706,11 +706,11 @@ class CI_Input {
 	 *				key is encountered
 	 * @return	string|bool
 	 */
-	protected function _clean_input_keys($str, $fatal = FALSE)
+	protected function _clean_input_keys($str, $fatal = TRUE)
 	{
 		if ( ! preg_match('/^[a-z0-9:_\/|-]+$/i', $str))
 		{
-			if ($fatal === FALSE)
+			if ($fatal === TRUE)
 			{
 				return FALSE;
 			}
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -606,22 +606,13 @@ class CI_Input {
 		{
 			$_GET = array();
 		}
-		elseif (is_array($_GET))
+		else
 		{
-			foreach ($_GET as $key => $val)
-			{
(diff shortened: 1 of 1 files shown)


narfbg:
... these are not simple one-liners.

1. I don't understand what you're talking about here.
2. Change for the sake of change.
3. Change for the sake of change, on dead code.

4. You're not complying with our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).

tianhe1986:
1. The value of `$_GET['0']` will be overridden by the value of `$_GET['<*:887']` since `_clean_input_keys` returns  `FALSE`  with key '<*:887' and then set `$_GET[FALSE] = XXX`

2 Yes, but calling existed function is better than repeating the same code with just variables name change.

3 OK, it is indeed dead code :)

4 Which one do you mean? 
`is_array($_GET) && $_GET = $this->_clean_input_data($_GET);`
or
`if(($cleaned_key = $this->_clean_input_keys($key)) !== FALSE)`

narfbg:
2. Better by what metric? And even if so, you're not even keeping the same logic.
3. Revert it.
4. The second one. But really, you should get yourself familiar with the styleguide as opposed to me having to explain it.

tianhe1986:
2 Yes, it do not keep the same logic, but it bases on the first commit and keeps the same logic with the first commit. If I do not change to call `_clean_input_data()`, then maybe I need change it as follows.
```
if (is_array($_POST))	
{		
	foreach ($_POST as $key => $val)		
	{
		if (($cleaned_key = $this->_clean_input_keys($key)) !== FALSE)
		{
			$_POST[$cleaned_key] = $this->_clean_input_data($val);	
		}		
 	}		
 }
```
And also for `$_GET`.

3 and 4, I'll modify it.

narfbg:
I'm not arguing that your patch is inconsistent, that's not the point.

The problem is that you're making a BC break.

tianhe1986:
So, what's your advice ?

narfbg:
I don't really have any advice here ...

tianhe1986:
@narfbg 
Ah, the problem is the value changed after my patch.
I'll take `$_GET` for example, list its structure before and after sanitizing. 

Suppose the original `$_GET` is
```
$_GET = array(
	'0' => 'Jay Chou',
	'1' => array(
		'0' => 'Jacky Cheung',
		'1' => 'Louis Koo',
		'>*;2' => 'Chow Yun Fat'
	),
	'a' => 'Andy Lau',
	'<*:8' => 'Aaron Kwok',
);
```

Without my patch, after sanitizing it will be 
```
$_GET = array( 
	'0' => 'Aaron Kwok',
	'1' => array(
		'0' => 'Chow Yun Fat',
		'1' => 'Louis Koo',
	),
	'a' => 'Andy Lau',
	'<*:8]' => 'Aaron Kwok',
);
```

With my patch, after sanitizing it will be
```
$_GET = array( 
	'0' => 'Jay Chou',
	'1' => array(
		'0' => 'Jacky Cheung',
		'1' => 'Louis Koo',
	),
	'a' => 'Andy Lau',
);
```

So, what's the right expected value? I'll try to reach it.

narfbg:
I'm not sure we want to keep this input sanitization feature at all. In fact, I'm leaning towards removing it (and I have thought about this for a long time).

I don't want to kill your enthusiasm, but this decision doesn't have to be taken right now.

tianhe1986:
OK, I get it. I also prefer to sanitize input by myself in controller, but I saw it does something in `CI_Input`.
So, should I close this pull request?

narfbg:
Not necessarily, there's no harm in leaving it open. As I said, the decision doesn't have to be taken right now.

narfbg:
Dropped `_sanitize_globals()` in ff773c059cb984920767dd6187c30a77e5bf78c9

PR #4786, opened 2016-08-29
Removed URI_TO_ASSOC "caching"

renedekat:
When using the uri_to_assoc call multiple times within 1 request, with multiple default values for each call, only the default values of the first call are used. All subsequent calls will have used the cached values.

It might be useful to call this multiple times. For example, when you want to do a general check in your __construct for a param that always has to be set, but the controller methods might expect different parameters.

Removing this IF statement, will solve that issue.

Example:

```
public function __construct()
{
    $default = ['type'];
    $uri_params = $this->uri->uri_to_assoc(3, $default);
}

public function view()
{
    $default = ['type', 'sort_order'];  // Sort_order can be optional 
    $uri_params = $this->uri->uri_to_assoc(3, $default);
    $items = $some_object->get($uri_params['type'], $uri_params['sort_order']);
}
```
This will fail if sort_order isn't passed in the URI, as the default settings will be ignored in the uri_to_assoc call, because of the first call in the constructor.

This PR, will fix that.

### DIFF
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -450,11 +450,6 @@ class CI_URI {
 			return $default;
 		}
 
-		if (isset($this->keyval[$which], $this->keyval[$which][$n]))
-		{
-			return $this->keyval[$which][$n];
-		}
-
 		$total_segments = "total_{$which}s";
 		$segment_array = "{$which}_array";
 
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -450,6 +450,11 @@ class CI_URI {
 			return $default;
 		}
 
+		if (isset($this->keyval[$which], $this->keyval[$which][$n]))
+		{
+			return $this->keyval[$which][$n];
+		}
+
 		$total_segments = "total_{$which}s";
 		$segment_array = "{$which}_array";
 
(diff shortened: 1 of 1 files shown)


renedekat:
How can this change break a MySQLi test??

narfbg:
I'd argue that the provided example is bad, but regardless of that, the caching itself can be put inside an `else` condition [here](https://github.com/bcit-ci/CodeIgniter/blob/02fde67322db74df42ef8dc953cc2be3f9fc4ccd/system/core/URI.php#L496).

renedekat:
@narfbg Could you explain what you mean? I don't think that will solve anything if you provide the $default array for each call.

narfbg:
If the defaults are not cached - they won't be returned.

renedekat:
What is your proposed else? That'll make it easier for me to understand what you mean.

narfbg:
... look at the line I linked to.

It's the closing brace for an `if (count($default))` condition, and just below it is where the caching happens. Make it so that defaults aren't cached, by putting an `else` there.

renedekat:
That's what I thought you meant, but making sure.
Check

narfbg:
Wait, did you put the ENTIRE pull request description in your commit message?!

narfbg:
I've erased these commits. Please resubmit.

PR #4789, opened 2016-08-30
Bug fiexd: _excute() in Form_validation.php

YOUKI-T:
If run the code such as the following, set_message() does not work correctly.

`$this->form_validation->set_rules('password', 'Password', array(
           array($this->User_model, 'is_match_password')
));`

Please check it out.


narfbg:
Yes, if you use it incorrectly, it won't work.

https://codeigniter.com/userguide3/libraries/form_validation.html#callable-use-anything-as-a-rule

PR #4790, opened 2016-08-31
Enable/ disable PDO MySQL native types in results

renedekat:
Added option to enable/ disable returning native MySQL column types.
The option is disabled by default, so it will not change the default behaviour.

This can be enabled in config/database.php by setting the subdriver to 'mysql' and use_native_types to TRUE.
DO NOT USE THE DSN!

Example:
```
        'dsn' => NULL,
	'hostname' => 'localhost',
	'username' => 'YOUR_USERNAME',
	'password' => 'YOUR_PASSWORD',
	'database' => 'YOUR_DATABASE',
	'dbdriver' => 'pdo',
	'subdriver' => 'mysql',
	'use_native_types' => TRUE,
```

### DIFF
--- a/system/core/compat/hash.php
+++ b/system/core/compat/hash.php
@@ -119,7 +119,7 @@ if ( ! function_exists('hash_pbkdf2'))
 	 */
 	function hash_pbkdf2($algo, $password, $salt, $iterations, $length = 0, $raw_output = FALSE)
 	{
-		if ( ! in_array($algo, hash_algos(), TRUE))
+		if ( ! in_array(strtolower($algo), hash_algos(), TRUE))
 		{
 			trigger_error('hash_pbkdf2(): Unknown hashing algorithm: '.$algo, E_USER_WARNING);
 			return FALSE;
--- a/system/database/drivers/pdo/subdrivers/pdo_mysql_driver.php
+++ b/system/database/drivers/pdo/subdrivers/pdo_mysql_driver.php
@@ -83,6 +83,15 @@ class CI_DB_pdo_mysql_driver extends CI_DB_pdo_driver {
 	 * @var	string
 	 */
 	protected $_escape_char = '`';
+	
+        /**
+         * Determines whether results should return the native column type.
+         * By default this is disabled to be BC. Set this to true in config/database.php
+         * to let integers be integers, NULLS be NULLS etc
+         *
+         * @var bool
+         */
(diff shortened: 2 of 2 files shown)


narfbg:
We don't accept MySQL-only features.

renedekat:
May I ask why not? Using native types has a lot of benefits, which you are withholding now for your users.

renedekat:
I have an improvement ready for this that will pass the options in before the connect, instead of setting them afterwards.

narfbg:
> May I ask why not?

Because we aim for abstraction, not just wrapping one interface in another.

> Using native types has a lot of benefits, which you are withholding now for your users.

No, we are not. The PDO object is publicly accessible.

A feature doesn't have to be specifically built into CI for you to use it.

> And what about this??
>
> ```
		if (is_array($this->encrypt) && is_php('5.3.7'))
		{
			$ssl = array();
			empty($this->encrypt['ssl_key'])    OR $ssl[PDO::MYSQL_ATTR_SSL_KEY]    = $this->encrypt['ssl_key'];
			empty($this->encrypt['ssl_cert'])   OR $ssl[PDO::MYSQL_ATTR_SSL_CERT]   = $this->encrypt['ssl_cert'];
			empty($this->encrypt['ssl_ca'])     OR $ssl[PDO::MYSQL_ATTR_SSL_CA]     = $this->encrypt['ssl_ca'];
			empty($this->encrypt['ssl_capath']) OR $ssl[PDO::MYSQL_ATTR_SSL_CAPATH] = $this->encrypt['ssl_capath'];
			empty($this->encrypt['ssl_cipher']) OR $ssl[PDO::MYSQL_ATTR_SSL_CIPHER] = $this->encrypt['ssl_cipher'];
>
			// DO NOT use array_merge() here!
			// It re-indexes numeric keys and the PDO_MYSQL_ATTR_SSL_* constants are integers.
			empty($ssl) OR $this->options += $ssl;
		}
```

PostgreSQL also has some sort of support for encrypted connections, and we're expecting more in the future.

Either way, this is not an argument in favor of your feature.

> All the PDO sub @narfbg drivers have some specific code for that driver. 

I don't understand what you're saying here.

> And I don't see this in any other driver? Specific for MySQL??
>
> `if (isset($this->stricton))`

Yes, this is a MySQL-specific feature.

It's also a legacy feature that existed before I got involved with the project, and it wouldn't be accepted today.

PR #4800, opened 2016-09-08
Let choose if needed to escape identifier in "from" of database query builder

Naruto-kyun:
Allow from example to use NATURAL JOIN ! (and dont break compatibility )

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -520,7 +520,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		{
 			$type = strtoupper(trim($type));
 
-			if ( ! in_array($type, array('LEFT', 'RIGHT', 'OUTER', 'INNER', 'LEFT OUTER', 'RIGHT OUTER'), TRUE))
+			if ( ! in_array($type, array('LEFT', 'RIGHT', 'OUTER', 'INNER', 'LEFT OUTER', 'RIGHT OUTER', 'NATURAL'), TRUE))
 			{
 				$type = '';
 			}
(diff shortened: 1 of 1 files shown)


narfbg:
This method is meant to accept table names, not arbitrary SQL keywords. If you want to add something for JOINs, you should do it in `join()`.

mwhitneysdsu:
I know that query builder isn't intended to handle complex queries. However, a SQL from clause is meant to accept a table name, aliases for the table and its columns, a table sub-query (and associated aliases), and even joined tables. (In fact, though I think joins are a terrible reason to add this to the from clause, the NATURAL JOIN keywords are explicitly permitted in the standard.)

In other words, you can cram a lot of arbitrary SQL keywords into a from clause and have it work on almost any database which claims any significant level of support for the standard. This makes the from clause one of the most useful methods in query builder in which to disable the escape functionality.

I can think of a couple of reasons for not doing this:
- it's one of the more dangerous places to forego escaping identifiers
- stacking new arguments on methods is terrible for the API
- any reason anyone comes up with to justify doing it is probably not a good enough reason to forego the previous two
- there are probably a few previous issues requesting this which have been shot down

So, this is my extra-long version of adding a +1 to this request, but not really believing that I can change your mind on it.

Naruto-kyun:
Ok, I've made some change.
join now accept NATURAL JOIN with this usage: join("TABLE_NAME", '', "NATURAL")
but to respect the others functions which most of them have a disable option for the escaping, I would like to keep the first commit. Someone could have an use of it for some dark reason one day.

narfbg:
@mwhitneysdsu It accepts aliases already and we have `join()` for JOINs, so ... yea, hard to change my mind on it.

Not impossible, but so far nobody has presented good reasoning for it (indeed, it has been requested multiple times before).

@Naruto-kyun Don't just cram your `join()` additions to your original `from()` patch ... we don't want the `from()` additions.

Also, you'll need to read our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html), as your code currently does not comply with it.

PR #4801, opened 2016-09-08
supplied flag to turn off mysqli SSL verification if ssl_verify passed as false

intekhabrizvi:
As per http://php.net/manual/en/mysqli.real-connect.php its require to pass MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT flag if we don't want to verify SSL.

Current code is failing to connect mysql server securely without verifying the certificate despite of supplying false in database.php [encrypt][ssl_verify] array.


narfbg:
Skipping SSL isn't a solution to this.

intekhabrizvi:
You got it completely wrong, i am not saying to skip the SSL, i am saying its a only way to tell mysql to stop ssl verification via php.
The current code have missing flag necessary to do that. Pls ref the official php user guide link i supplied.
Right now the code written to do this is half bake. pls review the codes again thanks

intekhabrizvi:
Sorry, got it, will supply another patch to fix this issue.

PR #4803, opened 2016-09-09
quick fix error insert_batch using PHP7

iam-adty:
add reset($key) on line 1548

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1546,6 +1546,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 
 		is_bool($escape) OR $escape = $this->_protect_identifiers;
 
+		reset($key);
 		$keys = array_keys($this->_object_to_array(current($key)));
 		sort($keys);
 


narfbg:
And what does that fix?

iam-adty:
to fix this issue [https://github.com/bcit-ci/CodeIgniter/issues/4804](url)

narfbg:
Well ... in the future, please don't open separate issue and pull request.

I'm sure you can see how that's problematic - you have a pull request here with no description or context at all, and thus - no way of knowing why you want this change. And even if you did initially link to #4804, we'd still have to click around between two separate issues to understand what's going on.

pepegarciag:
Well i just going to leave here my comment instead in the related issue(#4804).

I had the same issue working with PHP 7.0.8, if I change to PHP 5.5.26 I have no errors.

If I create a static array like the one I'm using there is no problem (my array was created dinamically) so instead of changing CI core, i just reset my array pointer on my code and all the errors went out.

iam-adty:
@pepegarciag Yap. I just reset all of my arrays before throwing them to every functions of CI.

Case closed. Thanks everyone.

PR #4805, opened 2016-09-09
supplied flag to turn off mysqli SSL verification if ssl_verify passed as false

intekhabrizvi:
Updated codes to pass MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT flag

### DIFF
--- a/system/database/drivers/mysqli/mysqli_driver.php
+++ b/system/database/drivers/mysqli/mysqli_driver.php
@@ -183,7 +183,6 @@ class CI_DB_mysqli_driver extends CI_DB {
 					// https://bugs.php.net/bug.php?id=68344
 					elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT'))
 					{
-						$this->_mysqli->options(MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT, TRUE);
 						$client_flags |= MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
 					}
 				}
--- a/system/database/drivers/mysqli/mysqli_driver.php
+++ b/system/database/drivers/mysqli/mysqli_driver.php
@@ -184,6 +184,7 @@ class CI_DB_mysqli_driver extends CI_DB {
 					elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT'))
 					{
 						$this->_mysqli->options(MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT, TRUE);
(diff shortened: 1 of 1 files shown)


narfbg:
If this is a connect flag, do we still need the `options()` call?

And more importantly, while #4801 disabled SSL altogether (on PHP < 5.6.16), it did take into account what the [manual](https://secure.php.net/manual/en/mysqli.real-connect.php) says here:

> **MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT** - Like **MYSQLI_CLIENT_SSL**, but disables validation of the provided SSL certificate. This is only for installations using MySQL Native Driver and MySQL 5.6 or later.

Isn't it therefore better not to set ``MYSQLI_CLIENT_SSL`` *only if* this flag here is set?

intekhabrizvi:
I though the same, we are already using using `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` in `options` function but its not taking any effect hence i cross check PHP C source codes and found `options` function have no definition of MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT check line no 1733 of http://git.php.net/?p=php-src.git;a=blob;f=ext/mysqli/mysqli_api.c;hb=6baaa2bb1b0ec87d8c2d676da27d04370677a81c you can see definition of `MYSQLI_OPT_SSL_VERIFY_SERVER_CERT` but no definition of `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` hence no matter you write or not its not going to work.

Regards to `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT - Like MYSQLI_CLIENT_SSL` due to poor explanation of flag behavior, at first i thought may be writing `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` only also do the job of `MYSQLI_CLIENT_SSL` as well but i was wrong, as per PHP C source codes test case file http://git.php.net/?p=php-src.git;a=blob;f=ext/mysqli/tests/bug51647.phpt;hb=6baaa2bb1b0ec87d8c2d676da27d04370677a81c#l44 on line no `44` you can see they are using both flag with pipe hence both flag have individual work and both are necessary.

Seems i need to apply for another patch for this.
Thanks

narfbg:
You don't need to open a new pull request, just push another commit (removing the `options()` call) to the same branch and this one will be automatically updated.

intekhabrizvi:
Shall i comment that line or completely remove it, I am thinking may be in future they add those feature in PHP source codes so that we can start using options function again.

narfbg:
Remove it, we don't keep commented-out code.

intekhabrizvi:
Updated code pushed, pls check.

PR #4807, opened 2016-09-13
new upload method with a new name

castrohs:
i create a method to set a net name of file.
Use like this:
if ($this->upload->do_upload_with_name('image1', $new_name_image)) {
        }


narfbg:
No thanks.

Issue #96, opened 2011-08-19
issue with pages having multiple forms - form_helper


bitbucket-import:
i've created a page with two forms

when I submit one form, the radio buttons in the other form are not checked (checked=checked) even if the third parameter of set_radio() is TRUE. (no validations set for both forms)

I removed "count($_POST) === 0" from line number 818 of form_helper.php and the problem solved

https://bitbucket.org/ellislab/codeigniter/src/8af0fb079f90/system/helpers/form_helper.php

PR #4810, opened 2016-09-16
Remove inline style from hidden fields

Dutchy-:
I propose to remove the inline style "display:none" for hidden form fields.

1. I know of no browser that actually displays hidden form fields, so setting the style is redundant.
2. I can not imagine situations where developers rely on this style being present.
3. It generates an error when inline styles are disabled in a CSP header.

3 is my main reason for removing the style.

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -100,7 +100,7 @@ if ( ! function_exists('form_open'))
 		{
 			foreach ($hidden as $name => $value)
 			{
-				$form .= '<input type="hidden" name="'.$name.'" value="'.html_escape($value).'" style="display:none;" />'."\n";
+				$form .= '<input type="hidden" name="'.$name.'" value="'.html_escape($value).'" />'."\n";
 			}
 		}
 


narfbg:
1. By default they don't, but that's due to the default CSS they have for different elements. Once you change e.g. all input tags to have `display:block`, it can cause problems - I've experienced this personally.
2. I can, read point 1.
3. What kind of errors.
4. `#<number>` indicates an issue number - you've now linked this issue to a 5-year old one about database transactions.

Dutchy-:
These kinds of errors:

screen shot 2016-09-16 at 10 45 56

If you do not know how the CSP header works, I can suggest some resources for more information.

An alternative solution would be to use a class, but codeigniter does not come with any css files.
The errors are in the console only and do not actually break the form, but I believe it would be nice if they could be fixed in CI.


And my apologies for linking an old issue.

narfbg:
I don't think modifying code only so that it doesn't trigger a console error message outweighs solving the problem that I've explained in point 1.

Dutchy-:
I personally don't agree, but I can see your point.

Is it an idea to save this change for a release that already breaks backwards compatibility? Do you have any other suggestions?

narfbg:
If I don't like the change now, why would I want to save it for later? :)

I don't have other suggestions, but waiting for you to come up with such is why I haven't closed the issue, yet.

Dutchy-:
One last thing; when using the [report-uri](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/Using_CSP_violation_reports) functionality of CSP, each instance of the inline style will generate a report to that URL, potentially drowning out more interesting reports. 

Does this change the importance of this change for you? If not, close this (I will try to maintain my own patch locally).

narfbg:
Not sure about that.

But I'd suggest you take another approach ... you're trying to convince me to create a regression, that's never going to work because I'm supposed to solve problems and not create them. If this is so important to you, why not try to solve both problems, so that everybody is happy?

Dutchy-:
I made it configurable, does this work for you?

narfbg:
A global config option for a single HTML attribute in an optional helper? No ...

The function already has an `$attributes` parameter - you can use that to make it optional somehow.

mwhitneysdsu:
> The function already has an $attributes parameter - you can use that to make it optional somehow.

Since the `$attributes` parameter is applied to the form element, there's always going to be the chance that whatever key you choose to define this behavior will match an attribute name which someone was using because some JavaScript library which ignores the standards told them to (Angular comes to mind, but it's certainly not the only one). This means that adding configuration to this parameter would be a BC break.

The next logical place would be the `$hidden` parameter, but it's defined as an array of name/value pairs and would probably be more likely to run into key collisions if you specify a key for configuration information. So, again, this would require a BC break.

This leaves adding a parameter or some external configuration option.

Additionally, the style is not set in the `form_hidden()` function (and I'm sure this isn't the only reason the `form_open()` function isn't calling the `form_hidden()` function to add the hidden input(s)).

> By default they don't, but that's due to the default CSS they have for different elements. Once you change e.g. all input tags to have display:block, it can cause problems - I've experienced this personally.

For hidden inputs, the HTML5 specification doesn't allow a control, user interaction, height/width, and many other properties which would be required if you wanted to display the field. The only attribute you can change that should make it show up is the type attribute itself.

In HTML 4.01, a hidden input is supposed to generate a hidden control, which is not supposed to be rendered (the difference here is basically that the HTML5 spec is more explicit in not allowing a hidden input to be rendered for the user, but the end result should be the same).

In other words, this has to be a work-around for a bug in a browser, and it would be helpful to know the situation in which this is actually needed, so it could be tested, we can make sure the bug is reported, people can make an informed decision about whether to override the function to remove the work-around in their project, etc.

I've tested up-to-date versions of Chrome, Safari, and Firefox on OS X, and IE and Edge on Windows 10 with a form (on a page using `<!DOCTYPE html>`) on which the browsers' debuggers resolved the display property of the hidden inputs to "inline-block," and none of them displayed the inputs (in the cases where the built-in CSS was shown in the debugger, it was overridden).

narfbg:
Fair enough.

The `display:none:` isn't there for no reason though ...

Dutchy-:
Thanks for your input @mwhitneysdsu .

I've done some digging and found the first change to an extra hidden style at https://github.com/bcit-ci/CodeIgniter/commit/fd2ba8836d74ded06cfe14f728387698c0532148

Sadly, the change does not specify why the html failed to validate without that div. #96 doesn't look all that relevant, maybe there was a bitbucket import mismatch somewhere?

The change that actually moved the style to the input itself was done by @narfbg 2 years ago at https://github.com/bcit-ci/CodeIgniter/commit/93b4e788796d6da181512702d6431ffd7e199bee

After reading #2916 I'm still not able to understand how the style is still a solution to the original problem. Without further understanding, the last thing I can do is submit a patch for an extra parameter to form_open to disable the inline style.

narfbg:
It's not an import mismatch per se - the import itself only does commits, doesn't transfer over issues.

Anyway, it does seem like the `style` was necessary for `div`s, but is irrelevant for `input`s, so I guess I'll accept your patch. But it also looks like you've overwritten your original commit, so I can't cherry-pick it ...

Dutchy-:
I have redone the commit.

narfbg:
:+1:

PR #4813, opened 2016-09-18
add X-Requested-With as potential AJAX Header

tobiasfichtner:
X-Requested-With is mainly used to identify Ajax requests. Most JavaScript frameworks send this field with value of XMLHttpRequest


narfbg:
... that's what the existing code checks for.

PR #4820, opened 2016-09-25
Captcha filenames should be sha1 of timestamp and word 

gxgpet:
Hi,

Currently, captcha helper uses microtime function output as filename. I think we can find a better approach to this, so there is no logic exposed to users.

One of them, which is this pull request, is to save the captcha filenames as a SHA1 of concatenation between timestamp (and not micro-timestamp - see below) and the word itself. 

Based on that, usage of microtime() is no longer needed here; simple `int`s generated by time() should be enough, as the deletion from this pull request uses [`filemtime`](http://php.net/manual/en/function.filemtime.php) function to generate the file timestamp of creation, and by this way we will know which files are "expired" and can be deleted. Also, let's be honest: microseconds aren't the most realistic approach when talking about captcha files expiration.

This pull request also solves a problem where deletion of old files didn't occur on png files, since there were checked only jpg-ending files, and the captcha can generate png files too, if jpg generation is not available.

### DIFF
--- a/system/helpers/captcha_helper.php
+++ b/system/helpers/captcha_helper.php
@@ -105,12 +105,13 @@ if ( ! function_exists('create_captcha'))
 		// Remove old images
 		// -----------------------------------
 
-		$now = microtime(TRUE);
+		$now = time();
 
 		$current_dir = @opendir($img_path);
 		while ($filename = @readdir($current_dir))
 		{
-			if (substr($filename, -4) === '.jpg' && (str_replace('.jpg', '', $filename) + $expiration) < $now)
+            if (in_array(substr($filename, -4), array('.jpg', '.png'))
+                && (filemtime($img_path.$filename) + $expiration) < $now)
 			{
 				@unlink($img_path.$filename);
 			}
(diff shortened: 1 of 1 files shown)


narfbg:
I don't see how the current behavior is bad, nor how this is an improvement.

gxgpet:
Except for the fix of file deletion (searching for png files too), when exposing a numeric-named file to users, they may try to find other files as well by changing those numbers. Also saving information which is available from machine filesystem, in the filename, is not desired too, I think.

I didn't say the current behavior is bad, but only that there may be a better option to it :)

narfbg:
I'd accept a fix for the png files deletion thing that you've found, but the rest is your personal preference rather than objectively an improvement.

gxgpet:
Very well, then.

I will close this pull request and make another one during today evening.

Regards,

narfbg:
Before you do, please go through our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html) first. Your patch here didn't comply with it.

PR #4822, opened 2016-09-27
Fix file deletion on captcha helper

gxgpet:
At this point, captcha helper doesn't remove .png files (expired files) if the imagepng was used at captcha creation (and not jpg).

This pull request solves this issue.


narfbg:
... download helper changes?

I see you've had quite the mixup with branches.

gxgpet:
No, I just failed switching a branch. But I fixed it by reverting. :)

Sorry about it.

PR #4823, opened 2016-09-30
Fix session crashes with UTF-8 characters

fubbyb:
Bug fix: endless for loop when tryinig read session data from file with UTF-8 characters.

### DIFF
--- a/system/libraries/Session/drivers/Session_files_driver.php
+++ b/system/libraries/Session/drivers/Session_files_driver.php
@@ -189,7 +189,7 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 		$session_data = '';
 		for ($read = 0, $length = filesize($this->_file_path.$session_id); $read < $length; $read += strlen($buffer))
 		{
-			if (($buffer = fread($this->_file_handle, $length - $read)) === FALSE)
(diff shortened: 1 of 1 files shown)


narfbg:
What's your `mbstring.func_overload` ini value?

fubbyb:
My mbstring.func_overload set to "2". But my hosting provider doesn't allow to change this value separately for each site. So, one little change allow CI works fine with any func_overload. One more question is why you doesn't use file_get_contents function?

gxgpet:
The two main reasons that `file_get_contents` is not used on file session driver are that (1) a file handle is required to be mainly open during a request as a class member and (2) file locking mechanism.

narfbg:
I asked that to confirm my suspicion that mbstring is causing the issue, as `feof()` is a work-around here and renders a lot of the other logic useless. I've got another solution in mind ...

But if this is a configuration set by your hosting provider - I suggest that you switch away from them as soon as possible.

fubbyb:
In my project I totally rewrite this part of code:
`$session_data = '';
$length = filesize($this->_file_path.$session_id);

while (!feof($this->_file_handle)) 
{
	$position = ftell($this->_file_handle);
	if (($buffer = fread($this->_file_handle, $length - $position + 1)) === FALSE) break;
	
	$session_data .= $buffer;
}`

PR #4825, opened 2016-09-30
Download helper improvement on reading phase

gxgpet:
CodeIgniter uses the classic combination of `fopen` and `fread` functions to achieve the reading of a file, if a path is provided to `force_download` helper function. 

The only problem about this combination is that `fread` uses a hardcoded value for chunk size: 1048576 (1MB). This may be a memory usage problem on high-load applications developed with CodeIgniter (even on a dedicated server), and can be avoided by using PHP's function [`readfile`](php.net/manual/en/function.readfile.php), which is handling the reading according to server status. As you can see from official docs, it shouldn't create any memory problems, even on big files.

### DIFF
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -121,11 +121,6 @@ if ( ! function_exists('force_download'))
 			$filename = implode('.', $x);
 		}
 
-		if ($data === NULL && ($fp = @fopen($filepath, 'rb')) === FALSE)
-		{
-			return;
-		}
-
 		// Clean output buffer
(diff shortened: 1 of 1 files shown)


narfbg:
We've hard-coded the 1MB value because it's small enough ... If that's enough to push above your memory usage limit, I'd say your problem is deffinately somewhere else.

And you're not even describing a problem here - "may be a memory usage problem" sounds highly theoretical.

narfbg:
Fair enough, I see no downside.

Issue #4816, opened 2016-09-20
Table Libraries clear()


784238ddda92fee476352f3ccaa6cca2:
Hi!.
Library system/Table is call method clear() but not clear $this->caption variable.

gxgpet:
Yes, it appears to be a small bug.

Meanwhile a fix is released, use `->set_captation(NULL)` after `clear()`.

narfbg:
**If** we agree that this is a bug, which I'm not convinced it is.

Your own pull request - #4826 - alters the documentation to say that `clear()` affects captions, which it currently doesn't.

PR #4826, opened 2016-10-01
Small fix for HTML Table library

gxgpet:
- when calling `clear()`, caption should be cleared too;
- method chaining wasn't working when calling `set_caption()` method;
- updated docs for these changes.

### DIFF
--- a/user_guide_src/source/libraries/table.rst
+++ b/user_guide_src/source/libraries/table.rst
@@ -275,7 +275,7 @@ Class Reference
 		:returns:	CI_Table instance (method chaining)
 		:rtype:	CI_Table
 
-		Lets you clear the table heading, row data, and the caption. If you need to show multiple tables with different data you should to call this method
+		Lets you clear the table heading, row data and the caption. If you need to show multiple tables with different data you should to call this method
 		after each table has been generated to clear the previous table information. Example::
 
 			$this->load->library('table');
--- a/user_guide_src/source/libraries/table.rst
+++ b/user_guide_src/source/libraries/table.rst
@@ -275,11 +275,12 @@ Class Reference
 		:returns:	CI_Table instance (method chaining)
 		:rtype:	CI_Table
 
-		Lets you clear the table heading and row data. If you need to show multiple tables with different data you should to call this method
+		Lets you clear the table heading, row data, and the caption. If you need to show multiple tables with different data you should to call this method
 		after each table has been generated to clear the previous table information. Example::
 
 			$this->load->library('table');
 
+			$this->table->set_caption('Preferences');
 			$this->table->set_heading('Name', 'Color', 'Size');
 			$this->table->add_row('Fred', 'Blue', 'Small');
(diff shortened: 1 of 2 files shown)


narfbg:
In addition to my comment here - https://github.com/bcit-ci/CodeIgniter/issues/4816#issuecomment-251051151, we generally don't accept changes of multiple functionalities at once. Method chaining has nothing to do with this.

gxgpet:
Agree. Next time I will make a PR for it.

But as for now, the devs _don't have to_ really know how to reset the caption, let's say, that we want them to use set_caption again for resetting. In this case, should they use `set_caption(FALSE)`, `set_caption('')` or `set_caption(NULL)`? That is my point.

user_guide_src/source/libraries/table.rst
```diff
@@ -275,11 +275,12 @@ Class Reference
 		:returns:	CI_Table instance (method chaining)
 		:rtype:	CI_Table
 
-		Lets you clear the table heading and row data. If you need to show multiple tables with different data you should to call this method
+		Lets you clear the table heading, row data, and the caption. If you need to show multiple tables with different data you should to call this method
```
ValentinNikolaev: Please remove comma before "and"

gxgpet: Fixed. Thanks!

narfbg:
@gxgpet I haven't decided whether we want this or not yet, but either way I wouldn't merge this until it is stripped down to solving only a single problem. The method chaining thing has to go, sorry - if we don't comply with our own rules, they'd be meaningless.

@jim-parry @mwhitneysdsu @lonnieezell Thoughts on the `clear()` thing?

jim-parry:
I agree that clear() should reset the caption. That feels like an omission in the original.

gxgpet:
@narfbg I can take out the method chaining fix from this PR and make a separate one for it, lately this day. Is that ok with you?

narfbg:
Yep, that would work.

mwhitneysdsu:
My only reservation about clearing the caption in `clear()` is that it's a BC break.

I think @jim-parry is right in saying it feels like an omission, and I think people use captions on tables rarely enough that it probably won't cause a lot of issues (besides, the only reason it would cause an issue would be if someone was creating a series of tables with the same caption on purpose).

I also feel that the rarity of using both captions and this library is the only reason the omission has gone unnoticed this long. It is apparent that caption support was added after the fact, and someone would have needed to be very thorough to have noticed that the caption wasn't also added to the `clear()` method (and really, looking at the PR for `set_caption()`, I don't think it would have been accepted without some changes in the ~5 years I've been following this repo, so things were a little different in 2007).

So, I guess I'm saying that I support adding the caption to the `clear()` method. The documentation isn't specific enough to give someone the expectation that the caption wouldn't also be reset, so someone would only depend on that behavior based on observing the behavior of the code (or reading the code directly).

narfbg:
The BC break is my major concern as well. I guess we may merge, but keep it for version 3.2.0.

PR #4828, opened 2016-10-04
Zip Compress library improvements

gxgpet:
Right now, the Zip Compress library is saving all information of the archive directly into RAM memory (this means if we are creating a 10MB archive (with compression_level = 0), then an additional 10MB will be occupied by Zip library into RAM memory.

This can be easily fixed by using filesystem. In this PR, I'm suggesting to use [`tmpfile`](http://php.net/manual/en/function.tmpfile.php): a temporary file is creating into TEMP folder and there we shall save the `zipdata` property (which contains raw data of the archive). `directory` property will continue to live inside memory RAM.

I ran some tests regarding this change. As a test subject, I used a fresh copy of CodeIgniter folder for `read_dir` function. Compression level used was 0.

With the old ZIP library:

Memory usage before calling `read_dir`: 918,096 bytes
Memory after calling `read_dir`: 12,151,008 bytes
Execution time for `read_dir` call: 0.3140s
Total time for saving file (`archive` call): 0.0599s

With the changed ZIP library:

Memory usage before calling `read_dir`: 918,888 bytes
Memory after calling `read_dir`: 960,688 bytes
Execution time for `read_dir` call: 0.2253s
Total time for saving file (`archive` call): 0.0127s

It's obviously that 1MB RAM-usage request is better than a 12MB RAM-usage request. But as you can see, it's also faster!

**So how is the performance better by using the filesystem?**
1. Every time a write operation is performed to temp file, the result from `fwrite` function is directly added to `offset` property (as it represents the written number of bytes), removing the old `strlen` execution on `zipdata` property, thus using simple, pure math addition.
2. Every time `get_zip` (and as a result, `archive` too) was called (_and_ also every time a file was added to zip), an additional `strlen` was performed on `zipdata` property (_offset to start of central dir_), however, `offset` property already contains the necessary `zipdata` length :) 

I'm open to any suggestions or changes related to this PR, as I think we can make Zip Compress library really efficient on large archives too.

### DIFF
--- a/system/libraries/Zip.php
+++ b/system/libraries/Zip.php
@@ -55,11 +55,18 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 class CI_Zip {
 
 	/**
-	 * Zip data in string form
+	 * Zip data saved as a temp file
+	 *
+	 * @var resource
+	 */
+	public $zipdata = NULL;
+
+	/**
+	 * Zip data tmpfile path
 	 *
 	 * @var string
 	 */
-	public $zipdata = '';
+	public $zipdata_path = '';
 
 	/**
 	 * Zip data for a directory in string form
@@ -113,7 +120,33 @@ class CI_Zip {
 	public function __construct()
 	{
 		$this->now = time();
+
+		if ( ! ($this->zipdata = @tmpfile()))
+		{
+			log_message('error', 'Couldn\'t create Zip temp file.');
+			return;
+		}
+
+		// The 'uri' metadata holds the full-path to temp file
+		$zipdata_metas = stream_get_meta_data($this->zipdata);
+		$this->zipdata_path = $zipdata_metas['uri'];
+
 		log_message('info', 'Zip Compression Class Initialized');
+	}
+
+	// --------------------------------------------------------------------
+
+	/**
+	 * Close opened tmpfile (if any)
+	 *
+	 * @return	void
+	 */
+	public function __destruct()
+	{
+		if (is_resource($this->zipdata))
+		{
+			fclose($this->zipdata);
+		}
 	}
 
 	// --------------------------------------------------------------------
@@ -175,7 +208,7 @@ class CI_Zip {
 	{
 		$dir = str_replace('\\', '/', $dir);
 
-		$this->zipdata .=
+		$this->offset += fwrite($this->zipdata,
 			"\x50\x4b\x03\x04\x0a\x00\x00\x00\x00\x00"
 			.pack('v', $file_mtime)
 			.pack('v', $file_mdate)
@@ -188,7 +221,8 @@ class CI_Zip {
 			// below is "data descriptor" segment
 			.pack('V', 0) // crc32
 			.pack('V', 0) // compressed filesize
-			.pack('V', 0); // uncompressed filesize
+			.pack('V', 0) // uncompressed filesize
+		);
 
 		$this->directory .=
 			"\x50\x4b\x01\x02\x00\x00\x0a\x00\x00\x00\x00\x00"
@@ -206,7 +240,6 @@ class CI_Zip {
 			.pack('V', $this->offset) // relative offset of local header
 			.$dir;
 
-		$this->offset = strlen($this->zipdata);
 		$this->entries++;
 	}
 
@@ -260,7 +293,7 @@ class CI_Zip {
 		$gzdata = substr(gzcompress($data, $this->compression_level), 2, -4);
 		$compressed_size = strlen($gzdata);
 
-		$this->zipdata .=
+		$this->offset += fwrite($this->zipdata,
 			"\x50\x4b\x03\x04\x14\x00\x00\x00\x08\x00"
 			.pack('v', $file_mtime)
 			.pack('v', $file_mdate)
@@ -270,7 +303,8 @@ class CI_Zip {
 			.pack('v', strlen($filepath)) // length of filename
 			.pack('v', 0) // extra field length
 			.$filepath
-			.$gzdata; // "file data" segment
+			.$gzdata // "file data" segment
+		);
 
 		$this->directory .=
 			"\x50\x4b\x01\x02\x00\x00\x14\x00\x00\x00\x08\x00"
@@ -288,7 +322,6 @@ class CI_Zip {
 			.pack('V', $this->offset) // relative offset of local header
 			.$filepath;
 
-		$this->offset = strlen($this->zipdata);
 		$this->entries++;
 		$this->file_num++;
 	}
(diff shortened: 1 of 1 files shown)


system/libraries/Zip.php
```diff
@@ -113,12 +120,38 @@ class CI_Zip {
 	public function __construct()
 	{
 		$this->now = time();
+
+		if ( ! ($this->zipdata = @tmpfile()))
```
ValentinNikolaev: We shouldn't use dogbody. Find another solution here.

gxgpet: Ok then. I will also change this in `clear_data` method and the comparison will be made against `FALSE`.

gxgpet: I'm sorry, but I don't understand what are you pointing at. What is wrong, in this case, with `$this->zipdata === FALSE`? It's a strict comparison ...

ValentinNikolaev: I'm pointing on @tmpfile()

narfbg:
Before my actual comment, some context for the public - the OP asked me in private if we'd want such a change, as a form of performance optimization, claiming that writing straight to the file-system is actually faster than using memory, which made no sense to me.

-----

@gxgpet `strlen()` calls are not a memory vs file-system problem ... By eliminating them, you're just compensating for the negative impact of using the inherently slower file-system option.
I'm sure you know that perfectly well, and so I don't appreciate the kind of shady way you've presented it ...

gxgpet:
@narfbg We both know that somewhere to a lower level the `strlen` is parsing the memory until it reaches `\0`, is that right? C-way. Using simple addition to an integer it's better than parsing over and over the memory ... 

Yes, perhaps in the beginning it made no sense why would we rely on something that we know that is conventionally much slower than memory. But as the things progressed, we also found that having 10 users (a simple example) downloading a 10MB archive each would mean 100MB wasted by Zip library, instead of maximum 10MB ... ?

> as a form of performance optimization, claiming that writing straight to the file-system is actually faster than using memory, which made no sense to me.

I understand, but we don't just write something to memory/file-system and we go away, leaving it there and then compare the benchmarks. We perform additional operations on that data as well.

Also, I don't understand why you think I presented it in a shady way... there are just some issues I found and I thought that fixing them should be welcomed ... I just explained short and to the point the differences between the old implementation and the new one, I do not see where I did wrong ...

narfbg:
> We both know that somewhere to a lower level the strlen is parsing the memory until it reaches `\0`, right? C-way. Using simple addition to an integer it's better than parsing over and over the memory ... 

No, it doesn't work that way.

When the PHP engine creates a string (or any scalar) variable, it stores both its value *and length* in memory. `strlen()` uses that data and immediately knows the length number to return, it doesn't count it on the fly.
The overhead comes from the function call itself - this is a well-known fact - not from what the function does.

Either way, it does boil down to avoiding those `strlen()` calls and I don't disagree with the "simple addition" part of what you're saying, but that's not related to memory vs. file-system usage.

> Yes, perhaps in the beginning it made no sense why would we rely on something that we know is conventionally much slower than memory. But as the things progressed, we also found that having 10 users (a simple example) downloading a 10MB archive each would mean 100MB wasted by Zip library, instead of maximum 10MB ... ?

That's not wasted memory, it's *utilized* memory. This is not an argument.

>> as a form of performance optimization, claiming that writing straight to the file-system is actually faster than using memory, which made no sense to me.
>
> I understand, but we don't just write something to memory/file-system and we go away, leaving it there and then compare the benchmarks. We perform additional operations on that data as well.

I don't understand what you're trying to say here.

> Also, I don't understand why you think I presented it in a shady way... there are just some issues I found and I thought that fixing them should be welcomed ... I just explained short and to the point the differences between the old implementation and the new one, I don't see where I did wrong ...

The shady part is you saying that the whole thing is faster when avoiding memory usage, while *the entire performance gain* you have comes from eliminating `strlen()` calls.
And within our private conversation, you never mentioned that.

Yes, the code is currently written in a way that `strlen()` is necessary, but again - that's not a memory vs file-system problem, it's a logical one. It has literally nothing to do with the medium holding the data.
If you can refactor it *just* to eliminate the `strlen()` calls, keeping the memory usage, I promise that you'll see even greater performance gains.

gxgpet:
I can see where this conversation goes :) Unfortunately, for me, it stops here.

I don't have the time, nor the necessary energy (which I could use to make other contributions as well to this project) to write here, in the PR's comments section, a full book of arguments, since you are saying that 

> _utilized_ memory

is not an argument when talking about differences of MBs, invalidating it just because I named it _wasted_ and not _utilized_ :) But have it your way.

I'm not making anymore refactorizations regarding strlen, or regarding whatever for this lib. I tried to help, to contribute to this lib. Maybe because it's code written by you, that's why you may be so conservative about it, I don't know, it actually doesn't matter who is the author or what's the history of this file (or at least it should be this way).

The point is that this PR solves a big memory usage problem, and also helps with execution time, having example benchmark in the description. Perhaps you may believe the I am criticizing the library/or the code in anyway. I'm not doing it, nor I did. I just did some **Zip Compress library improvements**. That's all. :)

narfbg:
For the record, no - this is not my code and I'm not being defensive in any way.

The potential performance (speed) gains are the only takeaway that I'm actually interested in, and you're claiming that they come from using the file system. That is simply not true and hence why it annoys me that you've framed it that way.
I don't get why you're having such a hard time understanding this.

That "big memory usage problem" that you're talking about just doesn't exist for me - I'm not invalidating it because you said the wrong word. Quite the opposite, you've said the "right" word for what you meant to say.
But you seem to think that using memory should be avoided at all cost, and if that was the case, RAM wouldn't exist in the first place. Everything is a trade-off and when we want speed, we give up some memory to get it - that's what it is there for and that's why I say we're utilizing and not wasting it. To give an extreme example, a waste would be to fill it with random data that you never use.

aanbar:
I haven't used this library before; but I can see 1 advantage of writing to Disk over using RAM.

If using this library was ever hitting the memory_limit, this one should help resolve this issue.

vlakoff:
The key here is scalability. You should be able to compress large data, basically with RAM O(1) and CPU O(n).

Then, if optimizations can be made for small data (which entirely fit into RAM) they could be considered too, but it's a bonus.

Calico90:
First of all, @gxgpet thanks for this contribution and a well explanation including the benchmark.

> It's obviously that 1MB RAM-usage request is better than a 12MB RAM-usage request. But as you can see, it's also faster!

If the newer version performs as well for half the cost, the decision is a no-brainer. - https://en.wiktionary.org/wiki/no-brainer And it's actually 1/12th of the cost. 

I think @gxgpet has quite a good suggestion here that should probably be seriously reconsidered. _I assume that's also exactly why this PR is still open and not closed already._

narfbg:
This has sat here long enough and I want to close it. As there are benefits to either approach, the only potential solution I see is to provide a config option for which strategy to use - a future PR should include that if there's still interest in this.

PR #4832, opened 2016-10-08
Adding docker-compose / docker support

rawkode:


narfbg:
Sorry, but we don't accept stuff like that in general, unless the tools are a part of our own workflow.

PR #4833, opened 2016-10-08
Fix get an array data from memcached

vuongggggg:
Check if index 0 is not undefined or array has string as keys.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_memcached.php
+++ b/system/libraries/Cache/drivers/Cache_memcached.php
@@ -153,7 +153,7 @@ class CI_Cache_memcached extends CI_Driver {
 	{
 		$data = $this->_memcached->get($id);
 
-		return is_array($data) ? $data[0] : $data;
+		return is_array($data) && isset($data[0]) ? $data[0] : $data;
 	}
 
 	// ------------------------------------------------------------------------


narfbg:
While I'm not saying that there's something wrong with this, I also don't see the problem that it solves. Unless you've created that problem for yourself of course.

Could you elaborate?

vuongggggg:
Sorry sir!
It seems I did not understand its function. I will close this request.

tianhe1986:
@vuongggggg 
You should not close this pull request since some problems do exist. You could fix it.

I tried the following code.
```
		$this->load->driver('cache');
		$cache_info = [123,456,789,908];
		$this->cache->memcached->save('test_cache', $cache_info, 100);
		$out = $this->cache->memcached->get('test_cache');
		var_dump($out);
		$this->cache->memcached->save('test_cache', $cache_info, 100, TRUE);
		$raw_out = $this->cache->memcached->get('test_cache');
		var_dump($raw_out);  
```

and got

```
array (size=4)
  0 => int 123
  1 => int 456
  2 => int 789
  3 => int 908
int 123
```

However, they were expected to be the same.

My suggestion is using the same way as in `Cache_apc.php`, but it will cause BC break.  You may need to check the result of `unserialize()` to prevent it.

@narfbg 
Any better advice?

narfbg:
@tianhe1986 I don't know why you'd expect those to be the same, nor how that has anything to do with the patch here.

tianhe1986:
@narfbg 
Because the data I get from cache should be the same as what I put into cache if it is still effictive, whether or not stored with the raw value.

I mean while getting an array data from memcached stored with raw value, I will get the wrong result if this `$data` contains  `$data[0]`.  

Sorry, I should suggest @vuongggggg creating a new pull request.

narfbg:
For anything to be stored as a "raw value", Memcache must have the same data type and it doesn't have an array type ...

You're not getting a wrong result, you just want it to do something that it can't possibly do.

tianhe1986:
@narfbg 
I'm not talking about Memcache. I also know the data would be represented in a serialized form in Memcache.

I am just saying about `drivers/Cache_memcached.php`.

I mean, for exmple,
```
$this->cache->memcached->save($key, $cache_info, 100, TRUE);
$data = $this->cache->memcached->get($key);
```

`$cache_info` and `$data` should be the same, right?

narfbg:
No.

They *may* be the same for data types that both PHP and Memcache have, and talking about Cache_memcached **is** talking about Memcache.

The ``$raw`` parameter is not supposed to serialize, that's the whole point.

tianhe1986:
@narfbg 
OK, I get the key point. 

How about throw exception or return FALSE when encountering array type `$data` containing `$data[0]`  with `$raw` parameter set to TRUE ? 

Maybe better than storing it but returning unexpected value ?

vuongggggg:
@tianhe1986 I know this problem and I fixed it in forked source. Because I use pure PHP & CodeIgniter for my project, it really is a problem when I use string key to save the value.

I've closed this issue because i can't explain it in an understandable way.

PR #4847, opened 2016-10-16
Composer install or update commands generates autoload.php file in ve…

iarjunphp:
…ndor folder but CI application is looking at app directory. So In order to fix this issue changed the path to one level up


narfbg:
Duplicate of #3636.

Issue #4854, opened 2016-10-18
DBFORGE "text" column creation/modification is buggy


AmitMY:
If I try to create a new table with the following column, or modify the following column, it does not go well.
> {
>    "type": "text",
>    "constraint": "65535"
> }

Creates

> {
>    "type": "mediumtext",
>    "constraint": "16777215"
> }

I am not sure why it does that.

narfbg:
You can't control the length of a "text" field ...

I have to ask you again, please go through our [forums](http://forum.codeigniter.com/) first and only come here when you are **absolutely 100% sure** that something is a bug.

PR #4861, opened 2016-10-19
DB_Forge "_process_fields" can now handle "TEXT" fields

AmitMY:
Before, If trying to add a new field, with type="text", and "length"=65535 which is the length taken from the database "information" tables, it would add a "mediumtext" field with size "16777215" on MySQL

Addressing issue [#4854](https://github.com/bcit-ci/CodeIgniter/issues/4854)

### DIFF
--- a/system/database/DB_forge.php
+++ b/system/database/DB_forge.php
@@ -777,6 +777,8 @@ abstract class CI_DB_forge {
 			{
 				switch (strtoupper($attributes['TYPE']))
 				{
+					case 'TEXT': // TEXT columns can not have a length
+						break;
 					case 'ENUM':
 					case 'SET':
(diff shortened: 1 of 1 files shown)


narfbg:
I don't want such constraints ignored.
They *have* to blow up in your face, so that you know you shouldn't try to do them.

AmitMY:
@narfbg So what do you think about maybe throwing an exception in that case? Because leaving it the way it is is confusing, and does not blow up in your face...

narfbg:
It may not blow up in your face in MySQL, because MySQL by default allows you to do all kinds of stuff that you shouldn't be allowed to.
That isn't the case if you turn off that feature, or if you use a more strict by default database platform.

Either way ... I really don't understand the problem here.

PR #4862, opened 2016-10-19
New function for Inflector helper: ordinal_format() 

gxgpet:
Added ordinal_format() which return the ordinal numeral for a given number.


system/helpers/number_helper.php
```diff
@@ -92,3 +92,32 @@ function byte_format($num, $precision = 1)
...
+{
+	/**
+	 * Returns the English ordinal numeral for a given number
+	 *
+	 * @param  int    $number
+	 * @return string
+	 */
+	function ordinal_format($number)
+	{
+		if ( ! is_int($number) OR $number < 1)
```
narfbg: Why disallow numeric strings? *Especially* when you return a string.
Also, negative numbers?

system/helpers/number_helper.php
```diff
@@ -92,3 +92,32 @@ function byte_format($num, $precision = 1)
...
+		{
+			return FALSE;
+		}
+
+		$ends = array('th', 'st', 'nd', 'rd', 'th', 'th', 'th', 'th', 'th', 'th');
+		if ((($number % 100) >= 11) && (($number % 100) <= 13))
```
narfbg: While this is indeed a simple expression, its logic is not that easy to understand.
Why not just lookup `$number[strlen($number)-1]` ?

gxgpet: Because it won't work for 11 (eleventh), 12 (twelfth) and 13 (thirteenth). :)

system/helpers/number_helper.php
gxgpet: Fixed by using is_numeric.
Ordinal terms are [never used in the negative](http://english.stackexchange.com/questions/309713/ordinal-form-of-negative-numbers-especially-1-2-3).

system/helpers/number_helper.php
narfbg: I see ... Ok then.

But the extra parentheses around both comparisons need to be removed - there's a line between adding some for clarity and putting them around everything. :)

Also, `$ends` is kind of fine, but `$last_digit` is better, and explicitly setting its keys would make it a LOT easier to read.

system/helpers/number_helper.php
```diff
@@ -92,3 +92,32 @@ function byte_format($num, $precision = 1)
...
+{
+	/**
+	 * Returns the English ordinal numeral for a given number
+	 *
+	 * @param  int    $number
+	 * @return string
+	 */
+	function ordinal_format($number)
+	{
+		if ( ! is_numeric($number) OR $number < 0)
```
narfbg: `is_numeric()` also allows floats (and something else weird that I don't remember)

You need `ctype_digit((string) $number)`

system/helpers/number_helper.php
gxgpet: I have a better suggestion: how about `$suffixes`? :) Since they are really strings, I think it would look nicer with it.

system/helpers/number_helper.php
narfbg: Ok. Your new version allows 0 though; why that?

Plus, this conversation gets me thinking of why this isn't an inflector helper?

system/helpers/number_helper.php
narfbg: It wouldn't be entirely correct ... the suffix is what the function is adding.

system/helpers/number_helper.php
gxgpet: Fixed. Thanks!

system/helpers/number_helper.php
narfbg: You are playing by linguistic rules here ... math has no rules for "th" or "nd".

Well, there are also some sources that would say "the -5th floor" (despite it being apparently incorrect) ... I've seen that more often than "zeroth". In fact, I've never seen "0th".

system/helpers/number_helper.php
```diff
@@ -92,3 +92,42 @@ function byte_format($num, $precision = 1)
...
+			0 => 'th',
+			1 => 'st',
+			2 => 'nd',
+			3 => 'rd',
+			4 => 'th',
+			5 => 'th',
+			6 => 'th',
+			7 => 'th',
+			8 => 'th',
+			9 => 'th');
```
narfbg: Closing parenthesis needs to be on the next line.

system/helpers/number_helper.php
```diff
@@ -92,3 +92,42 @@ function byte_format($num, $precision = 1)
...
+		if ((($number % 100) >= 11) && (($number % 100) <= 13))
```
narfbg: You're still keeping extra parentheses.

gxgpet: Fixed this and coding style problem. Sorry.

system/helpers/number_helper.php
gxgpet: Yes, that's true as well ... I don't know what to say ...

So we should keep only non-negative natural numbers?

gxgpet: Cool.

I will do it during today evening.

system/helpers/inflector_helper.php
```diff
@@ -273,3 +273,43 @@ function is_countable($word)
...
+			2 => 'nd',
+			3 => 'rd',
+			4 => 'th',
+			5 => 'th',
+			6 => 'th',
+			7 => 'th',
+			8 => 'th',
+			9 => 'th'
+		);
+		if (($number % 100) >= 11 && ($number % 100) <= 13)
```
narfbg: There needs to be an empty line before this and the array's closing parenthesis.

system/helpers/inflector_helper.php
```diff
@@ -273,3 +273,43 @@ function is_countable($word)
...
+		{
+			return $number.'th';
+		}
+		else
+		{
+			return $number.$last_digit[$number % 10];
```
narfbg: This doesn't need to be inside an `else` block.

narfbg:
It appears that your empty lines contain tabs ... please make sure they're completely empty in the future.

PR #4864, opened 2016-10-21
Redis Lock sleep time to 0.1s from 1s.

yoshida-mediba:
Performance of the CPU could not use in sleep.

$ h2load -n2000 -c80 --header="Accept-Encoding: gzip" https://localhost/

```
1s: finished in 45.89s, 43.58 req/s, 169.77KB/s
0.1s: finished in 16.47s, 303.51 req/s, 1.15MB/s
```


narfbg:
This changes the logic of re-trying for 30 seconds before giving up and reduces it to 3.

yoshida-mediba:
There is also certainly so say problem,
web in 30 seconds, I think that is too long.

narfbg:
... that's the absolute maximum timeout, not something that happens on every request.

narfbg:
It's been a month without a reply, closing.

PR #4878, opened 2016-10-26
Fixed array parsing

Arthom:
Example data:
	$data = array('testing' => 'testingvalue', 'testingarray' => array('res1' => 'res1', 'res2' => 'res2'));

Example page with text:
	{testing}<br /><br />

	{testingarray}
		{res1}<br />
		{res2}
	{/testingarray}

And result:
	testingvalue

	res1
	res2

You can try that :)

### DIFF
--- a/system/libraries/Parser.php
+++ b/system/libraries/Parser.php
@@ -214,31 +214,25 @@ class CI_Parser {
 			PREG_SET_ORDER
 		);
 
-		foreach ($matches as $match)
-		{
+		foreach ($matches as $match) {
 			$str = '';
-			foreach ($data as $row)
+			$temp = array();
+			foreach ($data as $key => $val)
 			{
-				$temp = array();
-				foreach ($row as $key => $val)
+				if (is_array($data[$key]))
 				{
-					if (is_array($val))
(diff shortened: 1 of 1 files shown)


narfbg:
I don't understand the problem you're trying to fix, please elaborate.

Arthom:
I had a problem with array parsing (ex. testingarray).
$data = array('testing' => 'testingvalue', 'testingarray' => array('res1' => 'res1', 'res2' => 'res2'));

It threw me error.
screenshot from 2016-10-26 17-49-01

Line 258:
foreach ($row as $key => $val)

narfbg:
You're just not doing it right ... the correct usage is shown [here](https://codeigniter.com/userguide3/libraries/parser.html#variable-pairs).

Arthom:
But it's good idea to parse this method of array :)

narfbg:
You think it's a good idea, I don't necessarily do.
But I think we can both agree on what's bad - breaking existing functionality, and your patch does that.

Arthom:
But when you want to parse variable with array, you can't do that. You need to add extra array to foreach that object.

PR #4883, opened 2016-10-27
CI_Email option do disable SSL peer verification

punga78:
SSL3_GET_SERVER_CERTIFICATE:certificate verify failed

### DIFF
--- a/system/libraries/Email.php
+++ b/system/libraries/Email.php
@@ -223,6 +223,14 @@ class CI_Email {
 	 */
 	public $bcc_batch_size	= 200;
 
+	/**
+	 *  Verify peer in ssl connection 
+	 *
+	 * 
+	 * @var	bool
+	 */
+	public $ssl_peer_verification	= FALSE;
+	
 	// --------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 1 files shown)


narfbg:
If you want to make verification optional (but opt-out), we may talk about it.
But I won't accept just silently ignoring verification.

punga78:
Ok, why can, maybe set, config parameter then Ingoring Verification, what do you think ?

narfbg:
If it's configurable, it'd be acceptable, yes.
You'll have to get familiar with our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html) though.

narfbg:
Still not complying with the [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).
And it's still Off by default, which is the opposite of what we'd want.

Also, please leave a comment after you make changes, as commits don't issue notifications ... that's why I'm replying now - I had some spare time and decided to go through open issues and PRs; could've been months before I saw it. :)

system/libraries/Email.php
```diff
@@ -223,6 +223,14 @@ class CI_Email {
 	 */
 	public $bcc_batch_size	= 200;
 
+	/**
+	 *  Verify peer in ssl connection 
+	 *
+	 * 
+	 * @var	bool
+	 */
+	public $ssl_peer_verification	= FALSE;
```
narfbg: Rename this to `$ssl_verify_peer` (and set it to TRUE) please.

system/libraries/Email.php
```diff
@@ -2007,11 +2015,21 @@ protected function _smtp_connect()
 
 		$ssl = ($this->smtp_crypto === 'ssl') ? 'ssl://' : '';
 
-		$this->_smtp_connect = fsockopen($ssl.$this->smtp_host,
-							$this->smtp_port,
-							$errno,
-							$errstr,
-							$this->smtp_timeout);
+		if (($this->smtp_crypto === 'ssl') and (!$ssl_peer_verification))
```
narfbg: ... this can't work.

It's a class property, not a local variable.

narfbg:
On another note, titles like "fix" are really bad ... I've updated this one, but be more descriptive yourself in the future.

PR #4886, opened 2016-10-28
Match both single and double quote in DB_driver

tianhe1986:
I use the following code in controller.

```
$this->load->database();
$sql = 'select * from reader where reader_id = ? and level = ? and tag = "1?2"';
$binds = ['3456', '888'];
$compiled_sql = $this->db->compile_binds($sql, $binds);
echo $compiled_sql;
```

The expected output should be
`select * from reader where reader_id = '3456' and level = '888' and tag = "1?2"`

However,  get
`select * from reader where reader_id = ? and level = ? and tag = "1?2"`

It is because that double quote does not be matched in `compile_binds()`.

And the same with `is_write_type()`.

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -982,7 +982,7 @@ abstract class CI_DB_driver {
 	 */
 	public function is_write_type($sql)
 	{
-		return (bool) preg_match('/^\s*"?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $sql);
+		return (bool) preg_match('/^\s*["\']?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $sql);
 	}
 
 	// --------------------------------------------------------------------
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -938,7 +938,7 @@ abstract class CI_DB_driver {
 		$ml = strlen($this->bind_marker);
 
 		// Make sure not to replace a chunk inside a string that happens to match the bind marker
-		if ($c = preg_match_all("/'[^']*'/i", $sql, $matches))
+		if ($c = preg_match_all("/(['\"])[^\\1]*\\1/i", $sql, $matches))
 		{
 			$c = preg_match_all('/'.preg_quote($this->bind_marker, '/').'/i',
 				str_replace($matches[0],
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -982,7 +982,7 @@ abstract class CI_DB_driver {
 	 */
 	public function is_write_type($sql)
 	{
-		return (bool) preg_match('/^\s*["\']?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $sql);
+		return (bool) preg_match('/^\s*"?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $sql);
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


narfbg:
The double quote doesn't necessarily mean a literal string - this doesn't appear to matter for your patch, but you should have it in mind. EVERY major database platform except MySQL uses it as the escape character for identifiers.

system/database/DB_driver.php
```diff
@@ -982,7 +982,7 @@ public function compile_binds($sql, $binds)
 	 */
 	public function is_write_type($sql)
 	{
-		return (bool) preg_match('/^\s*"?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $sql);
+		return (bool) preg_match('/^\s*["\']?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $sql);
```
narfbg: Revert this. It has nothing to do with `compile_binds()` and you don't know what it does, I don't even know why it is here.

system/database/DB_driver.php
```diff
@@ -938,7 +938,7 @@ public function compile_binds($sql, $binds)
 		$ml = strlen($this->bind_marker);
 
 		// Make sure not to replace a chunk inside a string that happens to match the bind marker
-		if ($c = preg_match_all("/'[^']*'/i", $sql, $matches))
+		if ($c = preg_match_all("/(['\"])[^\\1]*\\1/i", $sql, $matches))
```
narfbg: You're not just adding the double quote to the negated character class, you're changing the entire pattern. Why?

system/database/DB_driver.php
tianhe1986: OK I'll revert it

system/database/DB_driver.php
tianhe1986: The original pattern means matching `'XXXX'`, and the new pattern means matching `'XXXX'` or `"XXXX"`

narfbg: Not really ... that's how you translate it for yourself, but not what it actually means.

The original pattern means "catch anything as many non-single quote characters as you can"
Your modified pattern includes the quotes in itself

tianhe1986: Yes, the original pattern  means "catch anything as many non-single quote characters as you can", and these characters are surround by a pair of single quote.

My modified pattern keeps this meaning, and adds "catch anything as many non-double quote characters as you can".

It is equivalent to `/'[^']*'|"[^"]*"/`

narfbg: Yes, I know what it is equivalent to.

What I meant is something else, but it turns out I didn't notice the quotes outside the range block in the original. My bad, sorry.

tianhe1986:
@it-can Discuss here. Could you provide your query and binding array ?

it-can:
This query fails...
```
SELECT
    a.id,
    CONCAT(a.id," - ",a.username) AS value
FROM users a
WHERE a.deleted=? AND a.username LIKE "%ja%"';

$query = $this->db->query($sql, array('no'));
$data = $query->result();
```

This query works...
```
SELECT
    a.id
FROM users a
WHERE a.deleted=? AND a.username LIKE "%ja%"';

$query = $this->db->query($sql, array('no'));
$data = $query->result();
```

tianhe1986:
@it-can @narfbg 
It does going wrong.But if I change the pattern to `"/'[^']*'|\"[^\"]*\"/i"`, it works well. So I'm a little confused.

narfbg:
Well, [this guy](http://stackoverflow.com/a/8057827/468027) has an explanation, the key being this:

> As far as the regex engine knows, `\1` could represent any number of characters

In other words, since the engine can't know if `\1` is a single or multiple characters, you can't put that inside a character class.

tianhe1986:
@narfbg OK, I see. So I'll change it to the  non-character-class pattern. How should I do? A new pull request or just commint in the same branch?

tianhe1986:
@narfbg One more thing. There are many the same use cases as I did in this patch in [Security.php](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Security.php#L162). However, it uses the non-greedy mode which works well.

narfbg:
This PR is already merged ... you can commit to the same branch, but would still have to submit those commits in a separate PR.

We'll have to test the `CI_Security` stuff, indeed, but that's a separate issue.

lxocram:
this extended fix sadly breaks this code

`$this->db->query('EXEC getSomething @params=\'' . rtrim(str_repeat('?,', count($selected_items)), ',') . '\'', $selected_items);
`
because i have a chunk inside a string that happens to match the bind marker
in

`EXEC getSomething @params='?,?,?,?' `

the bind marker is inside a string

I now have to bind outside of the query statement first and then use that bound value
`
$bound_params=$this->db->compile_binds(rtrim(str_repeat('?,', count($selected_items)), ',') , $selected_items);
$this->db->query('EXEC getSomething @params=\'' . $bound_params . '\'');
`

Issue #4887, opened 2016-10-28
File upload issue in 3.1.1 & 3.1.2


karthik-code:
After upgrading to 3.1.1, file upload not working. 
HTTP ERROR 500 occurs. 
PHP version : 5.6.15

The same code works with CI version 3.1.0 or lower.
I have tried with latest CI release 3.1.2, still it is not working.

narfbg:
The HTTP 500 status code means "server-side error", practically translating to "something doesn't work", which you're saying anyway.

But there's a ton of stuff going on in there ... please provide more info.

karthik-code:
I am using this simple code to upload files
```php
public function do_upload()
        {
                $config['upload_path']          = './uploads/';
                $config['allowed_types']        = 'csv';               

                $this->load->library('upload', $config);

                if ( ! $this->upload->do_upload('csvfile'))
                {
                        echo $this->upload->display_errors();

                }
                else
                {
                        $data = array('upload_data' => $this->upload->data());

                        echo 'Upload Success';
                }
        }
```
When trying to upload the file HTTP Error 500 appears.
There is no other specific error shown or logged.
Log ends at INFO - 2016-10-28 22:03:06 --> Upload Class Initialized

The same function works perfectly with CI system 3.1.0 or lower version.
My php version 5.6

The same issue reported by other users. 
http://forum.codeigniter.com/thread-66504-post-337924.html

narfbg:
Unfortunately, that's not saying anything more ...

karthik-code:
Yes, what other information do you expect?

narfbg:
Something that actually points to the supposedly erroneous part of the code ...

I don't know what your logging error level is, or if/where your PHP errors are logged - that may help.
Try commenting out some pieces of code in `CI_Upload` to see when the error disappears ... stuff like that, you know how to debug.

hstreams:
![codeigniter error - copy](https://cloud.githubusercontent.com/assets/13208801/19816802/7259aed6-9d51-11e6-92fe-2a94190a1d0c.png)
this is the issue, he wrote about, but i think he made the env. to production not development.

andersonunsonst:
, you need enable in php.ini ;extension=fileinfo.so or ;extension=php_fileinfo.dll

karthik-code:
Thanks. 
Enabling the php_fileinfo extension works.
I don't know why errors not displayed in development environment. Also, I set ini_set('display_errors', 1); at top of index.php file.
[...]

PR #4888, opened 2016-10-29
Check filefino extention is loaded while uploading

mszb:
Added function to check either fileinfo extention is loaded or not.
Before this check, if extention was not loaded, script jsut stop and
there is no log of what happend. Now this will generate an error log and
return.

### DIFF
--- a/system/libraries/Upload.php
+++ b/system/libraries/Upload.php
@@ -1218,6 +1218,13 @@ class CI_Upload {
 		// We'll need this to validate the MIME info string (e.g. text/plain; charset=us-ascii)
 		$regexp = '/^([a-z\-]+\/[a-z0-9\-\.\+]+)(;\s.+)?$/';
 
+		//Check fileinfo extention is loaded or not
+		if(!extension_loaded('fileinfo')) 
+		{
+			$this->set_error('upload_fileinfo_notfound', 'error');
+			return;
+		}
+
 		// Fileinfo extension - most reliable method
 		$finfo = @finfo_open(FILEINFO_MIME);
 		if (is_resource($finfo)) // It is possible that a FALSE value is returned, if there is no magic MIME database file found on the system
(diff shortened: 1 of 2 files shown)


system/libraries/Upload.php
```diff
@@ -1218,6 +1218,13 @@ protected function _file_mime_type($file)
 		// We'll need this to validate the MIME info string (e.g. text/plain; charset=us-ascii)
 		$regexp = '/^([a-z\-]+\/[a-z0-9\-\.\+]+)(;\s.+)?$/';
 
+		//Check fileinfo extention is loaded or not
+		if(!extension_loaded('fileinfo'))
```
gxgpet: A space [should](http://www.codeigniter.com/user_guide/general/styleguide.html#logical-operators) always precede and follow `!`.

narfbg:
The extension in question is [enabled by default since PHP 5.3](https://secure.php.net/manual/en/fileinfo.installation.php), which is why a `function_exists()` check was removed in a838279625becfba98ccb7635d35c67297129c42.

If we're to do anything about that, it would be to revert that change.

Referencing #4887.

andersonunsonst:
but, windows users yet need enable this, it would be interesting a alert with throw exception

mszb:
@narfbg .. yes i agree with @andersonunsonst . I've counter this issue in windows systems. That's why i created this pull-request. Because, if file info extension is not enable and there is no error log, user have no clue whats going on.

narfbg:
*sigh*

My first sentence explains *how and why* this issue appeared. And the reference to #4887 was meant to point you to an *open* issue where we're investigating it further.

And when I said this:

> If we're to do anything about that, it would be to revert that change.

I meant that if any change should be made, it would be to make it like [this](https://github.com/bcit-ci/CodeIgniter/blob/3.1.0/system/libraries/Upload.php#L1215) again.

PR #4891, opened 2016-10-31
Download helper added new use case 

gxgpet:
When using download helper, I can see 3 use cases:

1. You have the content of the file to be downloaded and you provide the name of file as it should be downloaded;
2. You have to read the content from disk and the downloaded file name will be the same as read file;
3. You have to read the content from disk and you provide the name of file as it should be downloaded.

At this point, the first 2 do exist and work. However, 3rd point can be achieved by dev by renaming the local file and then, most probably, renaming it back to original filename.
This PR adds the 3rd use case by letting you provide an array with one item in first param: [local file full path => browser filename].

### DIFF
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -70,7 +70,7 @@ if ( ! function_exists('force_download'))
 		elseif ($data === NULL)
 		{
 			// Is $filename an array as ['local source path' => 'destination filename']?
-			if(is_array($filename))
+			if(is_array($filename) && count($filename) === 1)
 			{
 				$filepath = key($filename);
 				$filename = current($filename);
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -70,12 +70,12 @@ if ( ! function_exists('force_download'))
 		elseif ($data === NULL)
 		{
 			// Is $filename an array as ['local source path' => 'destination filename']?
-			if(is_array($filename) && count($filename) == 1)
+			if (is_array($filename) && count($filename) == 1)
 			{
 				$filepath = key($filename);
 				$filename = current($filename);
 
-				if(is_int($filepath))
+				if (is_int($filepath))
 				{
 					return;
 				}
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -75,6 +75,11 @@ if ( ! function_exists('force_download'))
 				$filepath = key($filename);
 				$filename = current($filename);
 
+				if(is_int($filepath))
+				{
+					return;
+				}
+
 				if ( ! @is_file($filepath) OR ($filesize = @filesize($filepath)) === FALSE)
 				{
 					return;
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -72,7 +72,7 @@ if ( ! function_exists('force_download'))
 			// Is $filename an array as ['local source path' => 'destination filename']?
 			if (is_array($filename))
 			{
-				if (count($filename) != 1)
+				if (count($filename) !== 1)
 				{
 					return;
 				}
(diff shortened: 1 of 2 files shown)


system/helpers/download_helper.php
```diff
@@ -69,14 +69,28 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
 		}
 		elseif ($data === NULL)
 		{
-			if ( ! @is_file($filename) OR ($filesize = @filesize($filename)) === FALSE)
+			// Is $filename an array as ['local source path' => 'destination filename']?
+			if(is_array($filename) && count($filename) === 1)
```
narfbg: Not following the [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).

And what if it's an array with more than one elements?
Or if it's not an associative array?

user_guide_src/source/helpers/download_helper.rst
```diff
@@ -37,7 +37,9 @@ The following functions are available:
 	file data.
 
 	If you set the second parameter to NULL and ``$filename`` is an existing, readable
-	file path, then its content will be read instead.
+	file path, then its content will be read instead. You may also set ``$filename``
+	as an array containing a single entry, the key being the existing, readable file path
```
narfbg: This description isn't wrong, but sounds a bit broken ... needs different wording.

system/helpers/download_helper.php
gxgpet: Fixed the styleguide problem (matching using `==`).

If it's an array with more than one element, it will be used the element where array's pointer is pointing at. If it's not an associative array, it will `return;` since there will be no file named `0`, `1`, `2`.

Will fix soon the docs wording problem.

user_guide_src/source/helpers/download_helper.rst
gxgpet: I think it's better now. :)

system/helpers/download_helper.php
gxgpet: @narfbg Actually, I thought better that numeric arrays shouldn't be allowed at all (extensionless files such as `0`, `1`, `2` etc. may exist tho).

narfbg:
gxgpet/CodeIgniter@c953c128649e06bb1a3594e56459cf14bdb6e048 and/or gxgpet/CodeIgniter@b907c8ad41dfb4a9394b3a73182f0abf7fe14f94 are not what I meant with the [styleguide](https://codeigniter.com/userguide3/general/styleguide.html), these are not style changes (and we do prefer strict checks, although a simple `! empty()` check here would be more efficient).
What I meant is that you don't put a space after the `if` keyword, and with this being your 8th pull request you should really know that by now. Please do get yourself familiar with our styleguide.

And yes, I agree that [numeric key arrays should be disallowed](https://github.com/bcit-ci/CodeIgniter/pull/4891#discussion_r85880166) - that's what I meant, you were missing a check for that.

gxgpet:
@narfbg Yes, indeed, my bad there... tho my IDE should've warned me when doing "changes review" :/ Sorry. Next time I will pay more attention.

The reason for `count($filename) == 1` is that we strictly should allow only single-element arrays.

narfbg:
Yes, I see that's what you're trying to do, but now if the count doesn't match it falls into the string case, while every other incorrect input silently returns.

gxgpet:
Indeed, you are right. Now will enter the array case no  matter what kind of array is.

system/helpers/download_helper.php
```diff
@@ -70,8 +70,13 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
 		elseif ($data === NULL)
 		{
 			// Is $filename an array as ['local source path' => 'destination filename']?
-			if (is_array($filename) && count($filename) == 1)
+			if (is_array($filename))
 			{
+				if (count($filename) != 1)
```
narfbg: `!== 1`

PR #4893, opened 2016-10-31
Fix compile_binds: do not use back references inside a character class.

tianhe1986:
Fix pull request [4886](https://github.com/bcit-ci/CodeIgniter/pull/4886)


narfbg:
... instead of manually writing `[<issue/PR number>](<url>)`, please just prefix with `#` - a link will be created automatically, and by seeing the `#` sign in front, we'll know what you're linking to.

tianhe1986:
Fixed.

narfbg:
That was just for future reference. :)

PR #4896, opened 2016-11-01
URL Helper new function: tel()

gxgpet:
Added new helper function: `tel()`, which generates a `tel:` link.


narfbg:
Duplicate of #3784.

PR #4897, opened 2016-11-02
Small safe_mailto() improvements

gxgpet:
This PR is changing URL helper `safe_mailto()` with the following:

- `$attributes` are now parsed by `_stringify_attributes()` then added to array;
- removed unnecessary spacing (tabs, new lines, spaces);
- refactorised the JS code a bit (now `l` should be a local var, instead of global one).

### DIFF
--- a/system/helpers/url_helper.php
+++ b/system/helpers/url_helper.php
@@ -342,18 +342,18 @@ if ( ! function_exists('safe_mailto'))
 
 		$output = "<script type=\"text/javascript\">\n"
 			."//<![CDATA[\n"
-			."for(var l=[";
+			.'for(var l=[';
 
 		for ($i = 0, $c = count($x); $i < $c; $i++)
 		{
-			$output .= "'" .$x[$i] . "',";
+			$output .= "'".$x[$i]."',";
 		}
 
-		$output = rtrim($output, ',') . "]";
+		$output = rtrim($output, ',').']';
 
-		$output .= ",i = l.length-1; i >= 0; i=i-1)"
-			."l[i].substring(0, 1) === '|' ? document.write(\"&#\"+unescape(l[i].substring(1))+\";\"):"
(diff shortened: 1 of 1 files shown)


narfbg:
Why?

There's nothing wrong with the current code, there's only a chance that you break it with these changes (and you're probably doing that by using `_stringify_attributes()`).

gxgpet:
@narfbg Tested with Netscape Navigator 9.0.0.6. :) 

Because that spacing (for instance) won't help anyone (including browser). At it's space reduced. :)

narfbg:
Netscape Navigator? Seriously?

gxgpet:
So testing if the code is working on an ancient browser is not accurrate. OK ...

gxgpet:
Obviously that it works on any modern browser.

narfbg:
It's not about accuracy, it's about common sense.

Why would we want to support a browser that nobody uses? I shouldn't have to explain this.

gxgpet:
It's was about testing, NOT code written only for Netscape. Wtf?

narfbg:
We simply don't deal with such stuff, and you mentioned it when I asked you why you're requesting these changes ... what do you expect me to say?

gxgpet:
Honestly, I expected you to say "Yes, I looked over the committed code and it's definitely working only for Netscape" ...

narfbg:
I don't care about Netscape, nobody does.

gxgpet:
Neither do I, but you just raised 

>  there's only a chance that you break it with these changes

narfbg:
Yes, as in "you're changing code that works perfectly well and doesn't need any changes whatsoever".

How can that possibly bring you to freaking Netscape Navigator? Why are we even having this conversation?

gxgpet:
> Why are we even having this conversation?

I don't know :) 

Actually, I don't even know why I'm even trying. That's my big mystery.

PR #4899, opened 2016-11-03
Update to pdo_driver.php

cwhyman:
This  update adds the ability to specify a pdo dsn that is defined in php.ini
For example in the  php.ini file you can say pdo.dsn.<dsn name> = "<connection string>"
And in your application you just specify the dsn name instead of the connection string.  The patch also verifies that the dsn is defined in the photo.ini and if it isn't it defaults back to regular behavior.

### DIFF
--- a/system/database/drivers/pdo/pdo_driver.php
+++ b/system/database/drivers/pdo/pdo_driver.php
@@ -79,7 +79,11 @@ class CI_DB_pdo_driver extends CI_DB {
 	public function __construct($params)
 	{
 		parent::__construct($params);
-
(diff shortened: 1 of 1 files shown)


narfbg:
We generally aim for the opposite - override any INI settings ...

Anybody can do this:

```php
$db['default']['dsn'] = get_cfg_var('whatever');
```

cwhyman:
Oh I didn't think of that, when I was looking through your documentation and everything I could find online on how to connect using PDO DSN's, nowhere does it specify that snippet of code. Perhaps adding that example to your documentation could benefit others trying to do the same thing.

PR #4900, opened 2016-11-03
Provide ilike functionality

psuplat:
Hi folks,

as I often work with PostgreSQL and use its native function for case-insensitive LIKE function: ILIKE

This is just a very quick fix to allow for use of this function with Codeigniter.

NOTE: I admittedly do not have a whole deal of experience with testing, so it would be advisable if someone could give it a proper look over to it. From whatever testing I did it worked as expected and didn't caused any error.


narfbg:
The Query Builder is a platform abstraction ... this is a PostgreSQL-specific feature.

PR #4901, opened 2016-11-04
Update Form_validation.php

Batice77:
Adding ability to use a schema with a table. example: $this->form_validation->set_rules('identifiant', 'Identifiant', 'required|is_unique[cvven.client.identifiant]');


narfbg:
Thanks, but no thanks ... this is too much of an edge case and FV dealing with SQL shouldn't have happened anyway.

PR #4908, opened 2016-11-10
Update directory_helper.php

churavka:
Return only array directory names witout DIRECTORY_SEPARATOR.

### DIFF
--- a/system/helpers/directory_helper.php
+++ b/system/helpers/directory_helper.php
@@ -80,11 +80,9 @@ if ( ! function_exists('directory_map'))
 					continue;
 				}
 
-				is_dir($source_dir.$file) && $file .= DIRECTORY_SEPARATOR;
-
 				if (($directory_depth < 1 OR $new_depth > 0) && is_dir($source_dir.$file))
 				{
-					$filedata[$file] = directory_map($source_dir.$file, $new_depth, $hidden);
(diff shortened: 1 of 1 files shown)


narfbg:
Elaborate please.

vlakoff:
As already discussed in the past, we add a trailing separator after the directories because it's the only way to distinguish them from files.

system/helpers/directory_helper.php
```diff
@@ -80,11 +80,9 @@ function directory_map($source_dir, $directory_depth = 0, $hidden = FALSE)
 					continue;
 				}
 
-				is_dir($source_dir.$file) && $file .= DIRECTORY_SEPARATOR;
-
 				if (($directory_depth < 1 OR $new_depth > 0) && is_dir($source_dir.$file))
 				{
-					$filedata[$file] = directory_map($source_dir.$file, $new_depth, $hidden);
+					$filedata[$file] = directory_map($source_dir.$file.DIRECTORY_SEPARATOR, $new_depth, $hidden);
```
vlakoff: This change is actually useless

system/helpers/directory_helper.php
```diff
@@ -80,11 +80,9 @@ function directory_map($source_dir, $directory_depth = 0, $hidden = FALSE)
 					continue;
 				}
 
-				is_dir($source_dir.$file) && $file .= DIRECTORY_SEPARATOR;
```
vlakoff: As commented on the PR, this should be kept.

PR #4910, opened 2016-11-15
Support array parse in DB_query_builder::where/having

tianhe1986:
Support the following use.

```
$this->db->where('type', array(8, 9, 10));
$this->db->where('account IN', array('account751000872', 'account781000967'));
$this->db->where('phone_num NOT IN', array('13800000000'));
```

AND

```
$this->db->select('channel_id, count(1) as num');
$this->db->group_by('channel_id');
$this->db->having('num IN ', array(88, 888));
```

Indeed, I could use `where_in` and `where_not_in` in the first case. However, sometimes I prefer to use as follows.

```
$where = array('type' => 10, 
			'account IN' => array('account751000872', 'account781000967'), 
                        'login_token NOT IN' => array('13800000000')
                 );
$this->db->where($where);
```

The same with `having`.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -664,14 +664,20 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 
 			if ($v !== NULL)
 			{
+				$is_array_value = is_array($v);
 				if ($escape === TRUE)
 				{
-					$v = ' '.$this->escape($v);
+					$v = $this->escape($v);
+					if ($is_array_value)
+					{
+						$v = '('.implode(',', $v).')';
+					}
+					$v = ' '.$v;
 				}
 
 				if ( ! $this->_has_operator($k))
 				{
-					$k .= ' = ';
+					$k .= $is_array_value ? ' IN ' : ' = ';
 				}
 			}
 			elseif ( ! $this->_has_operator($k))
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -674,6 +674,10 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 					}
 					$v = ' '.$v;
 				}
+				elseif ($is_array_value)
+				{
+					$v = '('.implode(',', $v).')';
+				}
 
 				if ( ! $this->_has_operator($k))
 				{


narfbg:
`where()` already supports too many ways to do the same things and each next addition to it causes a headache for us - I can't remember the last time we changed something in there that didn't result in a regression.
This is primarily because it hadn't been restrictive enough to begin with and allowed everybody to develop their own understanding of how it should work, based on their own preference at the time. But we can't continue this forever - not least because we already have dedicated methods for `IN()`, and more importantly because it's bad to do this in the first place - we can't satisfy everybody's preferences.

Plus, this is more magic ...
We only accept scalars currently, because that's what we know are values we can translate into SQL. We can safely reject everything else because we *know* all other cases are erroneous, but with this patch this would no longer be the case - now virtually everything would have to be magically treated as valid input and raises a lot of questions:

- If an operator isn't present and the input is an array, your patch assumes we want an `IN()`, but do we? I'd argue that there's a 50/50 chance for that at best.
- If an operator is present, the input is an array and the operator isn't `(NOT)? IN` do we treat that as an error, or do we magically change the operator? There's a precedent for that with NULLs and both options have valid use cases, but again - we'd have to assume one over the other for what is again a 50/50 split.
- If we accept arrays, do we also accept objects?
- If we accept objects, do we try for `__toString()` before treating their properties as if they were array elements?
- If we bend-over backwards to treat literally every input as valid, do we merge multi-dimentional arrays until they become single-dimentional? At least this one will probably seem absurd to the majority of the public, but I assure you - there will be somebody who'd want to use this too, so why satisfy your preference and not theirs?
- At what point do we stop?

Magic is bad.

tianhe1986:
Thanks for explaining. Magic is bad :)

But, `CI_DB_driver::compile_binds` may also have the same questions.

And it seems I need to use `query()` for `HAVING(XXX) IN (XXX)`, do I?

narfbg:
I kind of get the feeling you'd want to add that only for the sake of having it, as opposed to a real need, but we could have `having_in()`, `having_not_in()` - that's a different thing.

tianhe1986:
> I kind of get the feeling you'd want to add that only for the sake of having it, as opposed to a real need

No,  In my project, I gather inputs from page as searching key and value, which are then put into one array . It would be more convenient  if I could just use this array as the input for `where()`, but now I need to loop the array, check the value type and determine which one should I use,  `where()` or `where_in()`.

> but we could have having_in(), having_not_in()

It is good news.

narfbg:
>> I kind of get the feeling you'd want to add that only for the sake of having it, as opposed to a real need
>
> No, In my project, I gather inputs from page as searching key and value, which are then put into one array . It would be more convenient if I could just use this array as the input for `where()`, but now I need to loop the array, check the value type and determine which one should I use, `where()` or `where_in()`.

I meant that for `having_in()` and `having_not_in()`. :)

tianhe1986:
>>> I kind of get the feeling you'd want to add that only for the sake of having it, as opposed to a real need

>> No, In my project, I gather inputs from page as searching key and value, which are then put into one array . It would be more convenient if I could just use this array as the input for where(), but now I need to loop the array, check the value type and determine which one should I use, where() or where_in().

> I meant that for having_in() and having_not_in(). :)

Oh  yes, you are quite right. It is just a possibility, forget it. :)

narfbg:
I'm just noting that adding stuff for the sake of having it is a bad idea in general. But in this case you are welcome to add that.

PR #4926, opened 2016-12-01
Changed Core\Config to check for array in base_url-item in case there is …

wackamole:
…defined more than 1 domain set for site.

### DIFF
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -106,7 +106,13 @@ class CI_Config {
 			{
 				$base_url = 'http://localhost/';
 			}
+			$this->set_item('base_url', $base_url);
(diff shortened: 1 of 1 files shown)


narfbg:
This is a lazy, bad practice IMO.

The fact that we give an [example in the userguide](https://codeigniter.com/userguide3/installation/upgrade_303.html#step-2-make-sure-your-base-url-config-value-is-not-empty) for it is for BC purposes, and because we know a lot of people will do worse on their own.
But baking-in this functionality would arguably encourage its usage, which is not what we want.

I'm leaving this open for now, if you want to try convincing us otherwise, but you'll need some strong arguments.

PR #4931, opened 2016-12-06
Update directory_helper.php

churavka:


narfbg:
Re-submitting #4908 won't make me merge it.
Actually replying to it may help.

PR #4938, opened 2016-12-09
Reset the position of the query result based on the function description so that we get all data out.

mrforsythexeter:


narfbg:
A bunch of drivers don't have `data_seek()` because they don't do prefetching ... What's the use case for this?

mrforsythexeter:
We are currently doing some work on the query before passing it to this function. It happened, that it meant that we always missed the first row in the data, this was a little embarrassing to find out. 

I understand this maybe an issue due to different cursors in different DB Engines.

narfbg:
What are you doing to it exactly? And why does it has to be done on the same resultset, yet keeping it intact? This looks very strange to me ...

Not that I see any immediate drawbacks from your proposed change, but I try very hard to avoid filling our codebase with work-arounds.

mrforsythexeter:
We have our own active record layer. As it happens we fixed it in our implementation layer (active record) rather than this function. When our active record is constructed it sets the internal pointer in the query to the first unbuffered row and retrieves it.

However I submitted the pull request, as I felt from documentation that it wasn't clear that this function would / could miss the start of the query.

narfbg:
... ok.

Issue #4892, opened 2016-10-31
Update batch wrongly escaping causing error


Rverm:
(Similar bug #4871 I think)

Started since v3.1.2: example code:
>$data = array(
			array(
				'setting_key' 	=> 'test',
				'setting_value' => 'test'
			)
		);
		$this->db->where('user_id', 1);
		$this->db->update_batch('settingstable', $data, 'setting_key');`

Returning

> ERROR - 2016-10-31 10:05:32 --> Severity: Notice --> Undefined index: setting_key system\database\DB_query_builder.php 1949
ERROR - 2016-10-31 10:05:32 --> Severity: Notice --> Undefined index: setting_key system\database\DB_query_builder.php 1955
ERROR - 2016-10-31 10:05:32 --> Severity: Notice --> Undefined index: setting_key system\database\DB_query_builder.php 1955

Line 1949

>$ids[] = $val[$index];

The items in $val are escaped and the $index search on non escaped name. I think the keys shouldn't be escaped?

ajdunn2:
I have come across the same issue with v3.1.2

Using the example data below:
```
        $data = array(
            array(
                'title' => 'My title' ,
                'name' => 'My Name 2' ,
                'date' => 'My date 2'
            ),
            array(
                'title' => 'Another title' ,
                'name' => 'Another Name 2' ,
                'date' => 'Another date 2'
            )
        );

        $this->db->update_batch('mytable', $data, 'title');
```

Produces the following incorrectly formed query

> UPDATE `mytable` SET `title` = CASE WHEN `title` = THEN 'My title' WHEN `title` = THEN 'Another title' ELSE `title` END, `name` = CASE WHEN `title` = THEN 'My Name 2' WHEN `title` = THEN 'Another Name 2' ELSE `name` END, `date` = CASE WHEN `title` = THEN 'My date 2' WHEN `title` = THEN 'Another date 2' ELSE `date` END WHERE `title` IN(,)

Prior to Fix #4871

> UPDATE `mytable` SET `name` = CASE WHEN `title` = 'My title' THEN 'My Name 2' WHEN `title` = 'Another title' THEN 'Another Name 2' ELSE `name` END, `date` = CASE WHEN `title` = 'My title' THEN 'My date 2' WHEN `title` = 'Another title' THEN 'Another date 2' ELSE `date` END WHERE `title` IN('My title','Another title')

ajdunn2:
[...]

PR #4942, opened 2016-12-13
use escaped $index

qdsay:
The $set's all key has escaped.


qdsay:
Here is a bug

narfbg:
Here is the original report: #4892

PR #4945, opened 2016-12-15
Small code simplification in character_limiter()

vlakoff:
Just noticed the bump of PHP minimum version :-)

So it should be a good moment to make the code here a bit less complicated. Follow-up to #1750, see notably this [comment](https://github.com/bcit-ci/CodeIgniter/pull/1750#issuecomment-8279512).


narfbg:
Long time no see, welcome back. :)

PR #4949, opened 2016-12-17
Modified Form_validation's  feature to accept ID param

MarkMatute:
Modified [is_unique] to accept a third optional parameter [id] to ignore.
example : is_unique[users.username.1]


narfbg:
No thanks.

PR #4955, opened 2016-12-22
Fix bug: DB_query_builder do not add parenthesis for cached group_by

boxsnake:
Every time executing a SQL sentence, the query builder will be emptied except the cached parts.

However, when using count_all_results method, the function does not test the cached "group_by" clause, in the case where group_by is emptied but cached group_by is there. This flaw results in mistaken count for results.

In this pull request, this is fixed.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1402,7 +1402,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			$this->qb_orderby = NULL;
 		}
 
-		$result = ($this->qb_distinct === TRUE OR ! empty($this->qb_groupby))
+		$result = ($this->qb_distinct === TRUE OR ! empty($this->qb_groupby) OR in_array('groupby', $this->qb_cache_exists))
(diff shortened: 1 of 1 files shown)


system/database/DB_query_builder.php
```diff
@@ -1402,7 +1402,7 @@ public function count_all_results($table = '', $reset = TRUE)
 			$this->qb_orderby = NULL;
 		}
 
-		$result = ($this->qb_distinct === TRUE OR ! empty($this->qb_groupby))
+		$result = ($this->qb_distinct === TRUE OR ! empty($this->qb_groupby) OR in_array('groupby', $this->qb_cache_exists))
```
tianhe1986: Maybe better using `! empty($this->qb_cache_groupby)`

boxsnake: Thanks for commenting.

boxsnake: Requested in another PR.

narfbg: You don't have to open a new PR for every single change - just push your commits to the same branch.

PR #4961, opened 2016-12-25
loaded_classes() function

ET-NiK:
If CI_Controller constructor called repeatedly, then arise error with loading libraries and drivers.
If add to config autoload:
```
$autoload['drivers'] = array('session');
```
Then we get error:
```
Unable to locate the specified class: Session.php
```
I detail described this problem here - https://github.com/ci-bonfire/Sprint/issues/177


narfbg:
Constructors are not supposed to run more than once.

ET-NiK:
Why? Why i can't create more than once CI_Constructor child objects?
[Sprint](https://github.com/ci-bonfire/Sprint) and [CI PHPUnit](https://github.com/kenjis/ci-phpunit-test) create several CI_Constructor objects. An error appears if there is a my library in the autoload config.

I can't change $_classes var in load_class() function, it is not good.

Issue #4963, opened 2016-12-26
PHPMailer - A critical vulnerability


natiq2004:
Please read: http://thehackernews.com/2016/12/phpmailer-security.html

Please update: /system/libraries/Email.php

rgcv:
Issue regarding the vulnerability: [PHPMailer#922](https://github.com/PHPMailer/PHPMailer/issues/922)

doodoori2:
Is it a related issue?

jim-parry:
How is this a CodeIgniter problem? CodeIgniter does not have any third party email packages, like PHPMailer, bundled with it.

Federkun:
`$this->email->from($this->input->post('email'));` is indeed vulnerable.

derekjones:
> `$this->email->from($this->input->post('email'));` is indeed vulnerable.

But only if you've intentionally turned off validation and aren't validating yourself, right? Which would be a self-inflicted vulnerability.

```php
$this->email->from('"attacker\" -oQ/tmp/ -X/var/www/cache/phpcode.php  some"@email.com');

// protected '_debug_msg' => 
//   array (size=1)
//     0 => string 'Invalid email address: "attacker\" -oQ/tmp/ -X/var/www/cache/phpcode.php  some"@email.com<br />' (length=95)
// protected '_recipients' => 
//   array (size=0)
//     empty
```

Federkun:
It may be not so obvious, 

    // this is a valid address
    var_dump(filter_var('"foo\"\ -bar/baz"@strange.example.com', FILTER_VALIDATE_EMAIL)); 

Even SQL injection are self-inflicted vulnerability, that's why any decent library shall not expect/pretend that the user provide parameters already escaped.

derekjones:
> that's why any decent library shall not expect/pretend that the user provide parameters already escaped.

Well it doesn't, it uses `filter_var()` to validate the values unless you explicitly turn it off. And it's unfortunate that `filter_var()` passes your example, but it is technically a valid address per the spec. The issue is whether or not there's a security vulnerability, which I'm honestly not able to reproduce with CI's Email library, which is not 1:1 with PHPMailer's use of `mail()`. Do you have a reproducible vector?

That said, I'm always for practical implementation of specs rather than literal. It might be worth adding the `FILTER_SANITIZE_EMAIL` flag to the validation, which would obviate all of this, with the rare negative consequence of failing on an uncommon, but technically valid email address.

[...]

PR #4966, opened 2016-12-30
prevent use of escaped arguments with mail/sendmail

derekjones:
Fixes issue #4963


narfbg:
Replaced with the linked commit.
Thanks either way; it's good to see you're still following us. :)

derekjones:
Of course! The security report caught my eye in my billion notifications, as they do.

PR #4968, opened 2017-01-03
Azure Web App Support for CodeIgniter

muhammad92:
Added Deploy to Azure button which would take the user to Azure and Deploy a Web App. It also deploys In App MYSQL Database when it deploys CodeIgniter.


gxgpet:
I don't think this will fit...

It's just a particular case, for your setup/project, for instance. CodeIgniter doesn't have to do anything with deployment.

muhammad92:
It's just and add on, would help users of Azure to deploy CodeIgniter on
single click.

andersonunsonst:
You are well-intentioned, but I suggest you create your own Repository, this is not the responsibility of the framework

narfbg:
Even if we put everything else aside, this is basically a request to promote Azure, and as a side-effect - MySQL too. We do not and will not promote any particular platform.

I don't even believe this is "well-intentioned", given that you want to put this on the very top of our README, above even our own project name.

And that you're converting it from rst to Markdown is just more annoying.

gxgpet:
@muhammad92 If you want to help other CI devs, you can always post your ideas and solutions to CI forums, in [Addins](https://forum.codeigniter.com/forum-13.html) section, for example. :)

muhammad92:
Apologies from my side. Was my first pull request, just wanted to value add what I could do.
If you really hate the idea of deployment on azure. I could be placed at bottom may be :) under heading of one click deployment. 

By the way thanks for your idea. I was never here to promote Azure or MYSQL but to just add something that could help many of the CI users. 

Thanks.

muhammad92:
@gxgpet  sure :)

narfbg:
You're free to blog about it, post about it on our [forums](https://forum.codeigniter.com/), publish it as a helper package on [Packagist](https://packagist.org/), etc.

But while I appreciate how it may make deployment easier for Azure users, it doesn't belong in our README.
Unless we provide equally convenient stuff for every other platform, doing this *would* mean that we are favoring/promoting Azure as a platform - that's just how it works.

PR #4970, opened 2017-01-04
editorconfig settings

arysom:
editorconfig settings may be helpful in keeping styleguide rules

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*]
+charset = utf-8
+
+# Tab indentation (no size specified)
+indent_style = tab


gxgpet:
This is not something related to the framework itself, in my opinion...

If you have any suggestions or ideas which may help other CI devs by improving their coding environment, you are most welcome to post and tell us about it on [CI forums](https://forum.codeigniter.com/). :)

arysom:
When I start new project maybe it can keep an eye on my coding practices. Editorconfig is unviersal kind of settings for many editors. If it's not good idea, you can close/ignore it, no problemo ;)

gxgpet:
I'm not saying it's not a good idea, I'm sure it can really be. Only that the place for it may not be here, along CI basecode. I'm just trying to help you make your solution known and used by CI community. :)

gxgpet:
[EditorConfig](http://editorconfig.org/) is not widely adopted and used by all editors out there, that was my point. In this case, should we include configs for other editors as well?

narfbg:
No, of course we won't accept configs for "other editors".

You're equating editorconfig to an editor, which is a gross over-simplification of the fact that not every major IDE has adopted support for it, yet. It is still the closest thing to a universal, editor-agnostic configuration in existence.

PR #4978, opened 2017-01-12
Add dump_cache() & restore_cache() to CI_DB_query_builder

boxsnake:
Add dump_cache() & restore_cache() to CI_DB_query_builder to enable temporary QB cache save.
 
Usage:

**Before flush:**
```
    $qb_cache = $this->db->dump_cache();
```

**Then, fell free to flush your cache:**
```
    $this->db->flush_cache();
```
**You do not have to flush the cache, you can also append some other conditions, and then restore with the save cache.**

**And after you make some other get (SELECT) calls then restore from your saved QB cache:**
```
    $this->db->restore_cache($qb_cache);
```

**And then, you can continue with other get (SELECT) calls with previously cached criteria.**

That's all this done.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -2642,7 +2642,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 */
 	public function dump_cache()
     {
-        return [
+        return array(
             'qb_cache_select'    => $this->qb_cache_select,
             'qb_cache_from'      => $this->qb_cache_from,
             'qb_cache_join'      => $this->qb_cache_join,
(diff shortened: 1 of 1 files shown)


narfbg:
I don't mind the idea in general, but there's a technical detail that makes it a no-go ... the `qb_*` properties are internal and meant to be inaccessible from userspace.

This proposal doesn't just make them accessible (we've rejected this in the past for people who just want to read them), it makes them *writable*. And that in turn means we can never assume they're not in a broken state.

boxsnake:
What if I cache it with a "tag", then restore it with the tag?

narfbg:
Don't know what you mean by that, but no - the cache in question is a horrible feature to begin with. I'd rather remove it than expand on it.

Issue #3478, opened 2015-01-13
valid_url() should validate a well formed URL not any string


adrianorsouza:
I think I'm not the only one facing this question https://github.com/bcit-ci/CodeIgniter/issues/3118 .

The method `valid_url()` in Form_validation library allow users to bypass any string instead of check whether is a valid url or not. As the opposite of function's name say it doesn't work as expected if the intention is to validate an url.

I don't consider it's a bug but maybe a misbehaviour of this method.

Imagine that you have a form with an input field named `txt_url` for users to fill in with an URL value, if the user type any string like `asdfasdf` the method `valid_url()`, should IMHO, check whether the user input is a valid URL format or not. As what I expected. But what it actually does is to get any input string prepend a [scheme][1] and pass its value through `filter_val()` function which will always evaluate to true.

I tested it out using the following code.

Controller:

```php
$this->form_validation->set_rules('txt_url', 'Url', 'trim|required|valid_url');
```

View: 
```html
<input type="text" name="txt_url" value="asdfasdf">
```

In my concept per RFC 2396 a valid URL should be:

    <scheme name> : <hierarchical part> [ ? <query> ] [ # <fragment> ]

The method's behaviour that I expected is the same as `valid_email()` but acting with URL values instead.

I think this method should be refactored and return a bool whether a given string is a well-formed URL, it could use a different pattern to check a given str format or maybe just return the built-in function `filter_var` e.g:

```php
public function valid_url($str)
{
    return (filter_var($str, FILTER_VALIDATE_URL) !== FALSE);
}
```

Anyway I'm able to refactor if if it's necessary but I'd like to hear from you guys if my thoughts makes any sense.

[1]: https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Form_validation.php#L1229

narfbg:
<pre>
# echo '127.0.0.1 asdfasdf' >> /etc/hosts
# echo 'test' > /var/www/html/index.html
# wget asdfasdf
--2015-01-14 01:29:10--  http://asdfasdf/
Resolving asdfasdf (asdfasdf)... 127.0.0.1
Connecting to asdfasdf (asdfasdf)|127.0.0.1|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 5 [text/html]
Saving to: ‘index.html’

100%[====================================================================================================================================================================================================>] 5           --.-K/s   in 0s      

[...]

PR #4979, opened 2017-01-13
Add second parameter 'scheme' to valid_url.

Red54:
This commit can let the rule 'valid_url' only accept valid URLs with the schemes.

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -1192,7 +1192,7 @@ class CI_Form_validation {
 	 * @param	string	$str
 	 * @return	bool
 	 */
-	public function valid_url($str)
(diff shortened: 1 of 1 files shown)


narfbg:
Referencing #3478.

@Red54 While rules can have additional parameters, them being boolean is not a good idea as a lot (if not most) users provide them as a pipe-separated string. If anybody wanted to be explicit for whatever reason and pass `false` that way, this happens:

    var_dump((bool) 'false');
    php shell code:1:
    bool(true)

Thus, it would be better to have this as a separate method.

---

On a side note: Your description implies that 'valid_url' currently accepts invalid URLs. This is not true and I will reject any future PRs that do the same.

PR #4986, opened 2017-01-16
spelling fixes

ka7:
patch contains some spelling fixes ( just in comments ) as found by a bot ( https://github.com/ka7/misspell_fixer )


jim-parry:
These make sense, except for the change to Security.php ... "preced" is just as wrong as "preceed" ... it should be "precede".

jim-parry:
@narfbg I am fine with this.

PR #4987, opened 2017-01-17
Add aliased tables cache in query_builder.

tianhe1986:
Set `$_protect_identifiers` to `TRUE` and also set `dbprefix` to `ok_`.

Then I used the following code:
```
$this->db->start_cache();
$this->db->select('us.name');
$this->db->from('user as us');
$this->db->stop_cache();
$sql = $this->db->get_compiled_select('');
echo $sql."\n";
$sql = $this->db->get_compiled_select('');
echo $sql."\n";
```

and got

```
SELECT `us`.`name` FROM `ok_user` as `us`
SELECT `ok_us`.`name` FROM `ok_user` as `us`
```

The table prefix was not properly processed in `CI_DB_query_builder::_merge_cache()`. 

And the similar situation when using qb cache with alias in JOIN statement.  For example,
```
$this->db->start_cache();
$this->db->select('user.name, jo.description');
$this->db->from('user');
$this->db->join('job as jo', 'user.name = jo.name');
$this->db->stop_cache();
$sql = $this->db->get_compiled_select('');
echo $sql."\n";
$sql = $this->db->get_compiled_select('');
echo $sql."\n";
```
And got
```
SELECT `ok_user`.`name`, `jo`.`description` FROM `ok_user` JOIN `ok_job` as `jo` ON `ok_user`.`name` = `jo`.`name`
SELECT `ok_user`.`name`, `ok_jo`.`description` FROM `ok_user` JOIN `ok_job` as `jo` ON `ok_user`.`name` = `jo`.`name`
```
Because $qb_aliased_tables was reset to empty array and not be reprocessed when merging JOIN cache.

So I add aliased tables cache in this PR to fix them.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -215,7 +215,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	protected $qb_cache_join			= array();
 
 	/**
-	 * QB Cache aliased tables list data
+	 * QB Cache aliased tables list
 	 *
 	 * @var	array
 	 */
@@ -2288,10 +2288,10 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			$table = trim(strrchr($table, ' '));
 
 			// Store the alias, if it doesn't already exist
-			if ( ! in_array($table, $this->qb_aliased_tables))
+			if ( ! in_array($table, $this->qb_aliased_tables, TRUE))
 			{
 				$this->qb_aliased_tables[] = $table;
-				if ($this->qb_caching === TRUE && ! in_array($table, $this->qb_cache_aliased_tables))
+				if ($this->qb_caching === TRUE && ! in_array($table, $this->qb_cache_aliased_tables, TRUE))
 				{
 					$this->qb_cache_aliased_tables[] = $table;
 					$this->qb_cache_exists[] = 'aliased_tables';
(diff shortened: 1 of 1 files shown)


system/database/DB_query_builder.php
```diff
@@ -215,6 +215,13 @@
 	protected $qb_cache_join			= array();
 
 	/**
+	 * QB Cache aliased tables list data
```
narfbg: Wouldn't care for this that much, but there are other changes to be made as well, so ... Keep either 'list' or 'data', one is superfluous.

system/database/DB_query_builder.php
```diff
@@ -215,6 +215,13 @@
...
+	 *
+	 * @var	array
+	 */
+	protected $qb_cache_aliased_tables			= array();
```
narfbg: Just `$qb_cache_aliases`.

system/database/DB_query_builder.php
```diff
@@ -2284,6 +2291,11 @@ protected function _track_aliases($table)
 			if ( ! in_array($table, $this->qb_aliased_tables))
 			{
 				$this->qb_aliased_tables[] = $table;
+				if ($this->qb_caching === TRUE && ! in_array($table, $this->qb_cache_aliased_tables))
```
narfbg: Always set `in_array()`'s third parameter to `TRUE`.

system/database/DB_query_builder.php
```diff
@@ -2616,16 +2628,17 @@ public function stop_cache()
...
-			'qb_cache_from'			=> array(),
-			'qb_cache_join'			=> array(),
-			'qb_cache_where'		=> array(),
-			'qb_cache_groupby'		=> array(),
-			'qb_cache_having'		=> array(),
-			'qb_cache_orderby'		=> array(),
-			'qb_cache_set'			=> array(),
-			'qb_cache_exists'		=> array(),
-			'qb_cache_no_escape'	=> array()
+			'qb_cache_select'			=> array(),
```
narfbg: Don't re-align.

system/database/DB_query_builder.php
```diff
@@ -2676,13 +2689,6 @@ protected function _merge_cache()
 				$this->qb_no_escape = $qb_no_escape;
 			}
 		}
-
-		// If we are "protecting identifiers" we need to examine the "from"
```
narfbg: Why is this removed?

system/database/DB_query_builder.php
tianhe1986: En,  this is because QB aliased tables list named `$qb_aliased_tables`,  and in `_merge_cache()` ,
```
$qb_variable	= 'qb_'.$val;
$qb_cache_var	= 'qb_cache_'.$val;
```

system/database/DB_query_builder.php
tianhe1986: There is no need to re-examine when using alias cache. They have been put into `$qb_aliased_tables`.

system/database/DB_query_builder.php
narfbg: I see ... ok.

PR #4992, opened 2017-01-23
Removing index.html files in application and system folders

DiederikLascaris:
I would like to propose to remove the index.html files in the application and system (sub)folders, or at least have some discussion about this.

Directory listing is disabled by default in most used webservers: Apache, NGINX, Windows IIS and LiteSpeed. I haven't seen any hosting provider that had directory listing enabled for its clients in the last 5 years or so. Before that time I have. 

To me all those index.html files are just dead space and sometime I get annoyed by their existence. Besides that they are unnecessary (IMHO) due to the (nowadays) default, in the case of Apache users the access to the folders itself is already blocked in the /application/.htaccess and /system/.htaccess. 

So even if some ignorant user has "+Indexes" in his main .htaccess file a curious visitor (or hacker if you will) will not see the contents of the index.html file at all. The same applies if the user uses some kind of Linux distro that has '+Indexes' setup by default. My system of choice (Debian) follows the Apache defaults and has it disabled, but I found that some distros (Gentoo and Centos for example) are shipped with a '+Indexes' setup.

Some research about the default settings:
http://httpd.apache.org/docs/trunk/mod/core.html#options
http://nginx.org/en/docs/http/ngx_http_autoindex_module.html
https://technet.microsoft.com/en-us/library/cc731109(v=ws.10).aspx
https://www.litespeedtech.com/support/forum/threads/solved-directory-indexing.14227/

Distro specific:
https://www.centos.org/docs/5/html/Deployment_Guide-en-US/s1-apache-config.html#s2-apache-options
https://wiki.gentoo.org/wiki/Apache#.2Fetc.2Fapache2.2Fhttpd.conf


narfbg:
Not the first time someone wanted to do that. Here's my opinion: https://github.com/bcit-ci/CodeIgniter/pull/1986#issuecomment-10181315

DiederikLascaris:
Yes, I saw your reply in that request. That request include the idea to move the application and system folder outside the public accessible document root. You also mentioned you probably would not add them again if they didn't exist already. We are a 4 years further down the road now so I thought why not rethink about this again. 

I honestly can't think of a real world scenario where someone would benefit from this unless they deliberately done some other crazy stuff (like deleting the .htaccess file for example or set AllowOverride to none for Apache or enable directory listing for other web servers). And even then it's only a security through obscurity type of solution...

narfbg:
> I honestly can't think of a real world scenario where someone would benefit from this unless they deliberately done some other crazy stuff (like deleting the .htaccess file for example or set AllowOverride to none for Apache or enable directory listing for other web servers).

You have no idea of the crazy environments CI runs on ...

> And even then it's only a security through obscurity type of solution...

True. But that loops back to what I said in the linked comment - there's zero drawback to keeping it.

DiederikLascaris:
Just thought it might be time to get rid of them. I see it as a somewhat bad (security through obscurity) second layer of defense for when people have totally no idea what their doing. I would blame the badly designed environment instead of a framework when the directory listing is showed.

But I understand your point and understand if you close the request. The only downside (besides my own annoyance) is some disk space.

narfbg:
Well, again - it's indeed of very little value, but it's already in there and has zero negatives outside of personal preferences ... it's a stalemate.

What usually happens in that case is that I have no objective reason to take either side, so the PR would sit open forever. Just to avoid that scenario, I'll leave this for @jim-parry, @lonnieezell and @benedmunds to decide with a 2 out of 3 vote.

lonnieezell:
I see zero reason not to keep it. And I think @DiederikLascaris last comment explains exactly why: 

> I see it as a somewhat bad (security through obscurity) second layer of defense for when people have totally no idea what their doing. I would blame the badly designed environment instead of a framework when the directory listing is showed.

Many CI users seem to rather new to programming, so I guarantee many are new to server management and mis-configuration can happen. Heck, it happened to Facebook once a few years back, so accidents happen to the best of us. 

> The only downside (besides my own annoyance) is some disk space.

And the disk space and inode count is very, very minimal.

If we had a structure that kept the index file in a `public` folder, like CI4 and most other frameworks, I'd agree we should kill them, but the way the default directory structure is in CI3, all of the directories will likely be in web root and accessible, so a second layer of (hopefully never needed) defense doesn't hurt here.

PR #4994, opened 2017-01-24
add attribute in form_dropdown in form_helper.php

lloricode:
I add some function of form_dropdown in form_helper.
that can add additionally attribute in `<option>`

**here is a sample usage:**

```
echo form_dropdown(
                'child_count',
                array(
                    '1'=>'a',
                    ''=>array('data'=>'test1','class'=>'imclass','id'=>''),
                    array('data'=>'test2','id'=>'imid()','value'=>'imval'),
                 ),
                set_value('child_count'),
                array('class' => 'cs-select cs-skin-elastic')
        );
```

**then the output:**

```
<select name="child_count" class="cs-select cs-skin-elastic">
     <option value="1">a</option>
     <option value="" class="imclass" id="" selected="selected">test1</option>
     <option  id="imid()" value="imval">test2</option>
</select>
```

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -419,16 +419,49 @@ if ( ! function_exists('form_dropdown'))
 					continue;
 				}
 
-				$form .= '<optgroup label="'.$key."\">\n";
+                                if(array_key_exists('data', $val))
+                                {
+                                        $attributes_ = array();
+                                        
+                                        foreach ($val as $k_=>$v_)
+                                        {
+                                            $k_ = (string) $k_;
+                                            
+                                            if($k_ === 'data')
+                                            {
+                                                continue;
+                                            }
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but no ...

Regardless of how you design or implement this, it would be inconsistent with the current/historical `$valueAttribute> => $displayText` pairing. If those were flipped, it probably would be achievable as then we could have:

    $displayText => [
        'value' => $value,
        'anotherAttributeName' => 'blahblah'
    ];

... which is the only sane thing. `data` is a very confusing key to denote the text, and would collide with a possible `data` attribute.

Also, the patch is pretty ugly and the feature itself is at best sparingly useful. The 90% case is CSS classes/ids/inline-style, which is easily (and better) done in other ways.

PR #5009, opened 2017-02-06
Do not send secure cookie if not https

appelflap:
Like in CI_Security::csrf_set_cookie.
Otherwise cookie will be exposed over http and can be used from client side to server over https

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -318,6 +318,13 @@ class CI_Input {
 			}
 		}
 
+		// Do not send secure cookie if not https
+		$secure_cookie = (bool) config_item('cookie_secure');
+		if ($secure_cookie && ! is_https())
+		{
+			return;
+		}
+		
 		if ($prefix === '' && config_item('cookie_prefix') !== '')
 		{
 			$prefix = config_item('cookie_prefix');
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -340,7 +340,7 @@ class CI_Input {
 			$path = config_item('cookie_path');
 		}
 
-		if ($secure === FALSE && config_item('cookie_secure') === TRUE)
+		if ($secure === FALSE && $secure_cookie === TRUE)
 		{
 			$secure = config_item('cookie_secure');
 		}


system/core/Input.php
```diff
@@ -318,6 +318,13 @@ public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path
 			}
 		}
 
+		// Do not send secure cookie if not https
+		$secure_cookie = (bool) config_item('cookie_secure');
```
narfbg: The variable should be reused ...

tianhe1986:
@appelflap 
I tried `set_cookie` when setting param `$secure` to `TRUE` with HTTP,  it won't set anything.
`set_cookie` calls `setcookie` in the end, and will do nothing in this case according to the [manual](http://php.net/manual/en/function.setcookie.php) .

appelflap:
That's indeed what you would expect. But tried with Apache/2.4.23 (Win32) PHP/7.0.11 on a non-ssl connection. Break in PhpStorm showed:
```
$_SERVER['HTTPS'] = null
$_SERVER['REQUEST_SCHEME'] = "http"
```

Request Header:
```
Cookie	__utma=222597775.318619583.1468571105.1468571105.1468571105.1; __utma=11747552.2055094958.1468571199.1475061281.1475142571.54; __utmz=11747552.1468571199.1.1.utmcsr=(direct)|utmccn=(direct)|utmcmd=(none)
```

Response Header:
```
Set-Cookie	
db_session=deleted; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; path=/; domain=my.domain.com
db_session=2e4df5c9c206e5edc5963d685ac7f406; expires=Tue, 07-Feb-2017 11:59:42 GMT; Max-Age=9600; path
=my.domain.com; domain=/; secure
```

It does say 'secure' at the end of the line, but still. That session token is in the db. So it has been exposed over non-ssl and could be used over ssl.

narfbg:
@tianhe1986 That's the point.

@appelflap Did *you* send that `db_session` cookie using `CI_Input::set_cookie()`, or is it generated by `CI_Session`? The latter doesn't use the former.

appelflap:
We use a custom session library that I forced to use CI_Input::set_cookie() function. The cookie is not usable in the browser. So it is not possible to actually log-in, but still, it is exposed.

narfbg:
Even if you're able to force `CI_Input::set_cookie()` usage in some places, you can't really override it. PHP itself generates and sends the session cookies.

tianhe1986:
@appelflap 
Yeah, you are right. I checked only cookie but forgot session.

PR #5010, opened 2017-02-07
Allow to omit trailing slash in config paths

vlakoff:
This always has bugged me :)

Let's be kind to the developer, don't break everything if he provided a path without trailing slash.

### DIFF
--- a/system/core/Exceptions.php
+++ b/system/core/Exceptions.php
@@ -161,6 +161,10 @@ class CI_Exceptions {
 		{
 			$templates_path = VIEWPATH.'errors'.DIRECTORY_SEPARATOR;
 		}
+		else
+		{
+			$templates_path = rtrim($templates_path, '/\\').DIRECTORY_SEPARATOR;
+		}
 
 		if (is_cli())
 		{
@@ -193,6 +197,10 @@ class CI_Exceptions {
 		if (empty($templates_path))
 		{
 			$templates_path = VIEWPATH.'errors'.DIRECTORY_SEPARATOR;
+		}
+		else
+		{
+			$templates_path = rtrim($templates_path, '/\\').DIRECTORY_SEPARATOR;
 		}
 
 		$message = $exception->getMessage();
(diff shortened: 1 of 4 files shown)


system/core/Exceptions.php
```diff
@@ -157,10 +157,7 @@ public function show_404($page = '', $log_error = TRUE)
 	public function show_error($heading, $message, $template = 'error_general', $status_code = 500)
 	{
 		$templates_path = config_item('error_views_path');
-		if (empty($templates_path))
-		{
-			$templates_path = VIEWPATH.'errors'.DIRECTORY_SEPARATOR;
-		}
+		$templates_path = ( ! empty($templates_path)) ? rtrim($templates_path, '/\\').'/' : VIEWPATH.'errors/';
```
narfbg: Inverted conditions like this one always bug me. ;)
Switching the sides will strip 5 characters -> less cognitive overhead

Also, please keep using `DIRECTORY_SEPARATOR`.

Same things apply to other similar lines in the patch.

vlakoff: About the cognitive overhead, I'm following the rule "value if set, else default value", which I think is the most pleasant. (note I also tweaked Output.php to uniformize on this rule)

About the `DIRECTORY_SEPARATOR`, it brings no benefit at all. Windows handles `/` as well, and it's much shorter.

narfbg: Well, I follow an "avoid unnecessary negations" rule and so does most of our code.

And yes - I'm aware that Windoes handles `/` just fine, but `DIRECTORY_SEPARATOR` is the semantically correct thing to use, and we do use that practically everywhere.  
I may've rejected a few patches in the past that tried to change `/` to `DIRECTORY_SEPARATOR`, but only because they did nothing else and thus were unnecessary changes. This is different.

vlakoff:
I have amended the commit.

narfbg:
Great, thanks!

vlakoff:
Can you add a changelog entry? Something like:
> Changed handling of `$config['log_path']`, `$config['error_views_path']` and `$config['cache_path']` to allow omitting the trailing slash.

narfbg:
Yea, I will ... Undecided yet if I want to merge it into 3.1-stable or keep it for 3.2.

vlakoff:
Just to point out, the change is not [*forward* compatible](https://en.wikipedia.org/wiki/Forward_compatibility).

narfbg:
Stays in develop only.

PR #5016, opened 2017-02-11
Increase version in userguide to 3.1.4-dev

zploskey:
The version number was not increased after the 3.1.3 release.


jim-parry:
We only accept contributions into the develop branch. The 3.1-stable branch is updated as part of release prep.

narfbg:
We do accept contributions in 3.1-stable ...

Although this particular thing is indeed handled during release preparations, that's just when the '-dev' suffix is removed.

jim-parry:
My misunderstanding, sorry!

PR #5017, opened 2017-02-13
DB_query_builder: Adding having_in(), or_having_in(), having_not_in() and or_having_not_in()

tianhe1986:
For generating SQL query such as 

`SELECT country, count(1) FROM user GROUP BY country HAVING count(1) IN(2, 3) OR country NOT IN('US', 'UK')`

### DIFF
--- a/tests/codeigniter/database/query_builder/group_test.php
+++ b/tests/codeigniter/database/query_builder/group_test.php
@@ -56,11 +56,11 @@ class Group_test extends CI_TestCase {
 	public function test_having_in()
 	{
 		$jobs = $this->db->select('name')
-					->from('job')
-					->group_by('name')
-					->having_in('SUM(id)', array(1, 2, 5))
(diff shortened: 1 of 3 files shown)


tests/codeigniter/database/query_builder/group_test.php
```diff
@@ -48,4 +48,74 @@ public function test_having_by()
...
 
+	// ------------------------------------------------------------------------
+
+	/**
+	 * @see ./mocks/schema/skeleton.php
+	 */
+	public function test_having_in()
+	{
+		$jobs = $this->db->select('name')
+					->from('job')
```
narfbg: Waaay more indentation than needed.

Other than this, the patch looks fine.

narfbg:
Still too much indentation, but I guess that's because the rest of that code gives a bad example ... I'll fix that myself later.

Issue #173, opened 2011-08-19
Multiple files upload


bitbucket-import:
Hi!

For one of projects I needed multiple files upload so I extended CI_Upload with two functions: do_multiple_upload() and get_data() (that are equivalents of do_upload() and data() respectively but for multiple files). My extension uses dirty hack with substituting $_FILES contents and calling do_upload(), but that's the best way I can think of (duplicating whole class or just do_upload() is much worse).

Maybe that feature is important enough to add to CI, or maybe someone would just find my code useful, so here it is:

```php
# !php

<?php 
class MY_Upload extends CI_Upload {

private $multiple_upload_data = array();

/**
 * Perform multiple file upload
 *
 * @return  bool
 */
public function do_multiple_upload($field = 'userfile[]')
{
    $files = $_FILES; // we would change $_FILES so we need a copy
    $flag = false; // flag indicates if at least one upload was successful
    $fld = substr($field, 0, strlen($field)-2); // cutting out '[]' at the end
    for($i = 0; $i < count($files[$fld]['name']); $i++) { // for each file
        // setting up $_FILES so it looks like we uploaded single file
        foreach($files[$fld] as $attr => $values) {
            $_FILES[$fld][$attr] = $values[$i];
        }
        // performing upload and updating flag
        $res = $this->do_upload($fld);
        $flag = $flag || $res;
        // saving upload's data (it can later be fetched using get_data()
        array_push($this->multiple_upload_data, $this->data());
    }
    return $flag;
}

// --------------------------------------------------------------------

/**
 * Returns data about all uploaded files as data() method does
 *
 * @return  array
 */
public function get_data()
{
    return $this->multiple_upload_data;
}

// --------------------------------------------------------------------


}

?>
```

Thank you for developing Code Igniter!

davidgv88:
I think codeigniter must have a multiple upload in the next release?

Te class CI_Upload  should have do_upload() method and do_multiple_upload() method.

Once uploaded files $this->upload->data() method should return an associative array with the information of each file.

pleaseremove:
Why has this been made "won't fix"? It still needs fixing. I can't believe I am still having to extend the upload class...

If people are struggling to write this here is a nice example: https://github.com/avenirer/MY_Upload

davidgv88:
+1 @pleaseremove

webeasystep:
[...]

PR #5019, opened 2017-02-14
create library Uploads.php

codeigniters:
library upload file or multi files in codeigniter

https://github.com/codeigniters/Uploads


narfbg:
No; see #173.

PR #5022, opened 2017-02-15
Do not fail if the array pointer is after the element

vlakoff:
We should not assume the passed array has its pointer set to the first element.

### DIFF
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -77,6 +77,7 @@ if ( ! function_exists('force_download'))
 					return;
 				}
 
+				reset($filename);
 				$filepath = key($filename);
 				$filename = current($filename);
 


narfbg:
Can just replace the `current()` call with `reset()` and move it before the `key()` one. :)

vlakoff:
Nope, because you're overwriting `$filename` ;)

narfbg:
Eh ... you're right.

And every other solution is a dead end because PHP 7 somehow managed to be worse than PHP 5 in this case. :(

gxgpet:
Since I made some submissions in this area (Download helper) as well, you might want to check the above lines from `reset()` :)

```
if (count($filename) !== 1)
{
	return;
}
```

narfbg:
Single count array doesn't mean its array pointer can't be moved, which is why I hate this even more.

gxgpet:
Oh, yeah, you are right here. I didn't see it in a procedural way of thinking.

PR #4797, opened 2016-09-06
Added the ability to have uploaded images automatically resized

nibblebits:
Take the following code: 
`    $config['upload_path'] = './uploads/';
        $config['allowed_types'] = 'gif|jpg|png';
        $config['max_size'] = '100';
        $config['max_width'] = '128';
        $config['max_height'] = '128';
        $config['auto_resize'] = 'true';

        $this->upload->initialize($config);
        if (!$this->upload->do_upload('image_file')) {
            $data = array('error' => $this->upload->display_errors());
            $this->load->view('test', $data);
        } else {
            die('Upload Success');
        }`

With auto_resize set to true any image uploaded thats bigger than 128 by 128 will be resized to 128 by 128.

PR #5024, opened 2017-02-16
Automatically rotate uploaded images if needed

DiederikLascaris:
I have added the functionality to the Upload library to automatically rotate an image when needed. 

Some camera's (like an Iphone for example) store the way you hold the camera as Exif Orientation metadata inside the JPEG file. Most image viewing software reads this metadata and turns the image for you when opening it. 
Also some basic image viewing tools support rotating images, but instead of changing the image pixels they 'only' change that same metadata.

The addition will prevent uploading an image that will look sideways on the web (since browsers ignore this rotation metadata), but normal on the computer before uploading.

I tested this addition with sample images from https://github.com/recurser/exif-orientation-examples and some of my own photo's I took with my Ipad.

I have set if up so you can enable/disable the feature and control the quality of the rotated image through a regular config array:

```
if (!empty($_FILES)) {

    $config['upload_path']      = './upload/';
    $config['allowed_types']    = 'gif|jpg|png';
    $config['file_ext_tolower'] = true;
    $config['overwrite']        = true;

    $config['auto_rotate']      = true;
    $config['rotation_quality'] = 90;

    $this->load->library('upload', $config);

    if ( ! $this->upload->do_upload('userfile'))
    {
        $data['error'] = $this->upload->display_errors();
    }
}
```

### DIFF
--- a/system/libraries/Upload.php
+++ b/system/libraries/Upload.php
@@ -245,6 +245,20 @@ class CI_Upload {
 	public $mod_mime_fix = TRUE;
 
 	/**
+	 * Auto rotage flag
+	 *
+	 * @var	bool
+	 */
+	public $auto_rotate = TRUE;
+
+	/**
+	 * Auto rotage JPEG quality
+	 *
+	 * @var	int
+	 */
+	public $rotation_quality = 95;
+
+	/**
 	 * Temporary filename prefix
 	 *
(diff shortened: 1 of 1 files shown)


narfbg:
Same thing as #4797 - image manipulation is not `CI_Upload`'s job.

PR #5025, opened 2017-02-20
Introduce APCu driver

katsew:
I recently work on PHP7 and APCu with CodeIgniter, but there is no built-in APCu driver.
So I create APCu cache driver and [Prefix]_Cache, but I want to use it as a built-in driver.

This is compatible with existing driver usage.

```php
$this->load->driver('cache', 
    array('adapter' => 'apcu')
);
$bar = 'bar'
$this->cache->save('foo', $bar);

$this->cache->get('foo') // output 'bar'
```

### DIFF
--- a/system/libraries/Cache/drivers/Cache_apcu.php
+++ b/system/libraries/Cache/drivers/Cache_apcu.php
@@ -1,0 +1,221 @@
+<?php
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014 - 2017, British Columbia Institute of Technology
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @package	CodeIgniter
+ * @author	EllisLab Dev Team
+ * @copyright	Copyright (c) 2008 - 2014, EllisLab, Inc. (https://ellislab.com/)
+ * @copyright	Copyright (c) 2014 - 2017, British Columbia Institute of Technology (http://bcit.ca/)
+ * @license	http://opensource.org/licenses/MIT	MIT License
+ * @link	https://codeigniter.com
+ * @since	Version 2.0.0
+ * @filesource
+ */
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+/**
+ * CodeIgniter APC Caching Class
+ *
+ * @package		CodeIgniter
+ * @subpackage	Libraries
+ * @category	Core
+ * @author		EllisLab Dev Team
+ * @link
+ */
+class CI_Cache_apc extends CI_Driver {
+
+	/**
+	 * Class constructor
+	 *
+	 * Only present so that an error message is logged
+	 * if APC is not available.
+	 *
+	 * @return	void
+	 */
+	public function __construct()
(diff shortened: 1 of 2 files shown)


system/libraries/Cache/drivers/Cache_apcu.php
```diff
@@ -0,0 +1,221 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @package	CodeIgniter
+ * @author	EllisLab Dev Team
+ * @copyright	Copyright (c) 2008 - 2014, EllisLab, Inc. (https://ellislab.com/)
+ * @copyright	Copyright (c) 2014 - 2017, British Columbia Institute of Technology (http://bcit.ca/)
+ * @license	http://opensource.org/licenses/MIT	MIT License
+ * @link	https://codeigniter.com
+ * @since	Version 2.0.0
```
narfbg: If accepted, this would go into 3.2.0 ...

system/libraries/Cache/drivers/Cache_apcu.php
```diff
@@ -0,0 +1,221 @@
...
+ */
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+/**
+ * CodeIgniter APCu Caching Class
+ *
+ * @package		CodeIgniter
+ * @subpackage	Libraries
+ * @category	Core
+ * @author		EllisLab Dev Team
```
narfbg: I know you've just copied this, but it's incorrect.

Historically, most of the code is EllisLab's.
But you're adding new code ... if you want to attribute it to our team, it should say "CodeIgniter Dev team", but ultimately - you are the author here.

system/libraries/Cache/drivers/Cache_apcu.php
```diff
@@ -0,0 +1,221 @@
...
+ * @link
```
narfbg: Why is this empty?
If there's nothing to put in it - don't add it at all.

system/libraries/Cache/drivers/Cache_apcu.php
```diff
@@ -0,0 +1,221 @@
...
+
+		if ($success === FALSE OR count($stored) !== 3)
+		{
+			return FALSE;
+		}
+
+		list($data, $time, $ttl) = $stored;
+
+		return array(
+			'expire'	=> $time + $ttl,
```
narfbg: Do alignment (not indentaion) with spaces, not tabs.

system/libraries/Cache/drivers/Cache_apcu.php
```diff
@@ -0,0 +1,221 @@
...
+	 * @param	bool	$raw	Whether to store the raw value
+	 * @return	bool	TRUE on success, FALSE on failure
+	 */
+	public function save($id, $data, $ttl = 60, $raw = FALSE)
+	{
+		$ttl = (int) $ttl;
+
+		return apcu_store(
+			$id,
+			($raw === TRUE ? $data : array(serialize($data), time(), $ttl)),
```
narfbg: If it can handle arrays, why do you need `serialize()`?

system/libraries/Cache/drivers/Cache_apcu.php
```diff
@@ -0,0 +1,221 @@
...
+
+	// ------------------------------------------------------------------------
+
+	/**
+	 * Cache Info
+	 *
+	 * @param	bool	Whether to exclude the individual list of cache entries
+	 * @return	mixed	array on success, false on failure
+	 */
+	 public function cache_info($limited = FALSE)
```
narfbg: The Cache library is an abstraction ... this `$limited` parameter isn't used in any other driver, so it shouldn't exist here either.

system/libraries/Cache/drivers/Cache_apcu.php
katsew: @narfbg 

I don't get this feedback.
I follow the implementation of the `Cache_apc.php`.
It serialize `$data` on `save`, then undeserialize on `get` or `get_metadata`
`$data`would be row data or structured array like below.

```php

array(
  'some data here',
  timestamp,
  specified time to live
)
```
see: https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Cache/drivers/Cache_apc.php#L110

katsew:
@narfbg 

Thanks for the review and spare your time!
I fixed the code and comment on the question.
Please review again, thank.

system/libraries/Cache/drivers/Cache_apcu.php
narfbg: You should follow other implementations blindly ...

katsew: Because apc and apcu is nearly compatible, I don't want to change the returning value structure.
Users confused by changing behaviour.

katsew: @narfbg 

Sorry, I misunderstood your feedback.
I'll fix it.

narfbg: Users use the Cache library, not `serialize()` and `unserialize()` directly.

If there's a reason to use `serialize()` here - fine, but you have to know *why* and not just copy it from somewhere else.
If you don't know why it is in the 'apc' driver, but you're copying it here regardless, then you cannot explain your "own" code and I cannot accept your patch ...

katsew: @narfbg 

I'm sorry for bothering you.
I do not correctly understand about the spec of the APCu.
I didn't know that APCu can directly save an array...

katsew:
@narfbg 

Thank you for your comment!
I understand and fixed it.
Please review again.

system/libraries/Cache/drivers/Cache_apcu.php
katsew: The answer to your question is, I don't want to serialize/deserialize data if it is not necessary, because it could affect to the performance.

PR #5029, opened 2017-02-20
Update oci8_driver.php

yayorock:
values length and type are optional

### DIFF
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -373,7 +373,7 @@ class CI_DB_oci8_driver extends CI_DB {
 				}
 			}
 
-			oci_bind_by_name($this->stmt_id, $param['name'], $param['value'], ($param['length']=='')?NULL:$param['length'], ($param['type']=='')?NULL:$param['type']);
+			oci_bind_by_name($this->stmt_id, $param['name'], $param['value'], empty($param['length'])?-1:$param['length'], empty($param['type'])?SQLT_CHR:$param['type']);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


narfbg:
Please elaborate.

yayorock:
oci_bind_by_name use 3 requiered parameters, the others are optional. 
the array of parameteres use the variable type and length with oci_bind_by_name even if you didn't set those parameters.

narfbg:
That doesn't answer anything really ...

*Where* do these parameters come from?
What use case does this improve/fix?

Explain it from a user's perspective ...

zploskey:
@yayorock Do you encounter a bug when then function is called? Why does this need to be changed?

Documentation for `oci_bind_by_name` is here: http://php.net/manual/en/function.oci-bind-by-name.php
`bool oci_bind_by_name ( resource $statement , string $bv_name , mixed &$variable [, int $maxlength = -1 [, int $type = SQLT_CHR ]] )`

Parameters `length` and `type` are indeed both optional int parameters. It is unclear from the documentation how this function should behave if these params are empty strings as they currently are used. If '' is treated as null as in == comparison then this change has no effect. Without an Oracle license we can't test this. If you have one, maybe you can tell us how it behaves.

As a side note, the potential failure of `oci_bind_by_name` does not appear to be handled, but that is another issue.

yayorock:
When use the function "stored_procedure" , if set  2 required parameters and not 4, raise an error because it expected 4.
@zploskey you are right, the comparasion has not effect, the correct change(already tried) is
`oci_bind_by_name($this->stmt_id, $param['name'], $param['value'], empty($param['length'])?-1:$param['length'], empty($param['type'])?SQLT_CHR:$param['type']);` 

This, set the default value if 'length' or 'type' not set. With this is not need to put for each bind param, 4 parameters in array.

Issue #2721, opened 2013-11-09
Query Builder Cache with get_compiled_select(null, false) without reset


GDmac:
While testing pull request #2719 for Query Builder Cache, i used get_compiled_select to output and verify the generated queries, and found the following issue.

When using QB-cache together with `get_compiled_select('',false)` (i.e. DON'T reset select) results in duplicate selects and where clauses. e.g.

``` php
$this->db->start_cache();
$foo = $this->db
    ->select('id, title')
    ->from('pages')
    ->where('id', 1);
$this->db->stop_cache();

echo '<br>' . $foo->get_compiled_select(null,FALSE);
echo '<br>' . $foo->get_compiled_select(null,FALSE);

// result:
// SELECT `id`, `title` FROM `pages` WHERE `id` = 1
// SELECT `id`, `title`, `id`, `title` FROM `pages` WHERE `id` = 1 `id` = 1
```

PR #5033, opened 2017-02-22
DB_query_builder: Add cache merge flag.

tianhe1986:
I used the following code
```
$this->db->select('id, country');
$this->db->start_cache();
$this->db->where('id >', -9);		
$this->db->stop_cache();
$sql = $this->db->get_compiled_select('user', false);
echo $sql."\n";
$query = $this->db->get('');
```
the last `get()` generated
`SELECT id, country FROM user WHERE id > -9 id > -9`
and then caused a error.

When calling `get_compiled_select()`, each item in `$qb_XXX`(such as `$qb_where`, `$qb_groupby`) would be complied into a string, and they are not reset due to the second param of `get_compiled_select()` set to `TRUE`, then `$qb_cache_XXX` will be merged into `$qb_XXX` again while calling get() later.

So I add a $qb_cache_merged flag to fix it.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -187,6 +187,13 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	protected $qb_caching				= FALSE;
 
 	/**
+	 * QB Cache merge flag
+	 *
+	 * @var	bool
+	 */
+	protected $qb_cache_merged				= FALSE;
+
+	/**
 	 * QB Cache exists list
 	 *
 	 * @var	array
@@ -2735,6 +2742,12 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 */
 	protected function _merge_cache()
 	{
+		if ($this->qb_cache_merged === TRUE)
+		{
+			return;
+		}
+		$this->qb_cache_merged = TRUE;
+
 		if (count($this->qb_cache_exists) === 0)
 		{
 			return;
@@ -2853,7 +2866,8 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			'qb_no_escape'		=> array(),
 			'qb_distinct'		=> FALSE,
 			'qb_limit'		=> FALSE,
-			'qb_offset'		=> FALSE
+			'qb_offset'		=> FALSE,
+			'qb_cache_merged'	=> FALSE
 		));
 	}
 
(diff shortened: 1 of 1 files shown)


narfbg:
If the problem is only with `get_compiled_select()` - why not fix it there?

tianhe1986:
My bad, it also with other `get_compiled_XXX()` 
I should say it is with `_merge_cache()`

narfbg:
Still, my question was rather why do we need a flag for this?
It feels like a work-around - flagging a procedure as "potentially broken" instead of finding out where/why it breaks.

Honestly, I'd rather drop this entire "caching" feature than add more code for it.

tianhe1986:
Ah, 
when merging cache, it uses `in_array()` to check whether cache has been merged(see [here](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_query_builder.php#L2755))

However, after `_compile_XXX()`,  take `_compile_wh()` for example,  every item in `$qb_where` and `$qb_having` will be complied to string,  while item in `$qb_cache_where` and `$qb_cache_having`  remained array.

I get your point. So, maybe also compiling corresponding `$qb_cache_XXX` in `_compile_XXX()`?

narfbg:
I don't understand the question.

tianhe1986:
The question is,  cached QB arrays would be repeatedly merged, if QB arrays has been complied and not  reset.

Take the above code for example, I print the content of `$qb_where` and `$qb_cache_where` to explain the progress.

After
```
$this->db->select('id, country');
$this->db->start_cache();
$this->db->where('id >', -9);		
$this->db->stop_cache();
```
The contents are
```
$qb_where = array ( array ( "condition" => "id > -9", "escape" => TRUE ) );
$qb_cache_where = array ( array ( "condition" => "id > -9", "escape" => TRUE ) );
```

After 
`$sql = $this->db->get_compiled_select('user', false);`
The contents are 
```
$qb_where = array ( "`id` > -9");
$qb_cache_where = array ( array ( "condition" => "id > -9", "escape" => TRUE ) );
```

Then when running `$query = $this->db->get(''); `,  at the end of `_merge_cache()`, The contents are
```
$qb_where = array ( array ( "condition" => "id > -9", "escape" => TRUE ), "`id` > -9");
$qb_cache_where = array ( array ( "condition" => "id > -9", "escape" => TRUE ) );
```

So $qb_cache_where has been merged again and finally generated this illegal sql.
`SELECT id, country FROM user WHERE id > -9 id > -9`


And the same with other cached QB arrays.

@narfbg You are right. I am a little lazy so I chose  a work-around way :)

tianhe1986:
@narfbg I am stupid, forgot to search existed issues. This is duplicate of #2721

PR #5034, opened 2017-02-22
points dot

uguraktas:
points dot


narfbg:
points dot?

zploskey:
It's a python script. You don't need semicolons.

uguraktas:
I got it

PR #5049, opened 2017-03-10
Changed oracle where in split 1000

ytetsuro:
In order to correspond to ORACLE `ORC-01795`, when 1000 or more arrays are passed as the second argument of the `where_in` or `where_not_in` method, divide it into 1000 units.


narfbg:
Anything close to 1000 is an insane amount of values to be explicitly checking for. CI automatically hiding that problem for you is not a good idea.

ytetsuro:
Thank you for your answer.

Issue #5057, opened 2017-03-18
Some minor bugs in Postgres Driver (invalid reference to 'key' attribute)


tsergeant:
I was actually going to issue a pull request but am not quite sure how to make that happen. So far I've cloned the repository, made suggested bug fixes in a new branch and committed the branch (locally). I'm not sure what to do after that.

Anyway, the issue is with the postgres driver (system/database/drivers/postgre/postgre_driver.php) which errors if using a connection string (with extra options) to connect to the database. When cycling over the options there are several references to $this->key that should read $this->$key.

Also, a fairly common option (sslrootcert) is missing from the list.

zploskey:
To submit a pull request, press the "Fork" button in the upper right corner. You'll need to add your copy as a remote in your existing clone by doing something like `git remote add github https://github.com/tsergeant/CodeIgniter.git`. While on your branch with your patch, run `git push github`. This will push the branch to your fork on Github. There should then be a link to compare and submit a pull request on your fork's page.

I see what you're saying about the use of `$this->key` instead of `$this->$key` here:
https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/drivers/postgre/postgre_driver.php#L133

The comment on that block suggests that this code not very often used.

narfbg:
Pushed bugfix patch, linked above.

As for options such as 'sslrootcert' ... At least that one isn't described on the php.net manual, as far as I can see, so that's why it's missing - we just don't know about it.  
Specifically for SSL settings, we already have made some available for MySQL and I'd rather reuse the same keys, as they're described in the config file. Please open a new issue/PR about that and I'd be happy to work with you on that.

Also, we do have a contribution guide: https://github.com/bcit-ci/CodeIgniter/blob/develop/contributing.md

PR #5063, opened 2017-03-23
Added complete list of of Postgres libpq-connect options to Postgres Driver

tsergeant:
Thanks to all of you who work to keep CodeIgniter running!

In an earlier bug report/issue (https://github.com/bcit-ci/CodeIgniter/issues/5057) there was discussion about the list of options currently supported in the Postgres libpq-connect string. In my own app I'm using the 'sslrootcert' option. I went to the Postgres documentation to see what else is allowed: https://www.postgresql.org/docs/9.1/static/libpq-connect.html

I have updated the list to include all valid options. I'm thinking the purpose of the foreach loop is to prevent programmers/whoever from generating non-valid options to postgres. The way it stands now it prevents valid (and in my case necessary) options.

BTW, it is not a big deal for me to just patch the core with my own option list ... and as far as I can tell, I'm the only person on the planet who uses this section of code! :-) Part of me is not particularly thrilled that I will have to go through such a long list every time I connect to the db. Anyway, let me know your thoughts ... I'm fine with whatever gets decided.

P.S. I wasn't sure what to do about the key/$key issue since that was already accepted.

### DIFF
--- a/system/database/drivers/postgre/postgre_driver.php
+++ b/system/database/drivers/postgre/postgre_driver.php
@@ -128,11 +128,15 @@ class CI_DB_postgre_driver extends CI_DB {
 		 *
 		 * postgre://username:password@localhost:5432/database?connect_timeout=5&sslmode=1
 		 */
-		foreach (array('connect_timeout', 'options', 'sslmode', 'service') as $key)
+		foreach (array('connect_timeout', 'client_encoding', 'options',
+			'application_name', 'fallback_application_name', 'keepalives',
+			'keepalives_idle', 'keepalives_interval', 'keepalives_count', 'tty',
+			'sslmode', 'requiressl', 'sslcert', 'sslkey', 'sslrootcert', 'sslcrl',
+			'requirepeer', 'krbsrvname', 'gsslib', 'service') as $key)
 		{
-			if (isset($this->$key) && is_string($this->key) && $this->key !== '')
+			if (isset($this->$key) && is_string($this->$key) && $this->$key !== '')
 			{
-				$this->dsn .= $key."='".$this->key."' ";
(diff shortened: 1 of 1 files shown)


tsergeant:
I did some more reading (https://www.postgresql.org/docs/9.1/static/libpq-pgservice.html) and learned something new. Apparently the 'service' option allows you to specify multiple parameters via a file ... which would allow getting around the absence of sslrootcert without having to add it to the list of accepted options.

So maybe the original list is fine.

narfbg:
If we'd have to do this for 20 options, I'd rather just get rid of the whitelist ...

However, for the SSL options in particular, please read https://github.com/bcit-ci/CodeIgniter/issues/5057#issuecomment-287690638) more carefully.

tsergeant:
I think you are about to talk me into deciding to make no change (unless it is to remove the whitelist altogether). It does seem to be overkill to list all the options the way I listed it.

 I re-read https://github.com/bcit-ci/CodeIgniter/issues/5057#issuecomment-287690638 and took a look at config/database.php and the options listed in its documentation. One issue is that sslrootcert exists as an option for Postgres and doesn't exist for MySQL.  Limiting options to ones that both DBMS's share will necessarily restrict functionality in one or both of the others.

FWIW, I don't even use database.php in my application so removing the whitelist would allow me to put any valid option. Also, now that I learned about the "service" option I can probably get around the absence of sslrootcert even if the whitelist is in place.

So, I think I'm recanting this proposed pull request in favor of one of these:
* leave the whitelist as it is
or
* remove the whitelist altogether (if this is something you think we should pursue I don't mind generating another pull request with the proposed change)

Thanks for your consideration!

narfbg:
Removing the whitelist is a better option.

PR #5064, opened 2017-03-24
"Named" callables; i.e. array('name' => $callable)

b1tc0re:


narfbg:
Named callables what?

b1tc0re:
I do not speak English well. I will make an example for 10 min and show.

b1tc0re:
rules **required** should be checked first and then all the rest.
Send form but do not fill in the field. To check the rule **required** 
https://dev.dilexy.ru/

b1tc0re:
Maybe need change 
`return array_merge($callbacks, $new_rules);`
to
`return array_merge($new_rules, $callbacks);`

narfbg:
No, callbacks do come before required.

PR #5066, opened 2017-03-25
Fix directory delimiter when load local file.

vibbow:
We are facing some problem which our product are origional development on Windows, so we can mixed use backword slash and forward slash with no problem. (and this across all over our product).

But one of our developer are have problem to loading resource (view template) because he is using Mac and Mac OS didn't support mixed slash.

So I add those two little line to auto convert to the correct slash


narfbg:
We don't, and will not intentionally support mixed slashes either.

Sorry.

PR #5067, opened 2017-03-26
New Cache method: get_loaded_driver()

gxgpet:
New Cache.php method added, for finding out which caching driver was loaded; Updated docs as well.

### DIFF
--- a/user_guide_src/source/libraries/caching.rst
+++ b/user_guide_src/source/libraries/caching.rst
@@ -188,6 +188,16 @@ Class Reference
 		.. note:: The information returned and the structure of the data is dependent
 			on which adapter is being used.
 
+	.. php:method:: get_loaded_driver()
+
+		:returns:	Loaded driver name, based on initialization config ('apc', 'apcu', 'dummy', 'file', 'memcached', 'redis' or 'wincache')
+		:rtype:	string
+
+		This method will return the caching driver currently used after initialization.
+		::
+
+			echo $this->cache->get_loaded_driver(); // Will return something like "file"
+
 *******
 Drivers
 *******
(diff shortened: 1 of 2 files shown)


narfbg:
What for?

gxgpet:
For instance, if you have a panel board where you can see "health" stats about your server/app, such as what caching method is currently used - maybe my Memcached instance is down and files are used instead - in some advanced situations, this could lead to some email notifications/alerts as well.

I know that usage scenarios and situations are pretty limited for this, but it may be help some people out there.

If you consider not adding this, I'm fine with it. :)

narfbg:
Do *you* need it for those use cases?

If not, we've had this conversation already ... We don't need features that nobody asked for and nobody actually uses.

gxgpet:
Yes, I do use it for similar cases like the ones mentioned by me.

But that's my use case, so please keep in mind that if you are considering it wouldn't be used by others as well, I'm really fine to use an updated version of Cache.php myself, instead of integrating this directly into CI's codebase, though I see a [similar approach](https://www.codeigniter.com/user_guide/database/db_driver_reference.html#CI_DB_driver::platform) on Db area as well.

user_guide_src/source/libraries/caching.rst
```diff
@@ -188,6 +188,16 @@ Class Reference
 		.. note:: The information returned and the structure of the data is dependent
 			on which adapter is being used.
 
+	.. php:method:: get_loaded_driver()
+
+		:returns:	Loaded driver name, based on initialization config ('apc', 'apcu', 'dummy', 'file', 'memcached', 'redis' or 'wincache')
```
narfbg: "based on initialization config" implies that it just returns what was configured ... this isn't correct (and it would be useless if that was the case)

gxgpet: Indeed. I changed to a different wording there.

PR #5073, opened 2017-03-29
Removed whitelist from Postgres driver code to allow additional valid options

tsergeant:
This is a revised approach to an earlier pull request (https://github.com/bcit-ci/CodeIgniter/pull/5063). That discussion ended with the idea of removing a whitelist in system/database/drivers/postgre/postgre_driver.php. In removing the whitelist I encountered on reason why it may have been in place: DB.php added the extra properties to the object itself which has MANY properties. Without the whitelist there was not a convenient way to traverse the extra properties.

To address that I made a modification to DB.php to store the properties in an array so they would be isolated in a field called 'extra'. This allows the driver to traverse the resulting small array to isolate the extra properties.

This is the solution I have currently implemented in my own application and it appears to be okay for my use-case (and a few others I played with).

### DIFF
--- a/system/database/DB.php
+++ b/system/database/DB.php
@@ -136,10 +136,11 @@ function &DB($params = '', $query_builder_override = NULL)
 				if (is_string($val) && in_array(strtoupper($val), array('TRUE', 'FALSE', 'NULL')))
 				{
 					$val = var_export($val, TRUE);
+					$extra[$key] = $val;
 				}
-
-				$params[$key] = $val;
 			}
+
+			$params['extra'] = $extra;
 		}
 	}
 
--- a/system/database/DB.php
+++ b/system/database/DB.php
@@ -138,6 +138,8 @@ function &DB($params = '', $query_builder_override = NULL)
 					$val = var_export($val, TRUE);
 					$extra[$key] = $val;
 				}
+
+				$params[$key] = $val;
 			}
 
 			$params['extra'] = $extra;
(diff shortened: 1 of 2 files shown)


narfbg:
What are those DB.php changes for?
And what's with the `$this->extra`?
Also, not following the [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).

narfbg:
+ the Travis tests clearly show this breaks stuff.

tsergeant:
Let me try again! :-)

I saw a styling issue in the for-loop and have fixed it. If I'm out of bounds somewhere else I may need a hint.

I re-modified DB.php to address what I think caused all of the havoc in the Travis tests. (We'll see)

New-and-Improved Explanation:
In the original whitelist code there was:

foreach (array('connect_timeout', 'options', 'sslmode', 'service') as $key)

which checked to see if $this->$key existed and if so added it to the dsn string in the postgres driver. (NOTE: $this->$key was being created in DB.php by the statement: $params[$key] = $val;)

Now in the postgres driver there is not a great way to know which extra options have been specified because $this has dozens of other attributes (i.e., we don't want to cycle through them). So, in DB.php I put this statement: $params['extra'] = $extra;  which provides all of the extra parameters in its own array allowing us to cycle through only those parameters to build the dsn string.

tsergeant:
Perhaps a bit too optimistic on passing the Travis tests. I'll try a few more tweaks to address that.

tsergeant:
Well ... I'm not sure what to say about the connection error.

tsergeant:
I'm going to go ahead and close this. Not sure what went wrong and it is not urgent.

PR #5074, opened 2017-03-31
Update pdo_odbc_driver.php

ft23:
ODBC doesn't support LIMIT, it must be "TOP".

https://docs.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql#DML

Thats work for me.


narfbg:
ODBC isn't an alias for MSSQL.
ODBC supports whatever it connects to supports.
One thing that *our odbc driver* doesn't support anymore is the Query Builder, so the method you're modifying is never used either way.

PR #5075, opened 2017-04-03
Hotfix/parser bug fixing

agungjk:
Fixing this error

Message: Object of class stdClass could not be converted to string
Filename: libraries/Parser.php
Line Number: 150


narfbg:
It's not made to accept objects.

PR #5091, opened 2017-04-13
file_helper: get_extension_by_mime() added.

s2software:
Find the proper file extension from a given mime type.


narfbg:
The same MIME type id can, and often is used for multiple filename extensions - that alone makes this an unrealiable feature at best.

s2software:
Maybe for every mime we could return the "typical" 3 chars extension (es.: image/jpeg returns jpg, not jpeg) if any. Else the first found.
I used this feature where someone uploaded me a file in base64 (so I knew the mime type but not the extension) and had to save it to hard disk.

PR #5095, opened 2017-04-18
mysqli use options MYSQLI_OPT_INT_AND_FLOAT_NATIVE

shijinzhan:
Convert integer and float columns back to PHP numbers.


narfbg:
While I also wish this was the default to start with, it would make it inconsistent with the rest of our DB drivers, which don't have this option. And that means less portability.

PR #5098, opened 2017-04-21
Accept array in query builder -> order_by

spacepille:
Sometimes it is useful to write `order_by` columns as array. `select` and `group_by` already have the possibility. I can write code like this:

`->order_by(implode(',', [`
`'FIELD(ev.country_iso3,"CAN","USA") ASC',`
`'FIELD(ev.country_iso3,"SWE","GBP","FRA","DEU") DESC',`
`'ev.country_iso3 ASC',`
`'ev.date ASC',`
`'ev.name ASC',`
`]))`

But it is not as performant and it doesn't look so nice.
So hopefully you accept the pull request :) I would be glad to add the feature to CI4 too.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1188,9 +1188,9 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	/**
 	 * ORDER BY
 	 *
-	 * @param	string	$orderby
-	 * @param	string	$direction	ASC, DESC or RANDOM
-	 * @param	bool	$escape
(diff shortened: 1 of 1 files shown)


narfbg:
I don't mind the feature itself, but unlike `select()` and `group_by()`, the method signature of `order_by()` has an extra parameter and that would make this feature ambigous ...

spacepille:
I couldn't get what you mean. How does an extra parameter make the change ambigous?

narfbg:
It is unclear when and how the `$direction` param is used.

PR #5099, opened 2017-04-21
Fix several potential XSS injections in the form_helper

Calico90:
Fix several potential XSS injections in the form helper mainly consisting on:

Applies html_escape multiple times on HTML attribute (`foo="bar"`).
- HTML attribute values (`bar`)
- HTML attribute names (`foo`)

Applies xss_clean sometimes on use outside HTML attributes like `<button>XX</button>`.

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -88,7 +88,7 @@ if ( ! function_exists('form_open'))
 			$attributes .= ' accept-charset="'.strtolower(config_item('charset')).'"';
 		}
 
-		$form = '<form action="'.$action.'"'.$attributes.">\n";
+		$form = '<form action="'.html_escape($action).'"'.$attributes.">\n";
 
 		if (is_array($hidden))
 		{
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -202,7 +202,7 @@ if ( ! function_exists('form_hidden'))
 
 		if ( ! is_array($value))
 		{
-			$form .= '<input type="hidden" name="'.$name.'" value="'.html_escape($value)."\" />\n";
+			$form .= '<input type="hidden" name="'.html_escape($name).'" value="'.html_escape($value)."\" />\n";
 		}
 		else
 		{
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -583,7 +583,7 @@ if ( ! function_exists('form_button'))
 		}
 
 		return '<button '._parse_form_attributes($data, $defaults)._attributes_to_string($extra).'>'
-			.xss_clean($content)
+			.$content
 			."</button>\n";
 	}
 }
@@ -618,7 +618,7 @@ if ( ! function_exists('form_label'))
 			}
 		}
 
-		return $label.'>'.xss_clean($label_text).'</label>';
+		return $label.'>'.$label_text.'</label>';
 	}
 }
 
@@ -641,7 +641,7 @@ if ( ! function_exists('form_fieldset'))
 		$fieldset = '<fieldset'._attributes_to_string($attributes).">\n";
 		if ($legend_text !== '')
 		{
-			return $fieldset.'<legend>'.xss_clean($legend_text)."</legend>\n";
+			return $fieldset.'<legend>'.$legend_text."</legend>\n";
 		}
 
 		return $fieldset;
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -234,8 +234,8 @@ if ( ! function_exists('form_input'))
 	{
 		$defaults = array(
 			'type' => 'text',
-			'name' => is_array($data) ? '' : html_escape($data),
-			'value' => html_escape($value)
+			'name' => is_array($data) ? '' : $data,
+			'value' => $value
 		);
 
 		return '<input '._parse_form_attributes($data, $defaults)._attributes_to_string($extra)." />\n";
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -992,7 +992,7 @@ if ( ! function_exists('_attributes_to_string'))
 
 			foreach ($attributes as $key => $val)
 			{
-				$atts .= ' '.$key.'="'.$val.'"';
+				$atts .= ' '.$key.'="'.html_escape($val).'"';
 			}
 
 			return $atts;
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -208,8 +208,9 @@ if ( ! function_exists('form_hidden'))
 		{
 			foreach ($value as $k => $v)
 			{
-				$k = is_int($k) ? '' : $k;
-				form_hidden($name.'['.$k.']', html_escape($v), TRUE);
+				$k = is_int($k) ? '' : html_escape($k);
+				$n = html_escape($name).'['.$k.']';
+				form_hidden($n, html_escape($v), TRUE);
 			}
 		}
 
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -234,8 +234,8 @@ if ( ! function_exists('form_input'))
 	{
 		$defaults = array(
 			'type' => 'text',
-			'name' => is_array($data) ? '' : $data,
-			'value' => $value
+			'name' => is_array($data) ? '' : html_escape($data),
+			'value' => html_escape($value)
 		);
 
 		return '<input '._parse_form_attributes($data, $defaults)._attributes_to_string($extra)." />\n";
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -420,7 +420,7 @@ if ( ! function_exists('form_dropdown'))
 					continue;
 				}
 
-				$form .= '<optgroup label="'.$key."\">\n";
+				$form .= '<optgroup label="'.html_escape($key)."\">\n";
 
 				foreach ($val as $optgroup_key => $optgroup_val)
 				{
@@ -435,7 +435,7 @@ if ( ! function_exists('form_dropdown'))
 			{
 				$form .= '<option value="'.html_escape($key).'"'
 					.(in_array($key, $selected) ? ' selected="selected"' : '').'>'
-					.(string) $val."</option>\n";
+					.(string) html_escape($val)."</option>\n";
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


system/helpers/form_helper.php
```diff
@@ -88,13 +88,13 @@ function form_open($action = '', $attributes = array(), $hidden = array())
 			$attributes .= ' accept-charset="'.strtolower(config_item('charset')).'"';
 		}
 
-		$form = '<form action="'.$action.'"'.$attributes.">\n";
+		$form = '<form action="'.html_escape($action).'"'.$attributes.">\n";
```
narfbg: Most obvious BC break here is `&` being converted to `&amp;` ...

system/helpers/form_helper.php
```diff
@@ -88,13 +88,13 @@ function form_open($action = '', $attributes = array(), $hidden = array())
...
 
 		if (is_array($hidden))
 		{
 			foreach ($hidden as $name => $value)
 			{
-				$form .= '<input type="hidden" name="'.$name.'" value="'.html_escape($value).'" />'."\n";
+				$form .= '<input type="hidden" name="'.html_escape($name).'" value="'.html_escape($value).'" />'."\n";
```
narfbg: Re-iterating on our internal discussion - not only is this supposed to be hard-coded into applications, but there's no use case for it being user input ever.

system/helpers/form_helper.php
```diff
@@ -202,14 +202,15 @@ function form_hidden($name, $value = '', $recursing = FALSE)
 
 		if ( ! is_array($value))
 		{
-			$form .= '<input type="hidden" name="'.$name.'" value="'.html_escape($value)."\" />\n";
+			$form .= '<input type="hidden" name="'.html_escape($name).'" value="'.html_escape($value)."\" />\n";
```
narfbg: Same as line 97 (and wherever else it's about a "name" attribute)

system/helpers/form_helper.php
```diff
@@ -419,7 +420,7 @@ function form_dropdown($data = '', $options = array(), $selected = array(), $ext
 					continue;
 				}
 
-				$form .= '<optgroup label="'.$key."\">\n";
+				$form .= '<optgroup label="'.html_escape($key)."\">\n";
```
narfbg: Who *labels* their forms with user input?

system/helpers/form_helper.php
```diff
@@ -434,7 +435,7 @@ function form_dropdown($data = '', $options = array(), $selected = array(), $ext
 			{
 				$form .= '<option value="'.html_escape($key).'"'
 					.(in_array($key, $selected) ? ' selected="selected"' : '').'>'
-					.(string) $val."</option>\n";
+					.(string) html_escape($val)."</option>\n";
```
narfbg: So far the only place I see real potential, but it's also a potential BC break similar to the one in `form_open()`.

You know my opinion in general - why should we go out of our way to HTML-encode values that are *never* supposed to be user input? Context matters.

system/helpers/form_helper.php
```diff
@@ -606,14 +607,14 @@ function form_label($label_text = '', $id = '', $attributes = array())
 
 		if ($id !== '')
 		{
-			$label .= ' for="'.$id.'"';
+			$label .= ' for="'.html_escape($id).'"';
```
narfbg: Same as for the `label` attribute previously.

ivantcholakov:
html_escape() usage smells a little here. Just want to mention that escaping HTML and escaping HTML attributes are different things.

narfbg:
All involved here so far know this, `html_escape()` is the correct thing to use. Question is *what* needs escaping and what not.

system/helpers/form_helper.php
Calico90: For example: When forms are dynamically generated from a database (which is not unthinkable nowadays, especially in advanced web applications). It's common to store (key value pairs) in database tables. Because the forms are then dynamically generated. Unfortunately, I think we make a big thinking mistake. Because technically we're talking about an HTML attribute value and it shouldn't matter if that attribute is `name, value, class, type` or whatever. 

When you use the `form_helper` in the current scenario, it means that you're for 50% "protected" against XSS because our current decision is that we do only protect the attribute value for the HTML attribute `value` but not for the HTML attribute  `name`, since we distinguish it based on "context" and "possible use". I think that's wrong and especially inconsistent.

system/helpers/form_helper.php
Calico90: Same as https://github.com/bcit-ci/CodeIgniter/pull/5099#discussion_r115603067.

system/helpers/form_helper.php
Calico90: Same as my comment on line 97. Dynamically generated forms. Apart from that, consistency. Choosing to protect one element but not the other based on "possible use" and "context" is a poor decision. Better escape it everywhere or escape ie nowhere so it's at least consistent and clear to anyone what is or isn't done. But not "it's maybe safe based on...".

system/helpers/form_helper.php
Calico90: I'm really a fan of context myself as well. But a framework (like CI) should facilitate. One of the things it should facilitate along with functionality, performance, structure and so on is security. As said before and above 50% escaping and 50% non-escaping makes it still insecure and inconsistent. 

To use a metaphor, why to fix all the holes in a sinking boat but leave one hole. Then we could as well not fix the others too, with that one hole it's still sinking. (Only a bit slower). It's the same for this, we try to escape most of it, but not all. Making it (depending on the context it's used in, harder but not impossible to use XSS). Therefore, we, as a framework should be consistent and clear about our choices.

system/helpers/form_helper.php
Calico90: Same thing. Dynamic content/forms... Consistency... Security everything or nothing...

system/helpers/form_helper.php
Calico90: This is not a BC break at all since `&amp;` is a perfectly fine HTML encoded character that will be interpreted by the browser as `&`. You can use `/?key=value&amp;test=123` without any problems in any browser. Test it yourself:

    <a href="test.html?key=value&amp;test=123">test</a>

narfbg: Perhaps ... until you meet the 2 people who would put user input into an `action` attribute, and one of them actually knew what they were doing, so they already encoded it.

system/helpers/form_helper.php
narfbg: Context is everything in HTML.
"You're 50% protected" is a gross over-simplification of your opinion, based on the assumption that everything is user input.

^ You know this.

system/helpers/form_helper.php
narfbg: > One of the things it should facilitate along with functionality, performance, structure and so on is security.

Are you implying that we don't facilitate security? CI isn't a security-centric framework, but do you really want to equate the two things?
You know the only way everything is secure - if it doesn't exist.

> To use a metaphor, why to fix *almost* all the holes in a sinking boat but leave one hole.

This "hole" you're talking about here happens to be one of those round windows on the side of the ship. :)

> Do it 100% or not at all.

Oh, come on man ... like if we did infact drop all escaping, you wouldn't be the first to scream about it. This is a silly persuasion tactic.

system/helpers/form_helper.php
Calico90: If, if, if... Check the context then.

narfbg: Don't get what you mean.

Calico90: I'll give a proper answer tonight ;) but I like this discussion.

Calico90:
@narfbg Will we do something with this or not? I still think we should. Otherwise I'll close the request because it's useless then.

narfbg:
Dunno, I had to read through all of it to remember what it was about ...

To be honest, I'd be much more worried if *any* of these attributes is user controlled. XSS in them would be the least of anybody's problems at that point. Especially if we're talking about `action`, and to some extent `name`.
At the same time, any BC break in `action`, `name` handling will automatically result in broken applications, and our reasoning for this - something XSS something, just in case, for possibly 1%, but also possibly 0% of our users - just wouldn't be serious.

On the rest of them, I still disagree, but I guess at least there's less potential for breaking existing applications, so ... maybe. But you're certainly using the wrong arguments on that (as far as convincing me goes anyway); I see it as over-cautious hardening at best.

spaze:
I'd like to reopen this. The [form helper documentation](https://www.codeigniter.com/userguide3/helpers/form_helper.html) says:

*If you use any of the form helper functions listed on this page, the form values will be automatically escaped, so there is no need to call this function. Use it only if you are creating your own form elements.*

Which obviously is not correct, because the values are not escaped always, see for example `form_dropdown`:

https://github.com/bcit-ci/CodeIgniter/blob/b37902586b10bdbd7a34480950098800ddf4b120/system/helpers/form_helper.php#L437

Using user input as dropdown values is a perfectly valid use case, e.g. let a user create a list of categories, then offer this list as a dropdown when adding a product).

To me, there seems to be three options:

1. merged this (I'm aware of the BC break) to fix the vulnerability
2. update docs:
    - remove the note about auto escaping
    - add a big fat warning that users have to escape the values manually
    - remove the *Strong Security* claim from the homepage because it's broken
3. merge this PR partially (e.g. the dropdown value escaping), and remove the note (or at least update it to mention unescaped attributes) and the claim.

Which is the preferred way to handle this?

narfbg:
> *If you use any of the form helper functions listed on this page, the form values will be automatically escaped, so there is no need to call this function. Use it only if you are creating your own form elements.*
>
> Which obviously is not correct, because the values are not escaped always

On the contrary, it is very correct because it specifically talks about *form values*. That is, the `value` attribute and/or what you would receive as input values - always escaped. That being said, I would approve of the `<option>` contents being escaped as well (as I've already indicated).

spaze:
Thanks, but sorry, the message is not correct, it's unclear and misleading, otherwise there would be no need to escape `<option>` too. Anyway, let me get that right, I'll prepare a doc update.

@Calico90 you want to prepare a new PR for escaping `<option>`, at least?

narfbg:
Unclear and misleading? OK, I can see where that comes from and that's why I agree on escaping `<option>` contents (note that I say *contents*; the actual value is already escaped).

But the "need" to protect users from mistakes is one thing, and being incorrect is a different one; is not incorrect.

PR #5111, opened 2017-05-08
Assignment by reference for get_instance() call

vmoulin77:


narfbg:
We do have this all over the place, but as it doesn't make a difference either way - it's a change for the sake of change, and we don't do that, sorry.

PR #5112, opened 2017-05-09
read() in DB_cache fails when catching E_WARNING with custom error handler

cerealbeer:
Hi,

when using `set_error_handler()` and matching E_WARNING, the read() method in CI_DB_Cache fails at [L146](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_cache.php#L146), this because [`file_get_contents()`](http://php.net/file_get_contents) emits an E_WARNING when it does not find the file.

This should fix it:

````
if ( ! is_file($filepath) OR FALSE === ($cachedata = @file_get_contents($filepath)))
````


narfbg:
Please revert your second commit (it's completely unrelated and doesn't do anything either way) and remove the error suppression op from the `file_get_contents()` call - it doesn't make sense anymore with the `is_file()` check.

cerealbeer:
Done! I'm sorry, I did not intended to push it here.

Issue #3549, opened 2015-02-03
ini_set(): Use of iconv.internal_encoding is deprecated


ozziest:
system/core/CodeIgniter.php [LINE: 266] 
ini_set(): Use of iconv.internal_encoding is deprecated.

[http://php.net/manual/en/iconv.configuration.php](http://php.net/manual/en/iconv.configuration.php)

It's about 3.1 RC

narfbg:
Lines 264,265:

                // iconv.internal_encoding is deprecated starting with PHP 5.6
                // and it's usage triggers E_DEPRECATED messages.

dknell:
@narfbg In PHP 7, I think this might be a bigger deal. I was getting a 500 Internal Server Error in our app with no other message or errors in the log and I tracked it back to this. When I commented out `@ini_set('iconv.internal_encoding', $charset);`, I get a 200 response and everything loads fine. If I remove the silencer (`ini_set('iconv.internal_encoding', $charset);`), I get this:
```
A PHP Error was encountered
Severity: 8192
Message: ini_set(): Use of iconv.internal_encoding is deprecated
Filename: core/CodeIgniter.php
Line Number: 269
```
We're running the latest PHP7 (7.0.15) in Ubuntu 16.04. By default iconv is enabled and if I disable it then other things start breaking, like composer.

According to the [iconv configuration page](http://php.net/manual/en/iconv.configuration.php), the recommended way to set the charset (as of PHP 5.6 or higher) is `ini_set('default_charset', 'UTF-8')`, which is already done [here](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/CodeIgniter.php#L194).

What do you think about doing something like this:
```
	// There's an ICONV_IMPL constant, but the PHP manual says that using
	// iconv's predefined constants is "strongly discouraged".
	if (extension_loaded('iconv'))
	{
		define('ICONV_ENABLED', TRUE);
		// iconv.internal_encoding is deprecated starting with PHP 5.6
		// and it's usage triggers E_DEPRECATED messages.
                if (! is_php('5.6')) {
    		    @ini_set('iconv.internal_encoding', $charset);
                }
	}
	else
	{
		define('ICONV_ENABLED', FALSE);
	}
```
I'd be happy to submit a PR if you think this is OK.

narfbg:
[...]

PR #5115, opened 2017-05-11
Fixed deprecation errors for php >=5.6

SCIF:
Guys, it's very bad idea to use error shut up operator if we can do that properly.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -196,9 +196,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 	if (extension_loaded('mbstring'))
 	{
 		define('MB_ENABLED', TRUE);
-		// mbstring.internal_encoding is deprecated starting with PHP 5.6
-		// and it's usage triggers E_DEPRECATED messages.
-		@ini_set('mbstring.internal_encoding', $charset);
+
 		// This is required for mb_convert_encoding() to strip invalid characters.
 		// That's utilized by CI_Utf8, but it's also done for consistency with iconv.
 		mb_substitute_character('none');
@@ -210,22 +208,24 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 
 	// There's an ICONV_IMPL constant, but the PHP manual says that using
 	// iconv's predefined constants is "strongly discouraged".
-	if (extension_loaded('iconv'))
-	{
-		define('ICONV_ENABLED', TRUE);
-		// iconv.internal_encoding is deprecated starting with PHP 5.6
-		// and it's usage triggers E_DEPRECATED messages.
-		@ini_set('iconv.internal_encoding', $charset);
-	}
-	else
-	{
-		define('ICONV_ENABLED', FALSE);
-	}
+    define('ICONV_ENABLED', extension_loaded('iconv'));
 
-	if (is_php('5.6'))
-	{
-		ini_set('php.internal_encoding', $charset);
-	}
+    if (is_php('5.6'))
+    {
+        ini_set('php.internal_encoding', $charset);
+    }
+    else
+    {
+        if (ICONV_ENABLED)
+        {
+            ini_set('iconv.internal_encoding', $charset);
+        }
+
+        if (MB_ENABLED)
+        {
+            ini_set('mbstring.internal_encoding', $charset);
+        }
+    }
 
 /*
  * ------------------------------------------------------


narfbg:
https://github.com/bcit-ci/CodeIgniter/issues/3549#issuecomment-297683685

SCIF:
@narfbg , [your comment](https://github.com/bcit-ci/CodeIgniter/issues/3549#issuecomment-297683685) does not cover `internal_encoding` directive which has more priority rather than `default_charset`. Check my code please — it sets correct directives and does not trigger ANY errors. Behavior is the same as expected.

narfbg:
I did check your code and it is incorrect in at least two different ways.

There's no `php.internal_encoding` directive. It is `iconv.internal_encoding` and `mbstring.internal_encoding` that have higher priority over `default_charset`.

SCIF:
@narfbg Heh. I doubt that you understand what we talking about.
1) http://php.net/manual/en/ini.core.php#ini.internal-encoding
2) It does not makes sense to set `iconv.internal_encoding` because it falls back to `default_charset`

SCIF:
@narfbg , hm, it looks like your code had typo and I copied it. Anyway, reason 2 is still actual and makes sense. According to docs, it makes sense just leave setting `default_charset` for php>=5.6 because `default_charset` and `*.internal_encoding` has the same value.

narfbg:
Yea, sure ... First I didn't check your code, and now I don't even understand what we're talking about.  

1. You're linking to an obscure description of `internal_encoding` (which says nothing about precedence), yet your code sets `php.internal_encoding`. Note the `php.` prefix; no such directive exists. No, my code didn't have a typo; **you** put this prefix in there.
2. The point is to *override* existing INI directives and NOT to rely on fallbacks. [My comment on #3549](https://github.com/bcit-ci/CodeIgniter/issues/3549#issuecomment-297683685) explains this. Who didn't read or understand something here?

`internal_encoding` does have a higher precedence over `default_charset` - I'll give you that. However, `iconv.internal_encoding` and `mbstring.internal_encoding` *still* have higher precedence even over `internal_encoding`.  
The entire precedency "chain" is shown here: https://github.com/php/php-src/blob/PHP-7.1.5/php.ini-production#L968

Did you seriously think we're happy with the error suppression? No, we *have* to use it in order to fix broken configurations.

P.S.: The docs on this are bad, indeed.

SCIF:
1. Are you [kidding](https://github.com/bcit-ci/CodeIgniter/pull/5115/files#diff-2b5d1b5e02bd017451cd950f04723379L227)??
2. Yes, I read your comment and understand it fully, but that can happen only in one case — admin/developer overrides default value of `*.internal_encoding`. And that is *deliberate choice* of user then. So would you like force setting other value and break user choice?

narfbg:
1. Fair enough, my bad. If you want to fix *that*, I'll be happy to merge your PR.
2. A lot to say about this ...

   - You're *assuming* it's a deliberate choice; this is far from guaranteed.
     - Left-over settings after upgrades.
     - Limited access environments are not rare, to say the least, and it is very common that devs don't have access to php.ini.
   - Even if intentional, php.ini directives apply to *PHP in general*; it is far from certain that the goal is to override CI in particular.
   - Deliberate choice doesn't equal correct choice.
   - Our users actually expect `config.php` to be all they need to configure.
   - And yes, I do like enforcement. :)

SCIF:
@narfbg ,
1. Yeah, I will fix it.
2. It's pretty bad approach — user supposed to be an idiot and you know better what he want :(
Ok, in this case I suggest next logic — check if some of `*.internal_encoding` directives setted and if so, force it with `ini_set()`. If null — just set `internal_encoding` and that will allow system to bootstrap without  suppresing  if everything configured correct.

Are you happy with that?

narfbg:
What would be the point of that? More code and overhead for zero benefit.

SCIF:
@narfbg , zero suppressed errors will be triggered if php server configured properly.

narfbg:
Sure, but how does that benefit anybody?

Error suppression is bad because it hides errors that you *should* care about. This isn't the case here.

SCIF:
I care about deprecation errors and my application does not suppres any errors in dev mode.

narfbg:
Do you *receive* the deprecation errors? Unless so, you care about a problem that doesn't exist.

SCIF:
Yes, I receive.

narfbg:
Where? How?

SCIF:
My debug panel show me all errors, including suppressed. Also, your current approach is bad because of next case: user has a correct setup and didn't use `*.internal_encoding` but he require to set custom encoding in his code. He will be forced to suppress errors and set deprecated values in his code, because CI setted it.

narfbg:
> My debug panel shows me all errors, including suppressed.

A debugger could even show you the exact memory blocks allocated by PHP while running your app. Using this as an argument isn't any better than "but I know the code is there".  

A valid issue would be if it appears in any output produced by the application itself, or in the error logs.

> Also, your current approach is bad because of next case: user has a correct setup and didn't use `*.internal_encoding` but he require to set custom encoding in his code. He will be forced to suppress errors and set deprecated values in his code, because CI setted it.

1. This is a bullshit scenario.
2. No, the user wouldn't have to change the INI settings at runtime. All mbstring and iconv functions accept charset as a parameter.
3. You're way too eager to make up excuses for telling us that we're "bad" and be snarky in general. This won't help you convince me in *anything*.

SCIF:
Yeah, only [freaks care](http://symfony.com/blog/new-in-symfony-3-3-persisted-deprecation-logs) about deprecations. You are on right track, go ahead. I will not spend my, and your, time on debugging and improving CI anymore.

PR #5117, opened 2017-05-11
Set checkbox default array

ytetsuro:
Considering the use case of set_checkbox, I think it is useful to be able to specify an array by default.
Checkbox is often used when Requesting an array.

For example, if you have an array with values that exist in the database in the edit screen, the following format is valid.

```
<?php echo set_checkbox('select[]', 'key1', $row['selected_list']);?>
```

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -768,7 +768,7 @@ if ( ! function_exists('set_checkbox'))
 	 * @param	string
 	 * @param	string
 	 * @param	bool
-	 * @return	string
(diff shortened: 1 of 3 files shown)


narfbg:
This is way too easy to do already:

    set_checkbox('select[]', 'key1', in_array('key1', $row['selected_list']))

ytetsuro:
Thank you for your comment.
I wrote too much of that code so PullRequest.
If implemented in FW, you do not need to write it a lot.

PR #5120, opened 2017-05-12
pdo odbc insert_id didnt work.

ft23:
this function return true result.


narfbg:
`@@IDENTITY` doesn't work on all RDBMS, nor are its results always integers.

... it's not a coincidence that ODBC doesn't support this - if it works for you, it won't work for somebody else.

ft23:
you right. do you have any idea else?
may be easy way http://stackoverflow.com/a/30381436

narfbg:
No, there's no way we can implement it in a way that works for everybody.
You'll have to do it in your own code.

PR #5121, opened 2017-05-13
A session reads a different file when a bug

yuchonghua:
If the system reads two different sessionIds will cause an exception

### DIFF
--- a/system/libraries/Session/drivers/Session_files_driver.php
+++ b/system/libraries/Session/drivers/Session_files_driver.php
@@ -201,7 +201,11 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 		}
 		else
 		{
-			rewind($this->_file_handle);
+			if ($this->_session_id === $session_id) {
+				rewind($this->_file_handle);
+			}else{
+				return $this->close() === $this->_failure ? $this->_failure : $this->read($session_id);
+			}
 		}
 
 		$session_data = '';


narfbg:
Please explain the problem.

yuchonghua:
When the visit is very large, sometimes session in the production session will read a production of $ session_id to determine whether there is, just this $ session_id is there, so the session to read the time, or Put back the original data, and then the program died on the cycle

narfbg:
This doesn't make any sense ...

yuchonghua:
Why is it meaningless? Do not understand

narfbg:
It's just incoherent, bad English ...

I can't blame you for not knowing a foreign language, but there's nothing I can do with that either.

yuchonghua:
In fact, this is a logical problem, you first look at the entire code, this change is necessary, if you think this change is not necessary can not change, but it is sometimes stuck with some user requests, because the session The reason for the session

tianhe1986:
@yuchonghua 
What will cause the two IDs not matching? And, how could it be reproduced?
I kind of prefer you to use our native language,  because I can't  understand what you are talking about, even as a Chinese.

yuchonghua:
其实生成会话的具体机制我不是很了解，但是我发现有些时候，在会话文件特别多的时候，会出现生成会话的阶段就卡死了，原因估计是session在生成新的session_id的时候是否会先查下一下session_id是否已经存在过，如果存在过，如果没有调用关闭，就继续读取一个新的session_id会导致一个问题就是，读取第二个session_id会拿到第一次session_id的内容，理解吗？因为文件指针被缓存了，但是需要读取另外一个文件

jim-parry:
@yuchonghua English only, please.

tianhe1986:
@jim-parry Yes, I'm just kidding :-)

tianhe1986:
@yuchonghua 
You should explain how to reproduce it.
When will `read()` be called with two IDs not matching? You could add some debug codes and find it out.

PR #5123, opened 2017-05-18
Update input.rst

milihuo:
when i use set_cookie() or delete_cookie() to delete cookie (this cookie set domain),I find cannot delete this cookie, so i seach google , if you wang to delete cookie must set the domain. Otherwise, it will not be deleted.


narfbg:
... unless there was no Domain attribute when the cookie was set.

This is how cookies work; it is not CodeIgniter's job to explain [RFC 6265](https://tools.ietf.org/html/rfc6265), sorry.

PR #5124, opened 2017-05-19
Zip library improved to have ability of storing timestamp in name if …

Tpojka:
Usually when you are working with large amount of same/similar files you could intend to organize those files setting name related to time when file was created. 
Set second parameter to true and you will get `$filename_$timestamp` named file. 
If you set third parameter to true timestamp part will come first in name `$timestamp_$filename` in case time is more important for your backups when you are seeking for some file in list.


narfbg:
No. I'd rather remove this method than extend its functionality ... because SRP.

PR #5128, opened 2017-05-21
Profiler add pre tag

lloricode:
.. in config and session when array


narfbg:
You're using spaces for indentation; we use tabs (read our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html)).

Also, the PR title is *very* ambiguous and you've provided no description at all ... Accurate descriptions are very important if you want to achieve anything.

PR #5130, opened 2017-05-24
fix bad attribute handling of form_label

aroche:
When passing a string for extra attributes in form_label helper function, these attributes were skipped.

### DIFF
--- a/system/helpers/form_helper.php
+++ b/system/helpers/form_helper.php
@@ -609,13 +609,7 @@ if ( ! function_exists('form_label'))
 			$label .= ' for="'.$id.'"';
 		}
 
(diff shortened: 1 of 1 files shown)


narfbg:
I'll merge this for "consistency" with the other form helpers, as silly decisions in the past have allowed this feature to creep in.

But, for the record:

- There's nothing to "fix" as there's nothing broken here in the first place. The function is *designed* to only accept arrays. This is very obvious if you look at the manual.
- Saying that this is "bad" is objectively wrong. Strings are absolutely arbitrary, and mean that you are responsible to generate half the output - beats the entire purpose of using `form_label()` to do the work. Arrays, on the other hand, are easy to handle/validate in this case.

PR #5139, opened 2017-06-01
Update News section

AbhaysinghBhosale:
I have updated an single line instruction .all other code block's in instruction's have mentioned that where to add them but this block does not have such instruction,there is just its description is there .So i have added one line to indicate that where to add it.


narfbg:
It's not like there's any reason to think the code goes elsewhere ...

PR #5143, opened 2017-06-05
Fix misc inconsistencies between code and doc comments

TysonAndre:
Various issues were detected by a static analyzer

### DIFF
--- a/system/core/Exceptions.php
+++ b/system/core/Exceptions.php
@@ -239,7 +239,7 @@ class CI_Exceptions {
 	 * @param	string	$message	Error message
 	 * @param	string	$filepath	File path
 	 * @param	int	$line		Line number
-	 * @return	string	Error page output
+	 * @return	void
 	 */
 	public function show_php_error($severity, $message, $filepath, $line)
 	{
--- a/system/core/compat/mbstring.php
+++ b/system/core/compat/mbstring.php
@@ -68,7 +68,7 @@ if ( ! function_exists('mb_strlen'))
 	 * @link	http://php.net/mb_strlen
 	 * @param	string	$str
 	 * @param	string	$encoding
-	 * @return	string
+	 * @return	int
 	 */
 	function mb_strlen($str, $encoding = NULL)
 	{
--- a/system/database/DB_forge.php
+++ b/system/database/DB_forge.php
@@ -488,7 +488,7 @@ abstract class CI_DB_forge {
 	 *
 	 * @param	string	$table		Table name
 	 * @param	bool	$if_exists	Whether to add an IF EXISTS condition
-	 * @return	string
+	 * @return	mixed	(Returns a platform-specific DROP table string, or TRUE to indicate there's nothing to do)
 	 */
 	protected function _drop_table($table, $if_exists)
 	{
@@ -991,8 +991,8 @@ abstract class CI_DB_forge {
 	/**
 	 * Process indexes
 	 *
-	 * @param	string	$table
-	 * @return	string
+	 * @param	string	$table	Table name
+	 * @return	string[] list of SQL statements
 	 */
 	protected function _process_indexes($table)
 	{
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -134,7 +134,7 @@ if ( ! function_exists('load_class'))
 	 *
 	 * @param	string	the class name being requested
 	 * @param	string	the directory where the class should be found
-	 * @param	string	an optional argument to pass to the class constructor
+	 * @param	mixed	an optional argument to pass to the class constructor
 	 * @return	object
 	 */
 	function &load_class($class, $directory = 'libraries', $param = NULL)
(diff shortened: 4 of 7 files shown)


system/core/Controller.php
```diff
@@ -48,6 +48,19 @@
  * @category	Libraries
  * @author		EllisLab Dev Team
  * @link		https://codeigniter.com/user_guide/general/controllers.html
+ *
+ * Note: the below aren't magic properties, they are dynamic properties (Set in the constructor, based on calls to load_class)
+ *
+ * @property CI_Benchmark	$benchmark (core)
```
narfbg: We don't use `@property`.

system/core/Exceptions.php
```diff
@@ -239,7 +239,7 @@ public function show_exception($exception)
 	 * @param	string	$message	Error message
 	 * @param	string	$filepath	File path
 	 * @param	int	$line		Line number
-	 * @return	string	Error page output
+	 * @return	void	Echoes the error page output to stdout
```
narfbg: `void` is correct, indeed.
But the description has nothing to do with the return type.

system/core/compat/mbstring.php
```diff
@@ -97,7 +97,7 @@ function mb_strlen($str, $encoding = NULL)
 	 * @param	string	$needle
 	 * @param	int	$offset
 	 * @param	string	$encoding
-	 * @return	mixed
+	 * @return	string|false
```
narfbg: We don't use `this|that` ... even if we did, `bool` would be the correct thing.
Either way, `false` is a rare edge case that we don't care for.

narfbg: On another note ... `string` is also incorrect. It returns an integer.
Better leave it `mixed`.

system/database/DB_forge.php
```diff
@@ -488,7 +488,7 @@ public function drop_table($table_name, $if_exists = FALSE)
 	 *
 	 * @param	string	$table		Table name
 	 * @param	bool	$if_exists	Whether to add an IF EXISTS condition
-	 * @return	string
+	 * @return	string|TRUE (Returns TRUE to indicate there's nothing to do)
```
narfbg: Same as `mb_strpos()` (on the `|true` thing).

system/database/DB_forge.php
```diff
@@ -992,7 +992,7 @@ protected function _process_primary_keys($table)
 	 * Process indexes
 	 *
 	 * @param	string	$table
-	 * @return	string
+	 * @return	string[] list of SQL statements
```
narfbg: ... it's an internal function. Why do you care?

Also, if you're fixing "inconsistencies", why does the return has a description, but the parameter not?

narfbg:
What is "the master branch of phan" and why should we care for it?

system/core/Controller.php
TysonAndre: removed.

system/core/Exceptions.php
TysonAndre: Moved to the top of the phpdoc description.

system/core/compat/mbstring.php
TysonAndre: Right, should be int, not string. Changed back to mixed

system/database/DB_forge.php
TysonAndre: Changed to mixed

system/database/DB_forge.php
TysonAndre: I meant type inconsistencies.

TysonAndre:
> What is "the master branch of phan" and why should we care for it?

Removed mention of phan from the commit description.

https://github.com/etsy/phan/ is a static analyzer for PHP that prefers to minimize false-positives.

system/core/Exceptions.php
```diff
@@ -234,12 +234,13 @@ public function show_exception($exception)
 
 	/**
 	 * Native PHP error handler
+     * (Echoes the error page output to stdout)
```
narfbg: Uses spaces for indentation (we use tabs) ... diff itself makes it look obvious. :)

Does this really need to be said in a docblock? The function name itself describes it pretty well.

system/core/Security.php
```diff
@@ -594,7 +594,7 @@ public function xss_hash()
 	 * Get random bytes
 	 *
 	 * @param	int	$length	Output length
-	 * @return	string
+	 * @return	string|false
```
narfbg: `false` is an edge case that could never happen in a sane environment. We don't care for it.

TysonAndre: Made the requested changes to Security and Exceptions

system/core/Exceptions.php
TysonAndre: Removed that line and the spaces

narfbg:
>> What is "the master branch of phan" and why should we care for it?
>
> Removed mention of phan from the commit description.
>
> https://github.com/etsy/phan/ is a static analyzer for PHP that prefers to minimize false-positives.

I see ...

Well, good if it helps you, but I absolutely despise the idea of some third-party tool dictating changes to our code. Please only change things that matter to you.

PR #5150, opened 2017-06-13
fix ORA- 00946

ytetsuro:
I got a `ORA- 00946 : missing TO keyword` error when running the following code.

```php
$this->dbforge->modify_column('foo', [
    'bar' => ['name' => 'foobar']
]);
```

Therefore, the `TO` keyword was added.


ytetsuro:
Thanks Merged!

PR #5153, opened 2017-06-14
Add responsive caching feature

MortadhaLaouini:
This feature enable developers to seperate storing cache files into two folders if they use different templates for desktop web pages and mobile web pages.


narfbg:
Thanks, but no ...

Caching isn't an easy problem to solve, it can't magically work just with the flip of a switch. Desktop vs. mobile may work for you, but someone else will want Firefox vs. Chrome, or geolocation, or another arbitrary criteria - we'll never see the end of it.

PR #5155, opened 2017-06-15
Ignore LIMIT when using CI_DB_query_builder::count_all_results

tianhe1986:
I used the following code for pagination search
```
                 $this->db->where('num >=', 9);
		$this->db->limit(2,2);
		$num = $this->db->count_all_results('test_user', false);
		$query = $this->db->get();
		$result = $query->result_array();
```
and found `count_all_results` generated
```
SELECT COUNT(*) AS `numrows`
FROM `test_user`
WHERE `num` >= 9
 LIMIT 2, 2
```

but 
```
SELECT COUNT(*) AS `numrows`
FROM `test_user`
WHERE `num` >= 9
```
was expected.

Surely I could use `limit(X,X)` after calling `count_all_results`, however, it would be better if `LIMIT` usage be ignored with `count_all_results`.

I also thought about generating sql like this
```
SELECT COUNT(*) AS `numrows`
FROM (
SELECT *
FROM `test_user`
WHERE `num` >= 9
 LIMIT 2, 2
) CI_count_all_results 
```
But it may be a rare usage.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1488,15 +1488,6 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			$this->qb_orderby = NULL;
 		}
 
-		//ignore LIMIT
-		if ($this->qb_limit OR $this->qb_offset)
-		{
-			$limit = $this->qb_limit;
-			$this->qb_limit = NULL;
-			$offset = $this->qb_offset;
-			$this->qb_offset = NULL;
-		}
-
 		$result = ($this->qb_distinct === TRUE OR ! empty($this->qb_groupby) OR ! empty($this->qb_cache_groupby))
 			? $this->query($this->_count_string.$this->protect_identifiers('numrows')."\nFROM (\n".$this->_compile_select()."\n) CI_count_all_results")
 			: $this->query($this->_compile_select($this->_count_string.$this->protect_identifiers('numrows')));
@@ -1505,19 +1496,10 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		{
 			$this->_reset_select();
 		}
-		// If we've previously reset the qb_orderby,qb_limit or qb_offset values, get them back
-		else
+		// If we've previously reset the qb_orderby values, get them back
+		elseif ( ! isset($this->qb_orderby))
 		{
-			if ( ! isset($this->qb_orderby))
-			{
-				$this->qb_orderby = $orderby;
-			}
-
-			if ( ! isset($this->qb_limit))
-			{
-				$this->qb_limit = $limit;
-				$this->qb_offset = $offset;
-			}
+			$this->qb_orderby = $orderby;
 		}
 
 		if ($result->num_rows() === 0)
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1488,7 +1488,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 			$this->qb_orderby = NULL;
 		}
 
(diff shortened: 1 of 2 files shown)


narfbg:
No ... it might make some sense for pagination, but in literally every other case this would be completely irrational behavior.

tianhe1986:
@narfbg  But it also generates a irrational sql when using `limit()` and `count_all_results()` at the same time. Any suggestion?

narfbg:
You use `limit()` and you get a `LIMIT` clause ... there's nothing irrational about that.

tianhe1986:
@narfbg No, `count_all_results()` is supposed to get the total number of the result rows.
For example, 
```
$this->db->group_by('num');
echo $this->db->count_all_results('test_user', false);
```

would generate
```
SELECT COUNT(*) AS `numrows`
FROM (
SELECT *
FROM `test_user`
GROUP BY `num`
) CI_count_all_results
```

In other words, 
```
SELECT COUNT(*) AS `numrows`
FROM (
SELECT *
FROM `test_user`
 LIMIT 2, 10
) CI_count_all_results
```
may be right. 
 Could I change it like this?

narfbg:
I don't understand what you're saying.

tianhe1986:
@narfbg Sorry. 
If I use
```
$this->db->limit(10,2);
$num = $this->db->count_all_results('test_user', false);
```

then sql generated should be
```
SELECT COUNT(*) AS `numrows`
FROM (
SELECT *
FROM `test_user`
 LIMIT 2, 10
) CI_count_all_results
```
rather than
```
SELECT COUNT(*) AS `numrows`
FROM `test_user`
 LIMIT 2, 10 
```

narfbg:
Yes ... how is that a problem?

tianhe1986:
@narfbg 
This is what I try to fix, but choose a wrong way. I choose to ignore `LIMIT`.
Could I change this PR and fix it?

narfbg:
**Why does it need fixing?**

tianhe1986:
@narfbg  You just agreed that 
```
SELECT COUNT(*) AS `numrows`
FROM (
SELECT *
FROM `test_user`
 LIMIT 2, 10
) CI_count_all_results
```
should be the right sql, didn't you?

narfbg:
No ... I misread what you were saying - sorry about that.

Either way, I still don't see what the problem is.

tianhe1986:
So, you think that
```
SELECT COUNT(*) AS `numrows`
FROM `test_user`
 LIMIT 2, 10 
```
is right.

In my opinion, when using `count_all_results()`, `LIMIT` usage should be handled the same way as `GROUP BY` usage, to get the 'right' number of rows. 

That is,
```
$this->db->limit(10,2);
$query = $this->db->get('test_user');
$result = count($query->result_array());
```
and 
```
$this->db->limit(10,2);
$result = $this->db->count_all_results('test_user', false);
```
should get the same result.

narfbg:
Dude ... stop showing me arbitrary SQL and telling me what I think about it. I'm telling you I don't see the problem *with either query* and I'm asking you for the Nth time **what problem are you trying to solve?**

I don't even know what change are you proposing anymore ... I don't care if `LIMIT` and `GROUP BY` are handled differently as long as the results are consistent.

What is the current behavior? And what is wrong with it?

tianhe1986:
@narfbg  The current behavior and the wrong is , it **get different number of result rows** compare to the query executed.

narfbg:
In that case - sure, it needs fixing.

tianhe1986:
@narfbg Sorry for  just taking codes for example. 
I have resubmitted the branch .

narfbg:
I don't see any new PRs, nor commits to this one?

tianhe1986:
Yes, I commit it to this one. Maybe this PR is closed so no longer updating commits?

narfbg:
Perhaps ... try again.

PR #5163, opened 2017-06-26
Fix load config order for packages

oostation:
As libraries, models and helpers from packages, config files needs to be loaded first from package path and could be overloaded from application path and/or environnment specific path.

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -783,7 +783,7 @@ class CI_Loader {
 
 		// Add config file path
 		$config =& $this->_ci_get_component('config');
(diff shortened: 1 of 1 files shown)


narfbg:
I agree in principle, but ...

- `add_package_path()` is very obviously not the only thing that would need changing. The patch is incomplete.
- This behavior was introduced 5+ years ago via #1262 (as a hack solution indeed), so it's a BC break and would have to wait for 3.2.0.

oostation:
OK
...so I will use `MY_Loader` instead

narfbg:
I see you're a big OSS fan ...

PR #5168, opened 2017-06-30
Fix count_all_results on SQLSRV with query cache

adam-bell:
Fix for count_all_results on MSSQL when using start_cache
- usage scenario, pagination

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1482,14 +1482,14 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		// ORDER BY usage is often problematic here (most notably
 		// on Microsoft SQL Server) and ultimately unnecessary
 		// for selecting COUNT(*) ...
-		if ( ! empty($this->qb_orderby))
-		{
-			$orderby = $this->qb_orderby;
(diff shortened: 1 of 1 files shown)


narfbg:
Explain what's broken please.

adam-bell:
Using MSSQL / SQLSRV Driver with SQL Azure

Previously:
When using start_cache() with a query having an order_by() clause, and then performing count_all_results() this resulted in an SQL error (MSSQL specific, due to the order by in a count(*))

This is worked around for regular queries with the qb_orderby being overridden in count_all_results but the compile select uses the qb_cache_orderby when cache is enabled

After the change:
qb_cache_orderby is emptied while the count is performed, then reinstated.

narfbg:
Hope you don't mind me hacking the fix myself ... I wanted to simplify the code.

Also, your patch isn't following our [styleguide](https://www.codeigniter.com/userguide3/general/styleguide.html) - please read it if you ever want to contribute again.

narfbg:
Well, as evident by #5282, both of our patches are erroneous.

While it is my fault that I didn't test mine, yours should've never been submitted if you haven't solved your own problem with it ... please don't do that again.

PR #5170, opened 2017-07-05
Decreasing the probability of race condition in session lock

tianhe1986:
When using `redis` or `memcached` session driver, if a web-client perform multiple requests at the same time, two or more requests may get the session lock, and finally some of them would generate "Session: Error while trying to free lock for" error log.

Using `Redis::setnx` and `Memcached::add ` properly could help to decrease the probability of this race condition.

### DIFF
--- a/system/libraries/Session/drivers/Session_memcached_driver.php
+++ b/system/libraries/Session/drivers/Session_memcached_driver.php
@@ -310,7 +310,7 @@ class CI_Session_memcached_driver extends CI_Session_driver implements SessionHa
 			if ( ! $this->_memcached->replace($this->_lock_key, time(), 300))
 			{
 				return ($this->_memcached->getResultCode() === Memcached::RES_NOTFOUND)
-					? $this->_memcached->set($this->_lock_key, time(), 300)
+					? $this->_memcached->add($this->_lock_key, time(), 300)
 					: FALSE;
 			}
 		}
@@ -326,7 +326,11 @@ class CI_Session_memcached_driver extends CI_Session_driver implements SessionHa
 				continue;
 			}
 
-			if ( ! $this->_memcached->set($lock_key, time(), 300))
+			$set_result = ($this->_memcached->getResultCode() === Memcached::RES_NOTFOUND)
+				? $this->_memcached->add($lock_key, time(), 300)
+				: $this->_memcached->set($lock_key, time(), 300);
+
+			if ( ! $set_result)
 			{
 				log_message('error', 'Session: Error while trying to obtain lock for '.$this->_key_prefix.$session_id);
 				return FALSE;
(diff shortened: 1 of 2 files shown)


narfbg:
It is unclear how or why this patch improves anything ... You'll need to elaborate *a lot*. :)

tianhe1986:
Take `redis` driver for example.

When two requests with same cookie arrive at the same time, the two processes may also execute the code to [here](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Session/drivers/Session_redis_driver.php#L362)  simultaneously, and get $ttl = -2.
Then they continue to execute `$this->_redis->setex($lock_key, 300, time())` and would both succeed, resulting in they all get lock.

However,  `setnx()`  or  `set($key, $value, array('nx'))` could be used as a lock. In this situation, only one process could succeed. the others would fail. Refer  [here](https://redis.io/commands/setnx#design-pattern-locking-with-codesetnxcode)

narfbg:
Ok, that makes sense for Redis, although it can be simplified:

```
    $result = ($ttl === -2)
        ? $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300))
        : set_result = $this->_redis->setex($lock_key, 300, time());
```

(yes, `$result` instead of `$set_result` :))

---

What about memcache though?

tianhe1986:
memecache also could use `add()` for locking refer [here](https://github.com/memcached/memcached/wiki/Programming#proper-use-of-add)

Although it also has problem, `add()` would be better than `set()` .

I'll modify the code according to your advice :)

system/libraries/Session/drivers/Session_memcached_driver.php
```diff
@@ -326,7 +326,11 @@ protected function _get_lock($session_id)
 				continue;
 			}
 
-			if ( ! $this->_memcached->set($lock_key, time(), 300))
+			$result = ($this->_memcached->getResultCode() === Memcached::RES_NOTFOUND)
```
narfbg: With the method params matching ...

    $method = ($this->_memcached->getResultCode() === Memcached::RES_NOTFOUND) ? 'add' : 'set';
    if ( ! $this->_memcached->$method($lock_key, time(), 300))

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -365,7 +365,10 @@ protected function _get_lock($session_id)
 				continue;
 			}
 
-			if ( ! $this->_redis->setex($lock_key, 300, time()))
+			$result = ($ttl === -2)
+				? $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300))
+				: $this->_redis->setex($lock_key, 300, time());
+			if ( ! $result)
```
narfbg: Needs a single empty line prior to this.

PR #5173, opened 2017-07-06
Add array identifier for library loader

Syafiqq:
Add array type to $library param at PHPDoc, so the IDE will not show a warning when we passing an array to it.


narfbg:
Just change it to `mixed` (by pushing another commit to this branch; don't open a [new one](https://github.com/bcit-ci/CodeIgniter/pull/5172)).

PR #5182, opened 2017-07-11
Update date_helper.php

siturra:
substr() in php 7, If string is equal to start characters long, an empty string will be returned. Prior to this version, FALSE was returned in this case.

example: $time = '20170711';

Error display:
Severity: Warning
Message: mktime() expects parameter 1 to be integer, string given.


narfbg:
Duplicate of #4289 (and #5181).

PR #5183, opened 2017-07-12
add is_write_type method for merge query

ytetsuro:
I would like to get `affected_rows` when I run the following code:

```php
<?php
$sql = <<<EOD
MERGE INTO DUAL A
USING DUAL B ON (A.ID = B.ID)
WHEN MATCHED THEN
    UPDATE SET A.FOO = B.BAR
WHEN NOT MATCHED THEN
    INSERT (ID, FOO) VALUES (B.ID, B.BAR)
EOD;

$this->db->query($sql);
var_dump($this->db->affected_rows());
```

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -982,7 +982,7 @@ abstract class CI_DB_driver {
 	 */
 	public function is_write_type($sql)
 	{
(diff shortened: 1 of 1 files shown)


narfbg:
Why are you overriding the method for OCI8 instead of just adding `MERGE` to the original one?

ytetsuro:
I thought `MERGE` was the only function of` Oracel`.
But that is wrong. It function `MERGE` also in` SQL Server`.
Thank you for your reply.

ytetsuro:
merged thanks!

PR #5185, opened 2017-07-12
is_in validation

hipoloco:
Add a new validation function for check if the input value matches with some value in the specific database field.

In my project I needed to verify if the value entered in an input corresponded to the code of a country. Adding all the codes in the "in_list" function was not optimal (they are 253 values), plus they were already loaded in the database. That's why I created this function.

PS: Sorry for my english (I'm from Uruguay and my english isn't good).


narfbg:
Duplicate of #3574.

PR #5191, opened 2017-07-20
Fix some mistakes in the user guide

jim-parry:
Three of the user guide pages had mismatched quotes, causing sphinx problems.
urls.rst still has a block that sphinx doesn't like :(

### DIFF
--- a/user_guide_src/source/general/urls.rst
+++ b/user_guide_src/source/general/urls.rst
@@ -42,13 +42,11 @@ By default, the **index.php** file will be included in your URLs::
 If your Apache server has *mod_rewrite* enabled, you can easily remove this
 file by using a .htaccess file with some simple rules. Here is an example
 of such a file, using the "negative" method in which everything is redirected
-except the specified items:
-
-::
+except the specified items::
 	
 	RewriteEngine On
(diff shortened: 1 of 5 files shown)


user_guide_src/source/general/urls.rst
```diff
@@ -42,13 +42,11 @@ By default, the **index.php** file will be included in your URLs::
...
 of such a file, using the "negative" method in which everything is redirected
-except the specified items:
-
-::
+except the specified items::
 	
 	RewriteEngine On
-	RewriteCond %{REQUEST_FILENAME} !-f
-	RewriteCond %{REQUEST_FILENAME} !-d
+	RewriteCond %{REQUEST_FILENAME} -f [OR]
```
narfbg: I've forgotten almost everything about .htaccess rules by now, but this doesn't seem like a fix for anything.

jim-parry: That specifically was trying to identify the issue that sphinx has with that code block. The two should be equivalent, but I agree it doesn't solve the underlying problem. Will remove.

narfbg: Well, we need to find out why Sphinx has a problem with this ...

PR #5193, opened 2017-07-20
Security fix

freetom:
escape `dest_folder` so that a specially crafted destination folder cannot inject arbitrary commands to `exec`

### DIFF
--- a/system/libraries/Image_lib.php
+++ b/system/libraries/Image_lib.php
@@ -972,7 +972,7 @@ class CI_Image_lib {
 			$cmd_inner = 'pnmscale -xysize '.$this->width.' '.$this->height;
 		}
 
-		$cmd = $this->library_path.$cmd_in.' '.escapeshellarg($this->full_src_path).' | '.$cmd_inner.' | '.$cmd_out.' > '.$this->dest_folder.'netpbm.tmp';
+		$cmd = $this->library_path.$cmd_in.' '.escapeshellarg($this->full_src_path).' | '.$cmd_inner.' | '.$cmd_out.' > '.escapeshellcmd($this->dest_folder).'netpbm.tmp';
 
 		$retval = 1;
 		// exec() might be disabled


narfbg:
There's no such thing as a specifically crafted destination folder. This is never a user-provided value.
Plus, `escapeshellarg()` isn't bullet-proof.

Also, we have a `security@` email and a HackerOne page - if you think there's a security issue somewhere, please don't post it publicly here.

freetom:
Then it's weird that you escape `full_dest_path`, that is composed [here](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Image_lib.php#L602) by `dest_folder`, `filename`, etc.
For sure controlling the destination folder is less common than the file name but it could happen, why not.

For the future, I'll eventually post on the apposite channels

narfbg:
`full_dest_path` is escaped because it includes `filename`. And while that shouldn't happen, it's a relatively common mistake for developers to use original filenames (i.e. as provided during a file upload).

This isn't the case with `dest_folder`, and sorry but "why not" isn't serious reasoning. :)

freetom:
Yeah, but as the user might control the file, it could happen with the folder.. I know it sounds weird but for example the user might control the creation of folders when he creates an account or through some other nice feature. A folder is just a special type of file.

Furthermore, consider the case in which a singular developer names the `dest_folder` with some special bash characters.. then, that `exec` fails.

narfbg:
> Yeah, but as the user might control the file, it could happen with the folder.. I know it sounds weird but for example the user might control the creation of folders when he creates an account or through some other nice feature.

This is a very contrived example. But even if it was valid, let me explain why I mentioned `escapeshellarg()` not being bullet-proof - if the only option you have doesn't 100% solve the problem, then the problem is unsolvable. Partial security is just a false sense of security.

> Furthermore, consider the case in which a singular developer names the `dest_folder` with some special bash characters.. then, that `exec` fails.

I have considered this ... 1) not a single user has complained about this; and 2) while I don't mind doing something about it, it won't happen under the "security" label.

Yes, I am very nitpicky - sorry.

freetom:
Never mind, but putting that call can only help, not worsen that LoC

PR #5195, opened 2017-07-20
Instantiation of Hooks class after Config class.

ivantcholakov:
When Hooks class get loaded, within its constructor it loads the Config class. https://github.com/bcit-ci/CodeIgniter/blob/6cab78970f212a30ea923835510ffd1fb77a6ad0/system/core/Hooks.php#L90

I think, there is no reason within the file CodeIgniter.php Hooks to be presented as loaded before the Config class, because it is not the actual order of loading.

Also, after the change, the injected through $assign_to_config options would become accessible by 'pre_system' hooks.


narfbg:
This is a BC break *and* it makes the "pre_system" hook, be NOT pre-system.

PR #5197, opened 2017-07-24
Re-enable Query Builder for ODBC driver

simplysites:
Undo of a change made in CI 3.1.0: ODBC driver does not support Query Builder usage any more.

### DIFF
--- a/system/database/drivers/odbc/odbc_driver.php
+++ b/system/database/drivers/odbc/odbc_driver.php
@@ -50,7 +50,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
  * @author		EllisLab Dev Team
  * @link		https://codeigniter.com/user_guide/database/
  */
-class CI_DB_odbc_driver extends CI_DB_driver {
+class CI_DB_odbc_driver extends CI_DB {
 
 	/**
 	 * Database driver


narfbg:
It was disabled for a reason.

simplysites:
And what might that reason be? Breaking existing code without a proper explanation (not found in the change log) is not exactly best practice. We rely on query builder throughout a lot of projects and as such can not update to 3.1.0 or more recent versions.

narfbg:
I know it doesn't say much, but the changelog entry is this:

>- Fixed an SQL injection in the ‘odbc’ database driver.

Disabling the query builder **is** the fix, as QB relies on escaping, which is something ODBC doesn't have APIs for.

Proper explanation was in the release announcement: https://forum.codeigniter.com/thread-65803.html
And that includes saying it was 3.1.0 instead of 3.0.6 exactly because of the breaking change.

simplysites:
Dear Andrey,

Thanks for your explanation, now it is clear to me. Is this a temporary fix or a final solution?
Should there be an interest in implementing a proper prepared statement and escaping mechanism in the driver, I am willing to spend some time on that.

Best regards,
Mattias

narfbg:
Escaping isn't possible, only prepared statements with bound params - until we can do that, it's as good as permanent.

But that isn't an easy or small task, and if you want to contribute to it, you have to know that this can't be limited to ODBC. *All* drivers have to be transitioned to proper prepared statements, keeping the same APIs, but with as less BC breaks as possible.

simplysites:
Personally I only have experience with MySQL/MySQLi and ODBC, but I'm certainly willing to help out.

PR #5203, opened 2017-07-25
Added protocol param for cross protocol requests

biscofil:
With this parameter you can use csrf protection between different protocol pages of the same domain


narfbg:
Sorry, but we don't need this.

Nobody does. If you have HTTPS - use it everywhere, don't constantly switch between encrypted and plaintext connections. And especially don't switch while submitting forms that require encryption.

PR #5208, opened 2017-07-28
Offset as data attribute in Pagination links

nicolastavella:
Uses the offset as value of the link attribute.
Added the property $data_offset_attr in Pagination, that determines the name of the attribute. 
Only visible if $use_pages_numbers is false.


narfbg:
The patch doesn't comply with our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html) and you're not explaining what the purpose of it is.

panigrc:
The issue should probably close, as it's old and the pull request wasn't accepted.

PR #5209, opened 2017-07-29
fix for get_where not honoring table aliasses

DennisKofflard:
the line:
`$this->_track_aliases($table);`
is missing from get_where(), this is inconsistent with get()
https://github.com/bcit-ci/CodeIgniter/blob/4aaf3847dbd3a33bdd3341a93462be4f0cbcaeef/system/database/DB_query_builder.php#L1448
https://github.com/bcit-ci/CodeIgniter/blob/4aaf3847dbd3a33bdd3341a93462be4f0cbcaeef/system/database/DB_query_builder.php#L1531

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1528,6 +1528,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	{
 		if ($table !== '')
 		{
+			$this->_track_aliases($table);
 			$this->from($table);
 		}
 


narfbg:
Please explain what the issue is.

gduval91:
_track_aliases() is "Used to track SQL statements written with aliased tables." 
So DennisKofflard adds a _track_aliases() in the get_where(), just like it's done in get() or get_compiled_select() (for consistent issues).

I'm not sure this is necessary as there is already a _track_aliases() in the from() called in the get_where(), the same for get(), the _track_aliases() is already in from(), so maybe the real problem is to remove the _track_aliases() in the get() (and some more places).
This should be tested more!

narfbg:
The only thing I need to know is the only thing not explained yet - (again) what does this patch fix?

HyndexCoder:
This patch fixes consistency issues. All functions should behave as expected. If you're a child, you're expected to play. If you don't play, then there's something wrong.

jim-parry:
@HyndexCoder say what? "If you're a child, you're expected to play. If you don't play, then there's something wrong." 
What does that have to do with the PR??

HyndexCoder:
@jim-parry it's a metaphor to the situation.

narfbg:
> This patch fixes consistency issues.

When we talk about "fixing" something, that should mean the said thing is broken. There's no evidence of anything broken here, and "consistency issues" or what we more commonly refer to as "inconsistency" is a blanket term for "these 2+ things aren't the same and that annoys me". 

> All functions should behave as expected.

Care to explain which function doesn't *behave* as expected then?

> If you're a child, you're expected to play. If you don't play, then there's something wrong.

I don't know what kind of a metaphor that is, but it doesn't sound even remotely appropriate to mention here. You know, if you don't have anything to add to a conversation, you do have the option of just not saying anything.

PR #5215, opened 2017-08-02
Fix width and height validation

fengkaijia:
The new validation introduced in eac4adf is causing previous-worked job like cropping to break if the input width and height are floating point numbers (e.g. values returned by Jcrop) since no width and height will be given to the image processor.

This change allows float-type width and height pass.

### DIFF
--- a/system/libraries/Image_lib.php
+++ b/system/libraries/Image_lib.php
@@ -488,9 +488,12 @@ class CI_Image_lib {
 							continue;
 						}
 					}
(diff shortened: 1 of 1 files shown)


narfbg:
This change effectively removes the validation ... it will allow anything. Just cast your floats to integer.

narfbg:
On another note, I'd accept a change of the `ctype_digit()` check to a `is_numeric()` one (removing the string cast). That should allow floats to pass.

I'll reopen the PR if you want to do that.

PR #5221, opened 2017-08-16
MySQL integer field is returned as itself or string in PHP

aunhappy:
Convert integer and float columns back to PHP numbers. Only valid for mysqlnd.

### DIFF
--- a/application/config/database.php
+++ b/application/config/database.php
@@ -69,6 +69,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 |
 | The $query_builder variables lets you determine whether or not to load
 | the query builder class.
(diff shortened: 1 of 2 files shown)


narfbg:
Sorry, but no. Result sets must be consistent between all drivers we have.

aunhappy:
@narfbg well ,I add this feature for pdo dbdriver just now.
Can you open this pr again?In my opinion ,pdo and mysqli drivers are universal for developer.Also,I know less about other drivers. You know , it is hard for me to add this feature for each driver.
If can not, I hope that this feature for all drivers will be added the ci future version by other developer.
Thx!

narfbg:
No ... you don't get it.

The feature cannot possibly get into even a quarter of all drivers, and that's why I cannot accept it for any of them.

PR #5227, opened 2017-08-17
[cs] use short syntax for arrays

TomasVotruba:
In case you don't have some reason for keeping old array, since PHP 5.4 is min version short array can be used.


narfbg:
We do have a reason not to do this yet.

But even if we didn't, change for the sake of change is something we don't do.

Although

TomasVotruba:
What a nice tautology :D sure.

PR #5230, opened 2017-08-19
fix CI_Hooks object method call

netsan:
```php
$hooks = load_class('Hooks', 'core');
$hooks->enabled = true;
$hooks->hooks['test'] = [new Foo(), 'test'];
$hooks->call_hook('test');// not run
class Foo
{
    public function test()
    {
        echo 'hello,world';
    }
}
```


narfbg:
The sample code you've provided is incorrect usage of `CI_Hooks`.

https://codeigniter.com/userguide3/general/hooks.html#defining-a-hook

PR #5233, opened 2017-08-22
Add Indonesian Translation

ksatria:


narfbg:
Sorry, we don't accept translations into core.

PR #5239, opened 2017-08-27
Added STRAIGHT_JOIN and NATURAL JOIN types

jvyyuie:
supported the STRAIGHT_JOIN and NATURAL JOIN of mysql join types.


narfbg:
Sorry, we don't accept MySQL-only extensions like `STRAIGHT_JOIN`.

We could accept `NATURAL`, but it won't really work with your patch (there's an extra space where you check for it) and it is incomplete - you can prefix any type of join with the `NATURAL` keyword, it's not a stand-alone type.

PR #5240, opened 2017-08-28
added NATURAL JOIN types

jvyyuie:
according to PR #5239, remove STRAIGHT_JOIN types, rewrite code to support NATURAL JOIN types.


narfbg:
Not following our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).

Also, you're not supposed to add unnecessary stuff (STRAIGHT_JOIN) in one commit only to remove it in another.

jvyyuie:
I read styleguide and find my patch's mistake:

`$natural_join = FALSE;`

should be:

`$_natural_join = FALSE;`

have there any other incorrects?

narfbg:
No, the variable name is fine. The problem is with spaces around parentheses.

jvyyuie:
thank you

Issue #5164, opened 2017-06-27
Loader.php _ci_load_library method has a bug


ppanphper:
![image](https://user-images.githubusercontent.com/6057923/27582913-3bdf7afc-5b65-11e7-89a9-d4f0ba24b91f.png)
if first loaded library set object_name Initialized library.
second loaded library not set object_name, result is return. 
so, loaded library fail.

:) sorry, I do not speak English, hope you can understand what I mean！

narfbg:
Please test with the commit linked above.

Also, for future reference - please provide actual code and/or text instead of screenshots.

GrahamDj:
Isn't it maybe better to always load a library as an alias?

`if( $object_name === NULL )`
`{`
`    $object_name = $library;`
`}`

`$this->_ci_load_library($library, $params, $object_name);`
`return $this;`

ppanphper:
@narfbg 

eg:
$this->load->library('test', null, 'test1');
$this->load->library('test');
because first load library included 'test' class,
so second load library don't included 'test' class, and no set object_name, then return void.

I use version is 2.2.1, may be used method not be the same.

ppanphper:
Fixed the bug in the version 2.2.1

if (in_array($filepath, $this->_ci_loaded_files))
{

	// Before we deem this to be a duplicate request, let's see
	// if a custom object name is being supplied.  If so, we'll
	// return a new instance of the object
	$CI =& get_instance();
	$strLowerClass = strtolower($class);
	$classVar = $this->getClassVar($strLowerClass, $object_name);
	if ( ! isset($CI->$classVar))
	{
		return $this->_ci_init_class($class, '', $params, $object_name);
	}
	$is_duplicate = TRUE;
	log_message('debug', $class." class already loaded. Second attempt ignored.");
	return;

}



`protected function getClassVar($class, $object_name) {`
        `if (is_null($object_name))`
	`{`
		`$classvar = ( ! isset($this->_ci_varmap[$class])) ? $class : $this->_ci_varmap[$class];`
	`}`
	`else`
	`{`
		`$classvar = $object_name;`
	`}`
	`return $classvar;`
`}`

narfbg:
@GrahamDj No.
@ppanphper CI2 is no longer supported. Stop using it.

ppanphper:
@narfbg Production project... :(

narfbg:
That only means you need to upgrade quicker.

GrahamDj:
[...]

PR #5249, opened 2017-09-07
check if $class is actually a property of $CI before deeming it to be a duplicate request

kasperpeulen:
We sometimes load libraries using require_once, for example say that I have

```
class My_library extends My_other_library
```

if we then load My_library using CI, we then load My_other_library using require_once at the top of the file, so now the class My_other_library exists

but somewhere else in the code, we may want to load My_other_library directly, but now it won't load, because the class already exists. With this PR, it will actually add the class to $CI super object

We are migrating from CI2 to CI3, and this is one of the issue we have, would be great if something like this could land in CI3.

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -1039,7 +1039,7 @@ class CI_Loader {
 			$filepath = $path.'libraries/'.$subdir.$class.'.php';
 
 			// Safety: Was the class already loaded by a previous call?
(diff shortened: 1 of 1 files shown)


narfbg:
Duplicate of #5164.

kasperpeulen:
Thanks for fixing this already! Do you know when 3.1.6 will be released?

PR #5251, opened 2017-09-09
Feature: add track_id in log file

caty:
This saves your time when you search something in ton's of the messages

Signed-off-by: Huang Yuzhen <cacaty@gmail.com>


gxgpet:
If you can remember a random string with 20 characters length, why wouldn't you remember the actual line number of an entry? :)

narfbg:
Maybe it helps you, somehow. But that's because it's your own, homegrown hack ... not something that most people would enjoy.

PR #5257, opened 2017-09-12
fix phpdocs for library and model from CI_Loader

kasperpeulen:


narfbg:
Sorry, but we restrict that to a single type declaration.

PR #5262, opened 2017-09-14
Core/Codeigniter : fix too few args for php >= 7.1

firespeed:
After 7.1 migration we can't send too few arguments on uri without error : 

http://php.net/manual/en/migration71.incompatible.php
Fatal error: Uncaught ArgumentCountError: Too few arguments to function  ....

On this fix i try to count needed args and i compare with function proto

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -339,6 +339,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
  *	2) a query string which doesn't go through a
  *	   file_exists() check
  *	3) a regular request for a non-existing page
+ * 	4) Arguments count test
  *
  *  We handle all of these as a 404 error.
  *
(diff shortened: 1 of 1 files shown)


narfbg:
Too few arguments on ... what?

firespeed:
Test url :     test.com/welcome
Controller : welcome($param)

Result : An uncaught Exception was encountered
Type: ArgumentCountError
...

narfbg:
Then there's nothing to fix ... you should be getting an error in that case.

firespeed:
You shouldn't have a server error ... 
You should have a 404

narfbg:
There's nothing wrong with the URL.

Your app (i.e. the server) has erroneous logic, where a parameter is required but not actually guaranteed. That's a server error.

firespeed:
I don't agree with you with "erroneous logic" for example : 

- https://github.com/bcit-ci/CodeIgniter/pull/5262 <= works
- https://github.com/bcit-ci/CodeIgniter/pull/ <= 404 error because pull id is required

It doesn't work (you have a 404) but you don't have an application error.
Ok you can add a default value on each function and check them but i don't think it's better.

But ok i understand your opinion thank you for the previous response

PR #5267, opened 2017-09-17
Expanded Log Levels to match PSR-3

lucanos:
Expands the log levels from 4 ("ERROR", "INFO", "DEBUG", "ALL") to the 8 levels set out in PSR-3 ("EMERGENCY", "ALERT", "CRITICAL", "ERROR", "WARNING", "NOTICE", "INFO", "DEBUG").
Sets the default log level to "WARNING" or higher.

Is backwards compatible (except where "ALL" is used as the level, which I have not seen.

Allows the level to be specified by string or array key when calling `log_message($level, $message)`

https://www.sitepoint.com/logging-with-psr-3-to-improve-reusability/

### DIFF
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -64,10 +64,11 @@ class CI_Log {
 
 	/**
 	 * Level of logging
+	 * Defaults to 4 - "WARNING" and higher
 	 *
 	 * @var int
 	 */
-	protected $_threshold = 1;
+	protected $_threshold = 4;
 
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but no.

As far as I know, CI4 will follow PSR-3, so you can look forward to that if you like.

lucanos:
Happy to look into CI4. Do you have any resources which may assist me?

PR #5270, opened 2017-09-21
Automatically check duplicate rows for insert_batch function- DB_query_builder.php

tuhin18003:
# Automatically check duplicate rows for insert_batch function. 

**Example:**
   *If you want to check all the fields just use the following function:*
```
$this->db->insert_batch('table', $data);
```
*If you want to use any specific fields for duplicate check and update you can use as following*
```
$duplicate_fields = array(
'firstname',  'lastname'
);
$this->db->insert_batch('table', $data, $escape = NULL, $batch_size = 100, $duplicate_fields );
```

*And if you don't want to check for duplicate, then you can use as follwing*
```
$this->db->insert_batch('table', $data, $escape = NULL, $batch_size = 100, false );
```

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -148,6 +148,13 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 * @var	array
 	 */
 	protected $qb_set			= array();
+        
+	/**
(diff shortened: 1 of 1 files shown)


zploskey:
Checking all fields for duplicates by default is undesirable behavior.

PR #5272, opened 2017-09-21
Check duplicate row & update - insert_batch

tuhin18003:
### Check duplicate row & update - insert_batch function. 

**Example:**
   *If you want to update all the fields just use the following function:*
```
$duplicate_fields = array(
'all'
);
$this->db->insert_batch('table', $data, $escape = NULL, $batch_size = 100, $duplicate_fields );
```
*If you want to update any specific fields you can use as following*
```
$duplicate_fields = array(
'firstname',  'lastname'
);
$this->db->insert_batch('table', $data, $escape = NULL, $batch_size = 100, $duplicate_fields );
```

*And if you don't want to check for duplicate, then you can use as following function.*
```
$this->db->insert_batch('table', $data);
```

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -155,6 +155,13 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 * @var	array
 	 */
 	protected $qb_set_ub			= array();
+        
+        /**
+	 * QB update data sets when insert
+	 *
+	 * @var	String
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, we don't want this feature.

And this is kind of a duplicate of #4452 ... in addition to the other 5 pull requests you opened in quick succession. Please stop, it's just spamming us.

PR #5278, opened 2017-09-29
Fix auto_link trailing slash

Uekawa:
`auto_link('http://example.com/example/');`

Result
`<a href="http://example.com/example">http://example.com/example</a>`**/**

Fix Result
`<a href="http://example.com/example/">http://example.com/example/</a>`


narfbg:
Please don't change file permissions.

I only noticed this one after merging ... would've refused it if I had seen it before that.

Uekawa:
Sorry. Permissions were unintentionally changed

PR #5279, opened 2017-09-29
moving constants.php content into core/Common.php

gbarosio:
While browsing CI's code, I thought about moving the content of /config/constants.php into Common.php.

Arguments, IMHO, a  secure place for noob thingers, and the removal of a file (constants.php). 

* I don't think this is an issue for anyone, but just polishing. (eventually)
* I don't think there's documentation to be modified, or at least I cold not find it. 

Last, I did not pay attention to the sign off chapter in the contributing mentions on how to open a pull request. Apologies. Please do let me know if this whole idea is not worth, or if I need to do anything else. 

And hey, thanks for the code :)

Best,


narfbg:
Anything under application/ should be considered configuration. You don't hard-code configuration ... and this doesn't really help with anything.

gbarosio:
Sounds good, thanks for your time!

PR #5282, opened 2017-10-02
Fix ORDER BY clauses being reset by count_all_results()

pgee70:
recent changes broke query ordering giving a warning in line 1238 = '$this->qb_cache_orderby = array_merge($this->qb_cache_orderby, $qb_orderby);'
the warning message was about array_merge not working with a NULL object.

this was because on subsequent re-queries, the qb_cache_orderby was set to NULL not Array() as it is should have been.  See the initial instantiation of the $this->qb_cache_orderby 
This PR should fix it.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1405,7 +1405,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		// for selecting COUNT(*) ...
 		$qb_orderby       = $this->qb_orderby;
 		$qb_cache_orderby = $this->qb_cache_orderby;
-		$this->qb_orderby = $this->qb_cache_orderby = NULL;
+		$this->qb_orderby = $this->qb_cache_orderby = array();
 
 		$result = ($this->qb_distinct === TRUE OR ! empty($this->qb_groupby) OR ! empty($this->qb_cache_groupby) OR $this->qb_limit OR $this->qb_offset)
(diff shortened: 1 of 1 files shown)


narfbg:
The "recent changes" in question is the patch for #5168.

But while I can confirm that this is a bug, the warning itself is only a symptom. What's the actual effect that this has on queries? How do they break?

pgee70:
This came up as a warning in my unit tests - I didn’t notice this affecting anything on prod.  But I use query caching to help with data tables.

if you array merge with null the output will be null.  So:

php > $a = null;
php > $b = [1,2,3];
php > $c = array_merge($a,$b);
PHP Warning:  array_merge(): Argument #1 is not an array in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. array_merge() php shell code:1

Warning: array_merge(): Argument #1 is not an array in php shell code on line 1

Call Stack:
   62.1301     351296   1. {main}() php shell code:0
   62.1301     351296   2. array_merge() php shell code:1

php > var_dump($c);
php shell code:1:
NULL

So, this means that the orders already set would be lost on subsequent orders, after that buggy line was executed.

narfbg:
Ok, thanks!

Changed the title to reflect what the PR is actually about. As a side note, please target the develop branch unless you know that something is only relevant to 3.1-stable. We do backport all PR patches manually.

PR #5286, opened 2017-10-04
Fix Error "Cannot create references to/from string offsets"

### LINKED ISSUE #5202: Form validation (Fields with multiple key)
**File:** system/libraries/Form_validation.php
**Line:** 603

**The correction is commented below:**
```
/**
 * Re-populate the _POST array with our finalized and processed data
 *
 * @return	void
 */
protected function _reset_data_array(&$data)
{
	foreach ($this->_field_data as $field => $row)
	{
		if ($row['postdata'] !== NULL)
		{
			if ($row['is_array'] === FALSE)
			{
				isset($data[$field]) && $data[$field] = $row['postdata'];
			}
			else
			{
				// before we assign values, make a reference to the right POST key
				if (count($row['keys']) === 1)
				{
					$data_ref =& $data[current($row['keys'])];
				}
				else
				{
					foreach ($row['keys'] as $val)
					{
-->						$data_ref =& $data_ref[$val]; // Replace with $data_ref =& $data[$val]
					}
				}

				$data_ref = $row['postdata'];
			}
		}
	}
}
```

zploskey:
Fix attempting to access data_ref before it is defined when validating an array input with multiple explicitly specified keys.

Fixes #5202.

### DIFF
--- a/tests/codeigniter/libraries/Form_validation_test.php
+++ b/tests/codeigniter/libraries/Form_validation_test.php
@@ -482,6 +482,20 @@ class Form_validation_test extends CI_TestCase {
 		$this->assertEquals('bar2', $this->form_validation->set_value('bar[]', $default));
 	}
 
+	public function test_issue_5202()
+	{
+		$data = array('person' => array('firstname' => 'Dick', 'lastname' => 'Tracy '));
+		$this->form_validation->reset_validation();
+		$this->form_validation->set_rules('person[firstname]', 'First Name', 'required|trim');
+		$this->form_validation->set_rules('person[lastname]', 'Last Name', 'required|trim');
+		$this->form_validation->set_data($data);
+		$valid = $this->form_validation->run('', $data);
+
+		$this->assertEquals(TRUE, $valid);
+		$this->assertEquals('Dick', $data['person']['firstname']);
+		$this->assertEquals('Tracy', $data['person']['lastname']);
+	}
+
 	public function test_set_select()
 	{
 		// Test 1: No options selected
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -591,6 +591,8 @@ class CI_Form_validation {
 				}
 				else
 				{
+					$data_ref =& $data;
+
 					// before we assign values, make a reference to the right POST key
 					if (count($row['keys']) === 1)
 					{


narfbg:
Linking #5202 because the title-mention doesn't do that for some reason.

zploskey:
I simplified the test case. Interestingly, this test seems to run indefinitely without the fix in place. I'm not entirely sure why that happens, but the equivalent code in a controller (as shown in the issue) produces the error.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -482,6 +482,15 @@ public function test_set_value()
 		$this->assertEquals('bar2', $this->form_validation->set_value('bar[]', $default));
 	}
 
+	public function test_issue_5202()
+	{
+		$this->form_validation->reset_validation();
+		$_POST = array('person' => array('firstname' => 'Dick', 'lastname' => 'Tracy'));
+		$this->form_validation->set_rules('person[firstname]', 'First Name', 'required');
+		$this->form_validation->set_rules('person[lastname]', 'Last Name', 'required');
```
narfbg: While this would go to the relevant execution path and thus show if *PHP errors* are triggered, it doesn't actually validate what `_reset_data_array()` is supposed to do.

That is, to modify the original input array. For example, if you add `trim` as a rule, and the input was "Tracy " (with the trailing space), then we should validate if that trailing space is stripped after validation.

zploskey: Definitely a good idea. This revealed a problem with the modified data being placed at the base level of the array instead of in 'person' for the test case. The fix now just assigns `$data_ref &= $data;` before assigning values, basically as the code was before 0c06ba0, so that data gets put into the proper position nested in the array.

zploskey:
It looks like the only failure is an incidental one where phpenv failed to download php 5.5.

narfbg:
Yes, I restarted the test job and it's all green now.

One more thing though ... your test modifies `$_POST`, which is global state, and then doesn't reset its state after completion. This can cause side-effects to other tests.
Have it in mind that the logic to modify `$_POST` has always worked, and the change that caused the regression you're fixing were intended to have the same effect on custom data passed via `set_data()`, so ... Should either actually test that, or reset the `$_POST` array to its original state at the end of the unit test.

zploskey:
I went ahead and modified the test to use `set_data()` and avoid doing anything with `$_POST`.

On a related note, there seem to be many tests in this class that modify post data without reverting it at the end of the test, so that might be a good candidate for code cleanup. Possibly this could be as easy as setting a default value for `$_POST` in `set_up()` that gets run before each test.

narfbg:
Yea, such stuff should be cleaned-up.

PR #5288, opened 2017-10-05
 Adding new letters and cleanups in foreign_chars.php file

pawelkg:
Second try


narfbg:
What is the source of this?

You're making a lot of changes, and it is questionable how you'd know what is right (unless you're a linguist or something). For example, you've added `ʙ` as a capital B, but a) it looks like a non-capital; b) in cyrilic-based languages, that would translate to a V instead.
Also, changing the translations of some existing entries (e.g. `ä` from `ae` to `a`).

It certainly isn't the case of you just adding/correcting stuff for your own native language, so how do you know that somebody even benefits from it?

Paradinight:
@narfbg In germany ä would be ae not a. The same is ÄöÖüÜ. 

https://github.com/WordPress/WordPress/blob/master/wp-includes/formatting.php#L1142

ä is a or ä is ae depend on the location

pawelkg:
Mostly I use Wiktionary page: https://en.wiktionary.org/wiki/Appendix:Latin_script
As @Paradinight said in translation of eg. ä depend on language context. In fact 'ä' was moved by my by accident, but good is that because of that @Paradinight point out language context. I thought about it and I think that ä, ö, ü (with capital variations) should be translated to a, o, u respectively. In Kashubian language (https://en.wikipedia.org/wiki/Kashubian_language) those letters should be transcript to a, o, u rather to ae, oe or ue.

narfbg:
I had to google what Kashubian is and that alone should show why we shouldn't base any decisions based on how that language works ... It doesn't help that, according to what I found, it is being used by less than 0.1% of the world's population.

As for that Wikitionary page, this is also the first time I hear about "Wikitionary", meaning I have no idea how reliable it may be. And a list of alphabet variations of latin letters only shows that "modified" versions of letters exist, not how they'd be transliterated.

Sorry, but there are huge question marks all over this patch and nobody has actually asked for these changes, so I'm closing the PR. Again, if you want to make some additions, please limit it to your native language(s).

PR #5289, opened 2017-10-06
Clean up form validation tests

zploskey:
There are a number of places where `$_POST` is modified and not
reset back to its previous state. This could lead to side effects
impacting other tests. There are also a number unnecessary calls to
`reset_validation()`. A new form_validation object is already
initialized for every test in `set_up()`, so these lines do nothing
unless they are cleaning up from earlier in the same test.

To remedy this, we make the following changes:

Set `$_POST = array()` in `set_up()`, resetting to a default state for
every test run. Add a `tear_down()` method that does the same and is
called by PHPUnit's `tearDown()` after each test.

Remove now-redundant lines where tests set `$_POST = array()` at the
beginning or end of the test.

Remove unnecessary calls to `reset_validation()` intended as setup
or teardown.

### DIFF
--- a/tests/codeigniter/libraries/Form_validation_test.php
+++ b/tests/codeigniter/libraries/Form_validation_test.php
@@ -321,9 +321,6 @@ class Form_validation_test extends CI_TestCase {
 
 	public function test_set_data()
 	{
-		// Reset test environment
-		$_POST = array();
-		$this->form_validation->reset_validation();
 		$data = array('field' => 'some_data');
 		$this->form_validation->set_data($data);
 		$this->form_validation->set_rules('field', 'label', 'required');
@@ -342,9 +339,6 @@ class Form_validation_test extends CI_TestCase {
 
 	public function test_set_message()
 	{
-		// Reset test environment
-		$_POST = array();
-		$this->form_validation->reset_validation();
 		$err_message = 'What a terrible error!';
 		$rules = array(
 			array(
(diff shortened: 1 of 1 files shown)


narfbg:
This works, but ... it's the brute-force way.

`$_POST` should be reset by the tests that modify it, not via setup and teardown.

zploskey:
I looked into the PHPUnit documentation. From the current stable version documentation (https://phpunit.de/manual/6.4/en/fixtures.html#fixtures.global-state):
> Prior to version 6, by default, PHPUnit ran your tests in a way where changes to global and super-global variables ($GLOBALS, $_ENV, $_POST, $_GET, $_COOKIE, $_SERVER, $_FILES, $_REQUEST) do not affect other tests.
>
>As of version 6, PHPUnit does not perform this backup and restore operation for global and super-global variables by default anymore. It can be activated by using the --globals-backup option or setting backupGlobals="true" in the XML configuration file.

We are currently on 5.7, so `$_POST` and all of those other globals are reset by default. Thus we could remove all the cleanup of resetting `$_POST` altogether. In the event that we switch to PHPUnit 6+, it's probably a good idea to add `backupGlobals="true"` to the config. However, considering that PHPUnit 6 requires PHP 7, that may never happen and people will instead migrate to CodeIgniter 4.

With that in mind, what do you think about:
- Removing all resetting of $_POST as cleanup, since it is automatically done for us.
- Adding `backupGlobals="true"` to phpunit.xml, just to be explicit.

narfbg:
I don't care about `backupGlobals`, but I do insist that `$_POST` is reset by any test case that modifies it, regardless of how PHPUnit works.

zploskey:
As requested, `$_POST` is now explicitly reset at the end of each test that modifies it.

PR #5290, opened 2017-10-06
Allow where_in() etc. to handle row constructor syntax

fengkaijia:
This PR allows ```where_in()``` and other ```_wh_in()``` methods to accept IN queries involving row constructors like:

```sql
SELECT ... WHERE (first_name, last_name) IN(('James', 'Bond'), ('Johnny', 'English'))
```

using following structure:

```php
$this->db->where_in(['first_name', 'last_name'], [
    ['James', 'Bond'],
    ['Johnny', 'English']
]);
```

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -862,7 +862,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 * @used-by	or_having_not_in()
 	 *
 	 * @param	string	$qb_key	'qb_where' or 'qb_having'
-	 * @param	string	$key	The field to search
+	 * @param	mixed	$key	The field to search
 	 * @param	array	$values	The values searched on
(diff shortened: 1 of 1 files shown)


system/database/DB_query_builder.php
```diff
@@ -899,6 +899,14 @@ protected function _wh_in($qb_key, $key = NULL, $values = NULL, $not = FALSE, $t
 		{
 			$wh_in = array_values($values);
 		}
+		
+		if (is_array($key))
+		{
+			$key = '('.implode(', ', $key).')';
+			$wh_in = array_map(function ($value) {
```
narfbg: This makes the *assumption* that we have a multidimensional array, masking potential errors.  
It may've not been an array at all originally (previous `if` condition).

On another note, you're changing the docblock for this method, but not for the actual public methods ...

PR #5296, opened 2017-10-09
add mime type for doc associated with linux desktop

sanikkenway:
I've encountered this when I try to upload .doc type on Linux desktop, adding this additional type solve that problem


narfbg:
This MIME is associated with doc**x**: https://www.iana.org/assignments/media-types/application/vnd.openxmlformats-officedocument.wordprocessingml.document

PR #5301, opened 2017-10-15
Bugfix: DB_query_builder order_by / count_all_results 

webwerken:
The default type for $this->qb_orderby and $this->qb_cache_orderby is an empty array. 
Setting it to NULL in method 'count_all_results' causes an error on line 1238 when trying to merge it with newly added values.
The other option to prevent this would be to check before merging or to typecast.


gxgpet:
Duplicate of #5282, which looks like it was merged into `3.1-stable`, and not `develop`. You are also trying to merge something directly into the `master` branch.

webwerken:
OK. Missed that one.  Glad is was already fixed.

PR #5311, opened 2017-10-25
Download links for versions 2.2.4 to 2.2.6 added to docs

gonzalo-massa:
CodeIgniter 2.2.6 is the latest supported version of CodeIgniter2 and the documentation (Installation Instructions » Downloading CodeIgniter) only shows CI2 links up to 2.2.3. This links can be useful when trying to upgrade from CI2 to CI3, at least to update to the latest CI2 version before shifting to CI3.


narfbg:
We seem to have forgotten adding those, indeed.

That's the only valid reasoning though. I *have* to point out the following:

> CodeIgniter 2.2.6 is the latest supported version of CodeIgniter2

CodeIgniter 2 is **NOT** supported anymore; hasn't been since October 2015.

> This links can be useful when trying to upgrade from CI2 to CI3, at least to update to the latest CI2 version before shifting to CI3.

Nope. Whether you're upgrading from 2.2.0, 2.2.3 or 2.2.6 makes zero difference. In fact, the reason why we forgot to add those links is because 3.0.0 was already released before 2.2.3.

gonzalo-massa:
Thank you for clarifying those points, I will jump from 2.2.1 to 3.0.0 directly, then follow updating. Does this merge automagically update the documentation on the website?

narfbg:
No, it will be updated with the next framework release.

PR #5315, opened 2017-10-27
Made changes to get() and post() methods

vinodselvin:
Function calls have the access to read POST and GET data, so, I am giving the option to clean to have enough security.
For example run below code, methos: _callMeViaGet() has been called via function call still it has access to POST data, Which shouldnt so, i made this commit.
```

<?php

if (!defined('BASEPATH'))
    exit('No direct script access allowed');

class Test extends CI_Controller {

    public function index() {
        
    }

    function callMeViaPost(){
    	error_reporting(E_ALL); ini_set('display_errors', 1);
    	$data = $this->input->post(array("id", 'key3'));

    	echo "POST Data:<br>";
    	print_r($data);

    	$this->_callMeViaGet($data['id'], "yes");
    }

    function _callMeViaGet($id, $option){
    	echo "GET Data:<br>";

    	echo "Param 1 Data: ";
    	print_r($id);
    	echo "Param 2 Data: ";
    	print_r($option);

    	if(!empty($this->input->post())){
    		echo "Voila! How i came here, This function should have only access to my get params i guess<br>";
    		print_r($this->input->post());
    	}
    	else{
    		echo "Hmm";
    		print_r($this->input->post());
    	}
    }
}
```

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -180,11 +180,19 @@ class CI_Input {
 	 *
 	 * @param	mixed	$index		Index for item to be fetched from $_GET
 	 * @param	bool	$xss_clean	Whether to apply XSS filtering
+	 * @param	bool 	$clean_get	unset GET Data
 	 * @return	mixed
 	 */
-	public function get($index = NULL, $xss_clean = FALSE)
+	public function get($index = NULL, $xss_clean = FALSE, $clean_get = FALSE)
 	{
-		return $this->_fetch_from_array($_GET, $index, $xss_clean);
+		$get = $this->_fetch_from_array($_GET, $index, $xss_clean);
+		
+		if($clean_post === TRUE){
+			
+			$_POST = array();
+		}
+		
+		return $get;
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


narfbg:
No, thanks.

vinodselvin:
This should not happen right? developers should have more control on POST and GET data through framework. This needs further clarification, and perfect reply not directly closing of pull request.

narfbg:
I don't even understand what you're saying shouldn't happen, to be honest.

Not every idea you can come up with will get merged, and nobody owes you a "perfect reply" or explanation for that. Sorry.

vinodselvin:
Ok bro.

vinodselvin:
What I was trying to say on that day was, when I create POST request to particular controller function (API),  to get post data I used, $this->input->post() and I am making function calls to another function from this function, in sub function also I used $this->input->post(), as it is also a API which works on both POST and GET, In that function it was taking POST data, I debugged and found that $this->input->post() getting POST data in subsequent function call also.

I think there should be a config, setting so that POST or GET can't be accessible to sub functions, I think I was clear with my question.

PR #5316, opened 2017-10-29
Migration has been changed so that all the remaining migrations will …

gagansharma:
…run at once - migrateup() will be use to run up method of all the remaining migrations and migratedown() will be use to run down method of all the migrations which were already executed. Migrations table contain all the versions which had been successfully executed. Output of both the methods is an array having status of all the migrations. One more type of migration type is added - datetime (this is of 14 digits which is same as timestamp) and timestamp is now the actual timestamp of 10 digits.

### DIFF
--- a/system/libraries/Migration.php
+++ b/system/libraries/Migration.php
@@ -153,12 +153,15 @@ class CI_Migration {
 		}
 
 		// Migration basename regex
-		$this->_migration_regex = ($this->_migration_type === 'timestamp')
-			? '/^\d{10}_(\w+)$/'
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but no.

This makes too many changes, BC breaks and the code itself is messy. And you've not even made an attempt to write up the motivation for it.

Also, I want to clarify something because I see people making this mistake way too often ... What you're calling an "actual timestamp" is called a UNIX timestamp. The term timestamp itself means something that represents a point in time, and time can be represented in many different formats. Thus, a "datetime" string is a timestamp by definition.

PR #5320, opened 2017-11-01
seed data and truncate table

gagansharma:
Created two methods for seeding initial data into the table and truncate the table in the DB_forge.php file


narfbg:
Why would this be merged if it wasn't in #5316? You've addressed none of my feedback there.

PR #5324, opened 2017-11-04
Fixed custom config values (enable_hooks) not properly initialize hooks #5323

### LINKED ISSUE #5323: Bug: custom config values with 'enable_hooks = true' not properly initialize hooks
### This bug is about overriding the default value of ‘enable_hooks’ in the config using the $assign_to_config provided in public/index.php.

- In public/index.php, line 167, it says the $assign_to_config variable is used to override the default config values. 
_*Note: See the 'CUSTOM CONFIG VALUES' section in public/index.php_

- To reproduce the bug, assume that a user wants to override the ‘enable_hooks’ default config by adding this line of code in public/index.php:
`$assign_to_config['enable_hooks'] = true;`
The bug arises, because:

1. Look into system/core/CodeIgniter.php:
    1. Starting from line 219, the config class is loaded in system/core/CodeIgniter.php, and $assign_to_config is passed to override right after.
	`$CFG =& load_class('Config', 'core');
	// Do we have any manually set config items in the index.php file?
	if (isset($assign_to_config) && is_array($assign_to_config))
	{
		foreach ($assign_to_config as $key => $value)
		{
			$CFG->set_item($key, $value);
		}
	}`

    2. However, in line 200, the Hooks class was instantiated before config class is loaded.
    `$EXT =& load_class('Hooks', 'core');`
2. Look into system/core/Hooks.php, from line 90 to line 98. The hooks class constructor also loads the config items. Then it verifies the ‘enable_hooks’ item that actually comes from the default config to determine if hooks feature is enabled or not. Obviously, the customized 'enable_hooks' value is not yet effective because of the loading order mentioned in point #1 above)
		`$CFG =& load_class('Config', 'core');
		log_message('info', 'Hooks Class Initialized');
		// If hooks are not enabled in the config file
		// there is nothing else to do
		if ($CFG->item('enable_hooks') === FALSE)
		{
			return;
		}
`

To fix this bug, I suggest a two-step approach: 1) instantiate the Hooks class after loading the Config class. 2) Using config_item('enable_hooks') in Hooks ctor which takes the current effective config sets instead of loading the default config.
*note: using the config_item('enable_hooks') makes more sense than calling the load_class('Config', 'core') class registry here.

JaydenLiang:
Fixed not possible to override the 'enable_hooks' config item using custom config value.

Fixed #5323.


narfbg:
You don't have to create an issue separately only to reference it in a pull request minutes later. That only makes communication harder, especially when you put all the detail in the issue and none it the PR ... You're even describing the patch in #5323 instead of here (more on that I'll leave inline).

(assume the entire conversation is here and I won't be replying to the issue)

I wouldn't classify this as a bug; it's a system limitation at best. And frankly, I'd rather drop `$assign_to_config` entirely than trying to deal with any potential bugs related to it ... That's a hacky old feature that maybe 1 in 1000 devs would use.
However, I like where you're going with it for purely architectural reasons. If `CI_Hooks` depends on `CI_Config` then it certainly makes sense to switch their order of loading.

system/core/Hooks.php
```diff
@@ -87,12 +87,11 @@ class CI_Hooks {
 	 */
 	public function __construct()
 	{
-		$CFG =& load_class('Config', 'core');
 		log_message('info', 'Hooks Class Initialized');
 
 		// If hooks are not enabled in the config file
 		// there is nothing else to do
-		if ($CFG->item('enable_hooks') === FALSE)
+		if (config_item('enable_hooks') === FALSE)
```
narfbg: You've said this in #5323:

> *note: using the config_item('enable_hooks') makes more sense than calling the load_class('Config', 'core') class registry here.

But that's not really true ... `config_item()` is easier to write, true, but there's nothing that determines what "makes more sense". I could go in detail about how `load_class(...)->item()` makes perfect sense, but that's a moot point.
What really shoud happen here is this:

    public function __construct(CI_Config $CFG)
    {
        // Use $CFG->item()
        // ...
    }

And this is how you pass the `CI_Config` instance from the bootstrap script:

    // $CFG already being declared:
    $EXT =& load_class('Hooks', 'core', $CFG);

JaydenLiang: I agree with you on passing the CI_Config instance to the hooks ctor from the bootstrap script.
Talking about the $assign_to_config, I use it when I design the unit testing using PHPUnit. I want to require the public/index.php as the bootstrap for the unit testing and also want to generate a different output so I come to the hooks feature. In the tests/ folder I have these lines of code:
`$assign_to_config['enable_hooks'] = true;//to override the output for unit testing only`
`require_once '../public/index.php';`

I find it an easy way to manipulate the basic configurations without touching the other parts of my product. If there is any easier way to do this business, I would definitely go that way. Otherwise, would you think $assign_to_config worth to keep?

narfbg: There's zero chance that anyone can get away with shipping an application that requires zero configuration changes. A small number of options are mandatory (like base_url) and others are environment/domain-dependant.
I don't know what you mean by an easier way, but the place to alter configuration is config.php, so ...

> Otherwise, would you think $assign_to_config worth to keep?

That's really vague, yet if I say "yes" you'd have a reason to want first-class support for it and if I say "no" you may immediately submit a PR to remove it ...

Let me put it this way: It's one of those legacy features that are grandfathered in - I'd never accept them if they were proposed today, but there's no reason to drop it if it doesn't get in the way.

JaydenLiang: true. I agree. There's no other reason to make change to this legacy feature then not do it :)
Thank you for your patience to explain!

narfbg: You're welcome, but ... this is a code review and you're neither making corrections to the patch, nor withdrawing it. What do you want to do with it?

JaydenLiang: I close it now. I just wanted to wait for another day to see if any incoming comment.

narfbg: Not the answer I was expecting, I guess I have to patch it myself.

JaydenLiang: Oh, I get it now so l am glad to patch it for you, in a new pr, with exactly the same code as you mentioned here.

JaydenLiang: NVM, just saw that you patched it already.

PR #5328, opened 2017-11-07
Use PSR-1 for PHPUnit TestCase

gabriel-caruso:
Use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/tests/codeigniter/Setup_test.php
+++ b/tests/codeigniter/Setup_test.php
@@ -1,6 +1,8 @@
 <?php
 
-class Setup_test extends PHPUnit_Framework_TestCase {
+use PHPUnit\Framework\TestCase;
+
+class Setup_test extends TestCase {
 
 	public function test_bootstrap_constants()
 	{
--- a/tests/mocks/ci_testcase.php
+++ b/tests/mocks/ci_testcase.php
@@ -1,6 +1,8 @@
 <?php
 
-class CI_TestCase extends PHPUnit_Framework_TestCase {
+use PHPUnit\Framework\TestCase;
+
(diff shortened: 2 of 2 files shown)


narfbg:
Have you tested this with PHPUnit 6? I highly doubt that these 2 lines are the only changes we'd need to make it work. Your brief PR description is the only place where PHPUnit 6 is mentioned and that really makes it seem you're suggesting this only for the sake of PSR-compliance.

gabriel-caruso:
@narfbg These are not the only changes that PHPUnit 6 has. Only change the `namespaces` are the first steps. PHPUnit has deprecated methods, requires `PHP 7`, etc. If you want, I can work on that in master branch :smile:

narfbg:
Just so you know, we only merge in `master` after releases. For anything else - target `develop`, or `3.1-stable` if the patch only applies to 3.1.x.

But that aside, this still looks like just changes for the sake of PSR compliance. If PHPUnit 6 requires PHP 7 and we don't ... why would we want to work towards PHPUnit 6?
I mean, if the changes don't break PHPUnit 5 - sure, maybe give v6 support to whoever wants it, but it a) is that the case?; and b) if so, make it work entirely.

gabriel-caruso:
@narfbg I'll change to `develop` branch, and work on that migration in another PR :sweat_smile:

PR #5330, opened 2017-11-10
Update Model.php

pgee70:
this is a lot more useful to see which model has loaded when analysing log files.

### DIFF
--- a/system/core/Model.php
+++ b/system/core/Model.php
@@ -55,7 +55,7 @@ class CI_Model {
 	 */
 	public function __construct()
 	{
-		log_message('info', 'Model Class Initialized');
+		log_message('info','Model class:'.get_class($this). ' initialised');
 	}
 
 	// --------------------------------------------------------------------


narfbg:
Not a bad idea in principle, but your patch changes the semantics of what is supposed to be logged there.

That message is supposed to tell you that `CI_Model` was loaded, which only happens once. Incidentally that also means it's in the wrong place, but so is yours as a lot of developers would override the constructor in their own models. That is, both the old message and yours need to be moved into `CI_Loader::model()` (or one of the methods it calls, if any).

In addition, please follow our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html) and be consistent. This is a one-line patch that's supposed to add a single word and yet you've managed to do this:

    // Old/current version:
    log_message('info', 'Model Class Initialized');

    // Your proposed version:
    //
    // - Removed space after the comma
    // - No space before concatenation, yet there's a space after it (there should be none)
    // - Would result in a message lacking space between the colon and the class name
    // - Typo in "initialised", which already existed as "Initialized" (s/z)
    log_message('info','Model class:'.get_class($this). ' initialised');

pgee70:
I am from Australia, so we use English spelling here. ;-)  or we spell correctly. ;-)
I started a new project today and saw my overrides and thought why not make a PR?
if you could add it to the loader it might seems a better fit.  I found this to be useful years ago before I had decent xdebug integration. Not so much of an issue now.

PR #5334, opened 2017-11-19
Add Nginx server setting guide for CI Urls. - Update urls.rst

yidas:
Add Nginx server setting guide for CI Urls.


narfbg:
This example is very questionable, with no 404 fallback and will fail under a lot of conditions due to how nginx works. I for one use a completely different nginx ruleset and what you suggest would break my apps.
It's not Apache, you can't just shove a few lines into its config and expect the thing to magically work.

Also removing a note for some reason ... Shouldn't be doing that.

PR #5341, opened 2017-11-24
PHP: bump to PHP 5.6

TomasVotruba:
### Why?

- **[PHP 5.4 and 5.5 may contain unfixed security issues](http://php.net/supported-versions.php)** as they're not supported for over ye year now

- majority of PHP [world bumping to PHP 7.1](https://gophp71.org/)
- upcoming [PHP 7.2](https://github.com/php-ai/php-ml/pull/147) in few days
- we could add PHP 7.0/7.1 typehints via coding standards] more easily in the future
- [as measured on Packagist - 67 % people](https://seld.be/notes/php-versions-stats-2017-2-edition) are on PHP 7.0+ and growing

What do you think?


gxgpet:
Isn't this a duplicate of #5227?

TomasVotruba:
@gxgpet I plan to add more, since some months passed and reasons were added.

gxgpet:
The issues you are saying about are caused by old PHP versions, not the CodeIgniter framework itself... The solution here, if security is a concern, is to update the PHP version, in my opinion.

Anyway, you can wait for some official feedback from one of the project members. :)

TomasVotruba:
That's technically right, but people don't know/care about security issues. Frameworks should take care of that and not to promote insecure practise like this one.

narfbg:
> - **[PHP 5.4 and 5.5 may contain unfixed security issues](http://php.net/supported-versions.php)** as they're not supported for over ye year now

I am very aware of this. If you look at the README file, you'll know that we always recommend at least PHP `<put earliest version still supported here>`.

> - majority of PHP [world bumping to PHP 7.1](https://gophp71.org/)

Go PHP 7.1 is a nice initiative, but you're using it to make a very deceptive statement and this annoys me quite a bit. *Some* major projects are planning or releasing new versions with PHP 7.1 as a minimum requirement, but that is not a majority of any kind, nothing to do with CodeIgniter, and you're suggesting 5.6 anyway.

[W3Techs.com still shows PHP 5 market share at 89% with 67% of that being below 5.6](https://w3techs.com/technologies/details/pl-php/5/all).

> - upcoming [PHP 7.2](https://github.com/php-ai/php-ml/pull/147) in few days

Also very well aware of this. In fact, I've contributed some bits of code to it.
But nothing to do with CodeIgniter's minimum supported PHP version (and you could've linked to some of the 7.2RC announcements instead of a random PR of yours).

> - we could add PHP 7.0/7.1 typehints via coding standards] more easily in the future

I don't know what "typehints via coding standards" means, but it doesn't sound particularly appealing to me and is contradictive with (again) this PR targeting PHP 5.6 instead of 7.

If you meant using newer PHP features - sure, that'd be nice, but it's not something that we hear or think about for the first time.

> - [as measured on Packagist - 67 % people](https://seld.be/notes/php-versions-stats-2017-2-edition) are on PHP 7.0+ and growing

The cheap counter-point to this is that Packagist is not the entire PHP eco-system (plus the above-referenced statistic). But the actually important one is that a LOT of CodeIgniter's userbase doesn't use Composer, and a significant part of it even uses CI precisely because it doesn't force them to use Composer.

> What do you think?

I think this looks like an excuse to re-submit #5227, and let me be clear about that - I will not merge any PR changing literally every file to use the new square brackets syntax for arrays! Because:

- `array()` works just fine and there's no reason to change it.
- It's a huge amount of code being changed, which means it's a nightmare to review the patch. If it ever happens, I or another member of our team would do it.
- Our 3.1-stable branch still technically runs PHP 5.3 and we need to be able to merge/cherry-pick commits between it and develop.

I'm not saying I don't like or want the syntax, but it can't happen now and any attempt to couple it with other proposals will 100% be the downfall of said proposals.

As it stands, the PHPUnit-related changes are the only thing I'd merge from this PR right now.

---

On the bump to 5.6 itself ... I could've added this above, but it wouldn't do it justice as I ended with talk about the square brackets array syntax.

First, a reminder that Composer itself only requires PHP 5.3 - any reasoning that they can give for this, you can count on that even stronger for CodeIgniter as it is a way more conservative project. Not to mention that CI is also way older and running on lots of legacy systems, and delivering *our* security patches is hard enough.

Personally, I think having a framework that runs on everything between 5.3 and 7.2 (as we currently do) is a good thing. Requiring PHP 7.x means that new and actively supported code doesn't run outdated PHP versions, but it doesn't do any favors to vast amounts of code out there that's been running for years without change.  
At some point, all of these legacy systems will need upgrades for one reason or another, but unless that is made as easy as possible, it will be delayed until they literally break down. Having a framework unaffected by PHP version changes does make it easy to upgrade - a silver lining, if you will. :)

That being said, you're of course targeting the develop branch, which would become CodeIgniter 3.2.0. And merges between the 3.1-stable and develop branches aside, it is highly likely that we will bump the bare minimum requirement.  
We did drop 5.3 from develop shortly after 3.1-stable was branched-out - that was a no-brainer, but progress on 3.2.0-dev hasn't been particularly fast and so far we haven't actually needed newer PHP features in it, and thus a higher requirement hasn't happened yet.

Mind you, there's a difference between what's supported and what the bare minimum requirement to run the framework is. We don't change code just to break it for a portion of environments (like you're doing in this PR), we change it when necessary for something.

Want a new feature that *has* to utilize newer PHP? Or to refactor something that becomes better if it uses a newer PHP feature? As long as it's not just syntax suguar, but a notable performance optimization, or drops 50 lines of code - absolutely go for it. Go crazy dropping compat code after something like that is merged - it wouldn't be a question then.

TomasVotruba:
tl;dr; I understand you're not opened to change this, since you've closed this without any discussion.

Thanks for your time though. Maybe in the future

narfbg:
Man, I wrote a 2-page reply and you immediately understand I am not open to change, *while* saying you didn't even read it?

There's thousands of reasons to close a PR and that doesn't prevent discussion. Do me a favor by reading it please.

TomasVotruba:
I was talking about change, not a discussion.

I don't want to force you agains your values. **I'm trying to help open-source projects that want this yet don't have energy and time to do this**. This was just proof of concept. It's not problem for me.

narfbg:
Well, you said I closed it without any discussion and that I'm not open to "this" change. I gave a pretty lengthy explanation why I closed it, along with pointing out how/where the change should start. You just didn't read it before replying.

Otherwise, I get where you're coming from and I do appreciate the effort, it's a good initiative.

Paradinight:
@TomasVotruba 
if you have time. go to https://github.com/bcit-ci/CodeIgniter4/issues and help there.

TomasVotruba:
@Paradinight Thanks a lot, that's new way for tagging for me.

PR #5347, opened 2017-12-02
Added paragraph helper

vinodselvin:
Added paragraph helper to HTML_helper, and added user_guide for paragraph tag and how to use it.


narfbg:
Thanks, but no. It literally takes less time to write the raw HTML instead of using such a function.

PR #5354, opened 2017-12-08
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertInternalType` instead of `is_*` functions;
- `assertInstanceOf` instead of `instanceof` operator;
- `assertNull` instead of comparison with `null` keyword;
- `assertFalse` and `assertNotFalse` instead of comparison with `false` keyword;
- `assertTrue` instead of comparison with `true` keyword;
- `assertSame` and `assertSame` instead of strict comparisons `===` and `!==`;
- `assertEquals` and `assertNotEquals` instead of comparisons `==` and `!=`;
- `assertContains` instead of `in_array` and `strpos` functions;
- `assertArrayNotHasKey` instead of `array_key_exists` function.


narfbg:
Some of those make sense, like `assertTrue()` instead of `assertEquals(TRUE, <expr>)`, but others are questionable - `assertContains()` replacing both string and array operations is ambiguous; not everybody is familiar with how `assertSame()` is different than `assertEquals()`; etc.

Either way, there's nothing inherently wrong in the current tests code, so how does this improve them? If anything, I see potential for them just being one tiny bit slower and harder to understand at times.

carusogabriel:
I'm doing this kind of refactoring for three major reasons:

1. Better message when test fails, helping us debug it;
2. Less complexity in tests, avoiding using PHP function on it;
3. Let PHPUnit handles comparisons, searches, and stuff better than us.

ccsalway:
Apart from the ambiguous one's, they get my vote.

narfbg:
OK ...

1. Can you give some examples on those better messages?
2. I don't see any reduced complexity, and avoidance of native PHP is not a positive of any kind.
3. Better how? Those messages you mention in p1 - maybe, better readability in *some* cases - sure, but I fail to see how it can possibly make comparisons and searches better.

I mean, 2 out of those 3 points don't make sense to me and in general you're just saying it's "better" ... That doesn't explain anything.

carusogabriel:
### 1º Better messages (grouped 2 by 2)
Assertion | Error Message
-- | -
`assertEquals(0, count(array('one')))` | Failed asserting that 0 matches expected 1
`assertCount(0, array('one'))` | Failed asserting that actual size 1 matches expected size 0
`assertTrue(in_array('foo_bar', ['bar', 'foo']))` | Failed asserting that false is true
`assertContains('foo_bar', ['bar', 'foo'))` | Failed asserting that an array contains 'foo_bar'
`assertTrue(1 === 2)` | Failed asserting that false is true
`assertSame(1, 2)` | Failed asserting that 2 is identical to 1

### 2º and 3º Avoid using PHP functions
While checking for types, for example, with `is_*`, besides the error messages, [`assertInternalType`](https://github.com/sebastianbergmann/phpunit/blob/4.8/src/Framework/Constraint/IsType.php) handles it so much better.
[`assertCount`](https://github.com/sebastianbergmann/phpunit/blob/4.8/src/Framework/Constraint/Count.php)  as well, always checking the best way to count.

narfbg:
Right ... Sorry for the delay, but I'm in holiday mode already.

I can see how some of the messages are better, although the list looks a bit confusing. Still, some of the things you're replacing wouldn't benefit even from that ... like this for example:

    -		$this->assertTrue(preg_match('#^[0-9a-f]{32}$#iS', $this->security->xss_hash) === 1);
    +		$this->assertSame(1, preg_match('#^[0-9a-f]{32}$#iS', $this->security->xss_hash));

A useful message here would be something like "X doesn't match regular expression Y". If we don't get that, it doesn't in any way matter if we use `assertTrue()` on an expression, or `assertEquals()`, `assertSame()`, something else on the return value - these are all just short ways to make a basic comparison in which the compared value represents the actual test result (thus the message would be meaningless because assertions only compare "truthiness" in those cases).

And yea ... just give up on 2 and 3; you're just repeating that PHPUnit does it better and there's no objective criteria for "better" or "best". If anything, PHPUnit adds a bit of overhead, so it can't possibly do better aside from giving friendlier messages.
And that's fine - friendlier error messages are an improvement. Just drop the replacements similar to the one example I gave above and I'll merge it. Tell me if you need me to point out which ones.

carusogabriel:
@narfbg No problem about your late man, the holidays are really close.

Please, point me what refactories you think are superfluous. I'll rollback them.

About your comment, PHPUnit has an [`assertRegExp`](https://phpunit.de/manual/4.8/pt_br/appendixes.assertions.html#appendixes.assertions.assertRegExp). Should we use it?

narfbg:
Ok, I'll leave a few inline comments. And if it does give a meaninful message - sure, we can use `assertRegExp()`.

tests/codeigniter/libraries/Form_validation_test.php
```diff
@@ -374,8 +374,8 @@ public function test_set_error_delimiters()
 		$this->form_validation->run();
 		$error_msg = $this->form_validation->error('foo');
 
-		$this->assertTrue(strrpos($error_msg, $prefix) === 0);
```
narfbg: This and the next line (aside from the regexp) don't look as improvements to me.

The bigger issue is that no matter how you write them, they'll always look confusing even as code (not just as far as error messages go). Kudos if you find a way around that.

carusogabriel: `strrpos === 0` sounds like `assertStringStartsWith`. Gonna test it.

carusogabriel:
Amend the commit and made the changes :+1:

narfbg:
:+1: 

But please just push new commits in the future - easier to keep track of.

PR #5357, opened 2017-12-12
The Second Time Get Config Fail  Becose Of Params  use_sections

jeary:
The Second Time Get Config Fail  Becose Of Params  use_sections


narfbg:
It is not clear what you mean, please elaborate.

And also, please don't use capital letters to start every word. :)

Issue #4393, opened 2016-01-21
base_url bug in 3.0.4?


hldgaofeng:
base_url bug in 3.0.4?
try to access:
http://192.168.0.5:8080/controller_1/method_1/

```php
class Controller_1 {
  public function method_1() {
    echo base_url('mypath'); // will got wrong path: http://192.168.0.5/mypath
 }
}
```

in system/core/Config.php
```php
    public function __construct()
    {   
        $this->config =& get_config();

        // Set the base_url automatically if none was provided
        if (empty($this->config['base_url']))
        {   
            if (isset($_SERVER['HTTP_HOST'])) // HTTP_HOST will taken :port 
            {   
                if (strpos($_SERVER['HTTP_HOST'], ':') !== FALSE)
                {   
                    $server_addr = $_SERVER['HTTP_HOST'];
                    //$server_addr = '['.$_SERVER['HTTP_HOST'].']'; // what's this mean??
                }   
                else
                {   
                    $server_addr = $_SERVER['HTTP_HOST'];
                }   

                $base_url = (is_https() ? 'https' : 'http').'://'.$server_addr
                    .substr($_SERVER['SCRIPT_NAME'], 0, strpos($_SERVER['SCRIPT_NAME'], basename($_SERVER['SCRIPT_FILENAME'])));
            }   
            else
            {   
                $base_url = 'http://localhost/';
            }   

            $this->set_item('base_url', $base_url);
        }   

        log_message('info', 'Config Class Initialized');
    }   
```

narfbg:
The port thing is not a bug: https://github.com/bcit-ci/CodeIgniter/pull/4217#issuecomment-154220634
And HTTP_HOST is not safe to use either.

PR #5359, opened 2017-12-12
Set port on base_url if not set in config

ccsalway:
On OSX, port 80 cannot be used (without special permission) when testing locally and if base_url is not set in config (in favour of using the Config _constructor to set the base_url), then it should be more clever.

### DIFF
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -99,13 +99,16 @@ class CI_Config {
 					$server_addr = $_SERVER['SERVER_ADDR'];
 				}
 
-				$base_url = (is_https() ? 'https' : 'http').'://'.$server_addr
-					.substr($_SERVER['SCRIPT_NAME'], 0, strpos($_SERVER['SCRIPT_NAME'], basename($_SERVER['SCRIPT_FILENAME'])));
-			}
-			else
-			{
-				$base_url = 'http://localhost/';
-			}
+                $base_url = (is_https() ? 'https' : 'http').'://'.$server_addr
+                    . ($_SERVER['SERVER_PORT'] == (is_https() ? '443' : '80') ? '' : ':' . $_SERVER['SERVER_PORT'])
+                    .substr($_SERVER['SCRIPT_NAME'], 0, strpos($_SERVER['SCRIPT_NAME'], basename($_SERVER['SCRIPT_FILENAME'])));
+            }
+            else
+            {
+                $base_url = (is_https() ? 'https' : 'http') . '://localhost'
+                    . ($_SERVER['SERVER_PORT'] == (is_https() ? '443' : '80') ? '' : ':' . $_SERVER['SERVER_PORT'])
+                    . '/';
+            }
 
 			$this->set_item('base_url', $base_url);
 		}


narfbg:
Duplicate of #4393.

FYI: All ports under 1024 are allocated for system processes and this is not limited to OSX, but that doesn't mean you can't use them.

ccsalway:
I know that. Saying "without special permission" meant that you just have to sudo the web server when it starts..  But that isn't what this pull request is specifically about.

ccsalway:
I appreciate it's a duplicate but mine is a better solution.  Plus I see you have closed both the tickets so does that mean it's an ignored issue?

ccsalway:
The PR description was an explanation of why I recorded it as an issue.  Rather than just posting the issue and letting you guess why it's an issue.

narfbg:
> I appreciate it's a duplicate but mine is a better solution.

Yours is better, indeed. But we could've just improved the previously suggested patch - if there ever was a perfectly OK patch for anything, I don't remember it; part of our role here is to insist that PRs are brought up to our standards before merging them, not just straight accept or reject them. The quality is not the point here though ...

> Plus I see you have closed both the tickets so does that mean it's an ignored issue?

It's not an "ignored issue", it's a rejected feature request.

> The PR description was an explanation of why I recorded it as an issue. Rather than just posting the issue and letting you guess why it's an issue.

I know why it's an issue and that's not it ... there's plenty of other reasons why one might not want or be able to use the standard port numbers. The issue is that the port, whichever it is, is not autodetected like some people would like.  
However, it is clearly stated in the config file itself that you are REQUIRED to manually set your 'base_url' and must not ever rely on auto-detection.

ccsalway:
> However, it is clearly stated in the config file itself that you are REQUIRED to manually set your 'base_url' and must not ever rely on auto-detection.

I see that but this is also not ideal.  Imagine you develop locally so set the `base_url` to `http://localhost:8080` and then upload to production. You will first have to change the `base_url` in the config before you commit to Git... and then back to `http://localhost:8080` to continue developing, then adding the config file to `.gitignore` so it's not accidentally uploaded again which would break the links....

My method auto detects the URL and port no matter the environment.

gxgpet:
@ccsalway So the solution for your issue is actually [this](https://www.codeigniter.com/user_guide/libraries/config.html#environments)? :)

ccsalway:
Yep, exactly :D

narfbg:
`localhost` usage is for quick and dirty tests, and nothing else. Even on your own machine you should setup a meaningful hostname (`projectname.local`, for example) and put that in the `base_url` ... the only reason why it defaults to `localhost` is that there's nothing else to default to.

Auto-detection can never be reliable, and even yours doesn't do it "no matter the environment" - it works on your local machine, not on any other server. This thing about uploading to a production server that you talk about, won't work without you modifying the framework (and if you do that, you've likely opened up your app(s) to `Host` header injections).
There's built-in per-environment configurations to solve that problem.

It may be inconvenient at first, but that's just how it is, and we've been through this a lot of times ... What we currently have is the product of lots of painful experience.

ccsalway:
Thanks for taking the time to give some meaningful response.  Appreciate it.

narfbg:
You're welcome. :)

plonknimbuzz:
@ccsalway mac is not the only one who facing this issue, 
im windows and linux user
Since CI 3 (if im not wrong), CI use IPv6 as base_url . (i dont know what exactly happen)
then my localhost becomes ::1

if i write 127.0.0.1 or localhost instead ::1 
that's really odd for me (old habit), but if i force to use localhost/127.0.0.1 all file will not loaded because cross domain rules.

and then i must change base_url, every time i upload to my VPS or download it again.

then i realize that ENVIRONMENT is the way to solve this.

PR #5366, opened 2017-12-19
Update security.rst

plonknimbuzz:
update old method php echo, to prevent deprecated possibility in the future.

### DIFF
--- a/user_guide_src/source/libraries/security.rst
+++ b/user_guide_src/source/libraries/security.rst
@@ -65,7 +65,7 @@ and ``get_csrf_hash()``
 
 	...
 
-	<input type="hidden" name="<?=$csrf['name'];?>" value="<?=$csrf['hash'];?>" />
+	<input type="hidden" name="<?php echo $csrf['name'];?>" value="<?php echo $csrf['hash'];?>" />
 
 Tokens may be either regenerated on every submission (default) or
 kept the same throughout the life of the CSRF cookie. The default


Paradinight:
Source? If not do not change it.

plonknimbuzz:
what source do you mean?

Paradinight:
Where did you get the information from?

plonknimbuzz:
from here:
https://www.codeigniter.com/userguide3/libraries/security.html#cross-site-request-forgery-csrf

i search on development branch -> https://github.com/bcit-ci/CodeIgniter/blob/develop/user_guide_src/source/libraries/security.rst#cross-site-request-forgery-csrf and then i edit it

Paradinight:
About <?= and not csrf

plonknimbuzz:
owh, i see
i write deprecated possibility because : 
- use short open tags is bad practice
- php docs, or any article always warn about this
- 5.4 disabled it by default



Version | Description
-- | --
7.0.0 | The ASP tags <%, %>,           <%=, and the script tag           <script language="php"> are removed from PHP.
5.4.0 | The tag <?= is always available regardless of the short_open_tag ini setting.

so there's a chance/possibility to deprecated next time (my subjective).
- http://php.net/manual/en/language.basic-syntax.phptags.php
- http://php.net/manual/en/ini.core.php#ini.short-open-tag

anyway, this is just suggestion minor fix to the documents, it's fine if this PR rejected.
i just not create any issue to discuss this before, because this PR is totally trash (not mean anything) to discuss, haha. 
And i thought most of us will agree about this short open tags

https://softwareengineering.stackexchange.com/a/151694
`... so you'll find a million and a half recommendations to turn short_open_tag off, which you should ...`

Paradinight:
<?= is always available and is not affected by short_open_tag config. 

> 5.4.0 The tag <?= is always available regardless of the short_open_tag ini setting.

plonknimbuzz:
what? i just know that. omg
i so shame right now. 

i think we should delete this PR then.
sorry for notify 1,7k people, because i think this will accepted easly.

next time, i will discuss on the commit stage first.
thanks @Paradinight

narfbg:
We can't delete PRs and issues; they can only be closed.

Just do your due diligence and be 100% sure about what you're doing before opening one.

PR #5367, opened 2017-12-20
Clean elses

carusogabriel:
I've cleaned the `else`s when we have already returned something, this saves us some lines of code :put_litter_in_its_place:

[`PHP-CS-Fixes`](https://github.com/FriendsOfPHP/PHP-CS-Fixer) helped me find them.

### DIFF
--- a/system/database/DB_result.php
+++ b/system/database/DB_result.php
@@ -163,10 +163,8 @@ class CI_DB_result {
 		{
 			return $this->result_object();
 		}
-		else
-		{
-			return $this->custom_result_object($type);
-		}
+
+		return $this->custom_result_object($type);
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


system/database/DB_result.php
```diff
@@ -336,7 +334,7 @@ public function row($n = 0, $type = 'object')
 
 		if ($type === 'object') return $this->row_object($n);
 		elseif ($type === 'array') return $this->row_array($n);
-		else return $this->custom_row_object($n, $type);
+		return $this->custom_row_object($n, $type);
```
narfbg: There needs to be an empty line before this one, to visually split it from the `if ... elseif`.

Everything else looks fine. :+1:

carusogabriel: Did it. Let's wait for the tests :+1:

PR #5369, opened 2017-12-25
Cache_memcached: setting initial value with increment/decrement when not exist

tianhe1986:
Just trying to keep consist with other drivers. I have tested `file`, `redis` and `apcu`, they would set a initial value when the item do not exist.

`Memcache` does not create an item if not exist refer [here ](https://secure.php.net/manual/en/memcache.increment.php#refsect1-memcache.increment-description).
`Memcached` would create an item, but only with binary protocol refer [here](https://secure.php.net/manual/en/memcached.increment.php#111187).I also tried it myself.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_memcached.php
+++ b/system/libraries/Cache/drivers/Cache_memcached.php
@@ -210,6 +210,7 @@ class CI_Cache_memcached extends CI_Driver {
 	 */
 	public function increment($id, $offset = 1)
 	{
+		$this->_memcached->add($id, 0);
 		return $this->_memcached->increment($id, $offset);
 	}
 
@@ -224,6 +225,7 @@ class CI_Cache_memcached extends CI_Driver {
 	 */
 	public function decrement($id, $offset = 1)
 	{
+		$this->_memcached->add($id, 0);
 		return $this->_memcached->decrement($id, $offset);
 	}
 
--- a/system/libraries/Cache/drivers/Cache_memcached.php
+++ b/system/libraries/Cache/drivers/Cache_memcached.php
@@ -210,8 +210,13 @@ class CI_Cache_memcached extends CI_Driver {
 	 */
 	public function increment($id, $offset = 1)
 	{
-		$this->_memcached->add($id, 0);
-		return $this->_memcached->increment($id, $offset);
+		if (($result = $this->_memcached->increment($id, $offset)) === FALSE)
+		{
+			$this->_memcached->add($id, 0);
+			$result = $this->_memcached->increment($id, $offset);
+		}
+
+		return $result;
 	}
 
 	// ------------------------------------------------------------------------
@@ -225,8 +230,13 @@ class CI_Cache_memcached extends CI_Driver {
 	 */
 	public function decrement($id, $offset = 1)
 	{
-		$this->_memcached->add($id, 0);
-		return $this->_memcached->decrement($id, $offset);
+		if (($result = $this->_memcached->decrement($id, $offset)) === FALSE)
+		{
+			$this->_memcached->add($id, 0);
+			$result = $this->_memcached->decrement($id, $offset);
+		}
+
+		return $result;
 	}
 
 	// ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


narfbg:
An argument could be made that increment/decrement on a non-existent value shouldn't be attempted in the first place ...

But even if we ignore that, I'm not happy with always calling `add()` only for it to fail in the 99% case.

tianhe1986:
@narfbg  So how about calling `add()` after increment/decrement if the return value is `FALSE` ?

narfbg:
Sounds better, yes.

narfbg:
Better, but we don't need that `$result` variable (just return as soon as possible) and there's no point in adding 0 only to later increment. Way simpler:

		if ($this->_memcached->increment($id, $offset) === FALSE)
		{
			return $this->_memcached->add($id, $offset) ? $offset : FALSE;
		}

		return TRUE;

tianhe1986:
1. In your above code, `return TRUE;`  is not right. If the return value of  `$this->_memcached->increment($id, $offset)` is not `FALSE`, then it should be returned.  The `$result` variable is used mainly for this purpose.

2. I used later increment, because many requests may arrive at the same time. In this case, only one request would succeed with `add($id, $offset)`,  the others returning `FALSE `. But with later increment, all of them would succeed with `increment($id, $offset)`. 
However, in low-traffic application, `add($id, $offset)` would surely reduce the access times to memcache server. So it is hard to say which one is better:)

narfbg:
Hmm, you're right about the first point, but at the very least you could directly return the second `increment()` call.

I'm not at all convinced that the second (concurrency) would be a problem though, and to be honest - I'd rather have all other drivers behave like Memcache does by default (return FALSE if the key doesn't exist). I don't see it as a problem that an operation fails when it shouldn't be attempted in the first place.

tianhe1986:
1. OK. I'll return the second call immediately.
2. Since it is memcache driver,  how about the following code according to the advise from the [official document](https://github.com/memcached/memcached/wiki/Programming#proper-use-of-incr-or-decr)? It seems better with both low/high concurrency :) 
```php
                if (($result = $this->_memcached->increment($id, $offset)) === FALSE)
		{
			return $this->_memcached->add($id, $offset) ? $offset : $this->_memcached->increment($id, $offset);
		}

		return $result;
```

But, with `decrement`  , just `add($id, 0)`,  because a negative number would cause later increment/decrement not working refer [here](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L277).

narfbg:
Sounds good, yes.

system/libraries/Cache/drivers/Cache_memcached.php
```diff
@@ -224,7 +229,12 @@ public function increment($id, $offset = 1)
 	 */
 	public function decrement($id, $offset = 1)
 	{
-		return $this->_memcached->decrement($id, $offset);
+		if (($result = $this->_memcached->decrement($id, $offset)) === FALSE)
+		{
+			return $this->_memcached->add($id, 0) ? 0 : $this->_memcached->decrement($id, $offset);
```
narfbg: I thought you'd just return 0 here? How would `decrement()` work if `add()` fails?

tianhe1986: If `add()` fails in this request, then `add()` from other requests may succeed and `decrement()`  in this request could work.

narfbg: If anything, I'd want to prevent concurrent requests messing with each other, not take extra steps to allow it ...

tianhe1986: Concurrent requests would not mess with each other,  every one of them could succeed only once with add/decrement/increment calling.

narfbg: I don't know what kind of logic you're going with in that statement. If `increment()` in request X succeeds only because `add()` succeeded in request Y, then they are clearly interfering with each other.

tianhe1986: Hmm, you are right, they surely interfere with each other.
However,  the result is the same as
1. request Y succeed with `add() `.
2. After request Y finishes, request X come in and succeed with `increment()`.

I just made some more retry, and as I showed above, the example code in memcache official document is also making more retry. 

I understand what you said,  but  still, it is hard to say retry is right or wrong :)

narfbg: It doesn't have to be right or wrong.

Isn't it consistency that's what you're aiming for in this PR? We don't retry anything else, so there's that.

tianhe1986: OK， I see.

PR #5371, opened 2017-12-27
fix docs (necessary when extending CI_Form_validation)

kasperpeulen:


narfbg:
I don't know what you mean by it being necessary. It's not necessary for anything.

kasperpeulen:
If you extend CI_Form_validation in MY_Form_validation, the type will be MY_Form_validation and not CI_Form_validation like it is now. So the documentation will be incorrect in that case, and IDEs that parse phpdocs will have the wrong type information.

So it is necessary if you care about correct documentation and good IDE integration.

narfbg:
Nope. An instance of `MY_Form_validation` would still match the `CI_Form_validation` type because it would be the parent class. You can easily test this with `instanceOf`, an argument type declaration, etc.

gxgpet:
I think that he's trying to say that if you create an extended `MY_Form_validation`, then you _can't_ make chain-calls to the custom methods added if you call any of `CI_Form_validation` methods in the first place since the `@return` docblock is misleading for the IDE. (wrote _can't_ as italic word because this is limiting only the IDE, not the code itself)

kasperpeulen:
@gxgpet That is true.

> Nope. An instance of MY_Form_validation would still match the CI_Form_validation type because it would be the parent class. You can easily test this with instanceOf, an argument type declaration, etc.

But this is about the other direction. According to the docs, the method set_rules returns an object of type CI_Form_validation. In reality, it is of type MY_Form_validation, and CI_Form_validation doesn't match MY_Form_validation. You can easily test this with instanceOf. 

For example, if I would add the method "validate" to `MY_Form_validation`, this method is not available to instances of `CI_Form_validation`.

image

narfbg:
> According to the docs, the method set_rules returns an object of type CI_Form_validation. At runtime, it is of type MY_Form_validation, and CI_Form_validation doesn't match MY_Form_validation. You can test this with instanceOf.

Yes it *does* match:

    $ php -a
    Interactive mode enabled
    
    php > class Foo {}
    php > class Bar extends Foo {}
    php > $bar = new Bar();
    php > echo ($bar instanceOf Foo) ? "variable matches type Foo\n" : "variable does NOT match type Foo\n";
    variable matches type Foo
    php > 

It matches *both* types and this has nothing to do with runtimes or anything, it's how inheritance works. If anything, an IDE doesn't recognize this because it doesn't actually have a runtime representation of the object and only blindly trusts phpdoc annotations. Even then, the way you say it behaves is as if we're talking about a static class, so that's a problem of the IDE itself in my book.

gxgpet:
I'm not sure if all IDEs can actually interpret what happens inside a method, runtime strictly speaking, for high complexity code.

I think we could consider changing the docblock from `CI_Form_validation` to `$this`, so it would fine for all scenarios :P

narfbg:
I'm not saying that they all can, it's not an easy thing to do. But it's certainly not our problem that an IDE can't do something and it's not our job to take care of IDEs (especially given that we do this for free while they get profits).

In case you were serious with that suggestion, `$this` is even more silly than `self`. It's either `CI_Form_validation` or `object` - that's all we've ever used, and it's always been for documentation purposes only.

PR #5372, opened 2017-12-31
Clean up

carusogabriel:
I've removed all white spaces and extra lines from our documentation :smile:


narfbg:
I'd happily merge a clean-up for the odd trailing space (after a semicolon), but most of these are intentional - empty lines for readability and tabs/spaces on seemingly empty lines to join code in a single block.

In some cases, these may even be functionally relevant as the docs parsing syntax is a bit weird.

PR #5373, opened 2018-01-03
prevent _ci_vars from being cached in _ci_cached_vars

TalonTR:
Signed-off-by: Christian Mohr <christian.mohr@insitu.de>

I propose to prevent variables from the second `$this->load->view()` parameter `$vars` to get stored within the `$this->_ci_cached_vars` cache. This cache should only be filled from a corresponding `$this->load->vars()` call and respectively cleared via `$this->load->clear_vars()`if desired. This would be the behaviour expressed within the [documentation](https://www.codeigniter.com/user_guide/libraries/loader.html?highlight=view#CI_Loader::view).

**Background**: The documentation states, that global view parameters could be set via `vars()` and afterwards used within all views without further requirements. This store can be cleared via `clear_vars()` if necessary. The `CI_Loader::view()` method provides an alternative approach for providing a single view with dynamic data. Neither does the documentation state that this data gets stored internally and send to all subsequent view calls, whether nested subviews or completely different ones. Nor is this an intuitive way to use the `view()` function since it suggests that the `$vars` parameter is used for this call only.

**Issue**: As soon as one uses `isset($variableName)` or similar within a view file, the behaviour depends on which `view()` calls with which `$vars` preceeded the current one. It should depend ONLY on which `$vars` where provided to the current call and also what was set beforehand via an explicit `vars()` call.

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -94,6 +94,13 @@ class CI_Loader {
 	protected $_ci_cached_vars =	array();
 
 	/**
+	 * stack of variables for nested _ci_load calls
+	 *
+	 * @var	array
+	 */
+	protected $_ci_vars_stack =	array();
+
+	/**
 	 * List of loaded classes
 	 *
 	 * @var	array
@@ -934,14 +941,30 @@ class CI_Loader {
 		}
 
 		/*
-		 * Extract and cache variables
+		 * Extract and stack variables
 		 *
 		 * You can either set variables using the dedicated $this->load->vars()
 		 * function or via the second parameter of this function. We'll merge
-		 * the two types.
+		 * the two types. Additionally we merge them with the last variables
+		 * from the nested call stack so that views that are embedded within
+		 * other views can have access to these variables.
 		 */
-		empty($this->_ci_cached_vars) OR $_ci_vars = array_merge($this->_ci_cached_vars, $_ci_vars);
-		extract($_ci_vars);
+
+		// merge with global cached vars (first call) or last state from nested
+		// call stack (subsequent nested calls)
+		// if (!empty($this->_ci_vars_stack)) {
+		// 	$_ci_vars = array_merge(end($this->_ci_vars_stack), $_ci_vars);
+		// } else if (!empty($this->_ci_cached_vars)) {
+		// 	// merge with cached vars
+		// 	$_ci_vars = array_merge($this->_ci_cached_vars, $_ci_vars);
+		// }
+
+		empty($_ci_vars) OR $this->_ci_cached_vars = array_merge($this->_ci_cached_vars, $_ci_vars);
+
+
+		// push current _ci_vars state to stack and extract it
+		// array_push($this->_ci_vars_stack, $_ci_vars);
+		extract($this->_ci_cached_vars);
 
 		/**
 		 * Buffer the output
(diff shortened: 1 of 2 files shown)


narfbg:
The documentation may not explicitly state this, and I agree that it's not very intuitive to an experienced programmer, but there's a reason for it and your patch just erases it from the inline comment blocks ...

This would be a major BC break and would make a lot of people unhappy. You'll have to be much, much more convincing than that.

TalonTR:
You´re totally right and i understand the reason for this. I'd find it totally acceptable if nested view calls share the `$vars` from the initial view call. But it still seams like wrong behaviour to me that subsequent, non-nested calls share an ever increasing amount of `$vars`. Additionally, if many views are used, this unnecessary bloats memory, because all data that was once used within a view never gets garbage collected.

Additionally this code 
```
$this->load->view('view1', ["var1" => 1]);
$this->load->view('view2', ["var2" => 2]);
```
will lead to very unexpected and in my opinion wrong behaviour if view2 relies on something like `isset($var1);`, which should evaluate to false regardless of previous operations.

I therefore propose some kind of stack handling for nested view calls which would provide the required "inheritance" without loosing isolation between separate view calls. This also allows for garbage collection for all non-global, previously used view `$vars`.

narfbg:
How do you determine which views are nested and which ones aren't?

TalonTR:
I guess there is no way to determine this, but also no need for this. It would be sufficient, if every view call would get all the `$vars` from its nested parent view call. This could be implemented via simple stack which stores the current state before including the view and removes the current state after the include.
I already implemented this approach within the referenced feature branch and it works fine as well in the unit tests as in our project, where my previously proposed change did break a lot of things.

narfbg:
That's not at all obvious from the patch and needs to be better explained with comments. And please use proper grammar (your comments currently don't use punctuation or form sentences).

Also needs to comply with our [styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html).

narfbg:
Might add unit tests as well. :)

TalonTR:
I improved the clarity and structure of both code and comments and added an additional unit test to confirm the correct behaviour. Additionally i've added another unit test that illustrates the incorrect behaviour of the previous solution.

narfbg:
I'm not happy with the comments yet (better formatted, but not any more easier to understand), but that's a secondary concern and I'll alter them myself after merging.

But the patch still doesn't conform to our styleguide (linked in my previous comment). Also, you changed an `if ... elseif` to 2 `if`s, and it seems that causes an unneccessary second merge in nested views, or am I wrong about that one?

TalonTR:
I restructured the merging process to introduce more clarity, which leads to one more merge then necessary in some cases, correct. I've done that to improve code clarity while sacrificing a bit of premature optimization. I'll take it you prefered the previous, optimised structure - no problem, i'll fix it.

Regarding the comments: sorry, thats the best i can do at the moment. If you have a better explanation with increased clarity feel free to alter them.

Would you mind pointing out whats wrong with my code format - since you seem to have something particular in mind? I've read your linked styleguide but since it's pretty lengthy i may not have checked every single requirement on every single line.

narfbg:
We put spaces around `!`, and we put opening curly braces on a separate line. You don't even need to look at the styleguide to figure this out, it's pretty obvious if you try to be consistent with how the rest of our code is written.

TalonTR:
Thanks for pointing that out, i fixed these styleguide violations.

I've considered your remark on the number of merges - i don't think one could safely reduce the number of array merges without loosing support for nested `vars()` calls to add further global variables. So i left that logic as is.

PR #5389, opened 2018-01-26
Bug fix that made the mobile not be identified by robots

stringariSM:
The way it was being done, did not allow the Google Adwords robot to identify the page as mobile.

At the time it was identified as "robot", it would not be more mobile. This issue caused alerts in Google Adwords and Analytics.

It can be checked in: https://search.google.com/test/mobile-friendly

### DIFF
--- a/system/libraries/User_agent.php
+++ b/system/libraries/User_agent.php
@@ -250,17 +250,19 @@ class CI_User_agent {
 	 * @return	bool
 	 */
 	protected function _compile_data()
-	{
-		$this->_set_platform();
+    {
+        $this->_set_platform();
 
-		foreach (array('_set_robot', '_set_browser', '_set_mobile') as $function)
-		{
-			if ($this->$function() === TRUE)
-			{
-				break;
-			}
-		}
-	}
+        foreach (array('_set_mobile', '_set_robot', '_set_browser') as $function)
+        {
+            if ($this->$function() === TRUE)
+            {
+                if ($function != '_set_mobile'){
+                    break;
+                }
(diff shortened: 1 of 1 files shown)


stringariSM:
Did I do something wrong or is it a problem with TraviCI?

narfbg:
Some builds randomly break failing to configure environments, not your fault that.

But I don't understand how `CI_User_agent' has anything to do with Google Adwords.

stringariSM:
Hello!

When you do a Google Adwords campaign for a site, it checks to see if the site is Mobile (not being mobile harms the campaign). as it was in CodeIgniter, Google's thefts did not identify the site as having a mobile version, appearing warnings that the dog would be harmed because the site was not optimized for mobile phones.

Fortunately I have already corrected this bug on the site that I was having problem with. I may try to create an example to show the problem, if necessary.

stringariSM:
In this link you can see a bug that was reported in PyroCMS, which uses CodeIgniter. The solution suggested in it was different from mine, but perhaps the explanation might help.

https://github.com/pyrocms/pyrocms/issues/3460

I apologize for my English, I'm still starting. :)

narfbg:
Please do provide an example, and explain the current vs expected result *in terms of code behavior*, not with Google Adwords. 

This was confusing enough even before you mentioned dogs being harmed ...

Lucifaar:
Buy this and Solve error on [codeigniter  Errors](https://www.fiverr.com/sajidhunxai/design-develop-codeigniter-website-with-frontend-backend)

PR #5391, opened 2018-01-28
Prevent tab hijacking when using the URL helper

mehdibo:
Fixed security issue:  allowing the target page to take control of the original page

Details about the issue here: https://mathiasbynens.github.io/rel-noopener/

### DIFF
--- a/system/helpers/url_helper.php
+++ b/system/helpers/url_helper.php
@@ -396,7 +396,7 @@ if ( ! function_exists('auto_link'))
 		if ($type !== 'email' && preg_match_all('#(\w*://|www\.)[a-z0-9]+(-+[a-z0-9]+)*(\.[a-z0-9]+(-+[a-z0-9]+)*)+(/([^\s()<>;]+\w)?/?)?#i', $str, $matches, PREG_OFFSET_CAPTURE | PREG_SET_ORDER))
 		{
 			// Set our target HTML if using popup links.
-			$target = ($popup) ? ' target="_blank"' : '';
+			$target = ($popup) ? ' target="_blank" rel="noopener"' : '';
 
 			// We process the links in reverse order (last -> first) so that
 			// the returned string offsets from preg_match_all() are not


narfbg:
Web development used to be relatively pleasant, now you need to protect yourself from the browser at every step. :(

Have you checked if `auto_link()` the only place where that's necessary?

mehdibo:
No I haven't but I will definitely check and add them.

mehdibo:
@narfbg I've checked, there is one in url_helper.php:207
But some may actually use it to control the opened popup, so I only added emptied the window.opener object if the link is external

narfbg:
Ok, cool. Thanks!

But the patch doesn't comply with our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html) (spaces before/after opening/closing parentheses; curly braces always on a new line) and the code becomes quite hard to read with all of that inline JS (maybe `sprintf()` would help).

Should be good to go once that's corrected.

mehdibo:
Sure I'll edit it, totally forgot about the style guide

mehdibo:
Fixed (I hope so I read the Style Guide), but one note, who wrote this function? I don't think it does what it is supposed to do.

It should open a popup not a new window, it works fine if you pass the `$attributes` parameter, otherwise it return a new window link.
I didn't want to fix it as I don't know the reason behind that.

narfbg:
Whoever wrote it hasn't been involved in CI for years already. If it doesn't open a popup, that's a bug.

system/helpers/url_helper.php
```diff
@@ -197,14 +197,27 @@ function anchor_popup($uri = '', $title = '', $attributes = FALSE)
 		$title = (string) $title;
 		$site_url = preg_match('#^(\w+:)?//#i', $uri) ? $uri : site_url($uri);
 
+		// Check whether the $site_url is local
+		if (substr($site_url, 0, strlen(base_url())) === base_url())
```
narfbg: Too many function calls for this check, and the if/else makes it way too long for something that simple ... This should do:

    $empty_opener = (strpos($site_url, base_url()) === 0) ? '' : ' newWindow.opener = null;';

Might also move it after the `$title` check, to be closer to where it is inserted, for clarity.

system/helpers/url_helper.php
```diff
@@ -197,14 +197,27 @@ function anchor_popup($uri = '', $title = '', $attributes = FALSE)
...
+
 		if ($title === '')
 		{
 			$title = $site_url;
 		}
 
 		if ($attributes === FALSE)
 		{
-			return '<a href="'.$site_url.'" onclick="window.open(\''.$site_url."', '_blank'); return false;\">".$title.'</a>';
+			$onclick = "var newWindow = window.open('', '_blank');{$empty_opener} newWindow.location = href; return false;";
```
narfbg: No need for this to be a variable, just split the `sprintf()` params on separate lines and it should be readable enough.

system/helpers/url_helper.php
```diff
@@ -231,10 +244,11 @@ function anchor_popup($uri = '', $title = '', $attributes = FALSE)
 		}
 
 		$attributes = _stringify_attributes($attributes);
+		$js_atts = _stringify_attributes($atts, TRUE);
+
+		$onclick = "var newWindow = window.open('', '{$window_name}', '{$js_atts}');{$empty_opener} newWindow.location = href; return false;";
```
narfbg: Same thing as with previous `$onclick` declaration.

system/helpers/url_helper.php
```diff
@@ -231,10 +244,11 @@ function anchor_popup($uri = '', $title = '', $attributes = FALSE)
 		}
 
 		$attributes = _stringify_attributes($attributes);
+		$js_atts = _stringify_attributes($atts, TRUE);
```
narfbg: What's `$atts`?

mehdibo: `$atts` was originally there, they are the attributes for the JS window, actually, the whole function is a mess IMHO as it doesn't do what it is supposed to

narfbg: I see ... Didn't notice it the first time. But now that I did look at the entire function, the description says:

> Creates an anchor based on the local URL.

So all of those changes don't seem necessary. Otherwise, as I said, if the function doesn't open a popup - that's a bug, but a separate issue from the opener hijacking possibility.

mehdibo: Well s**it... that's okay anyway the `rel` attribute is useful though, can I roll the changes back to the first one? I'm not that good with git

narfbg: You can `git reset --hard <commit>` to a specific commit and then `git push --force`.

mehdibo: I think I did it...

PR #5394, opened 2018-01-29
Give force_download a header conforming to RFC 6266

ytetsuro:
When I created an application to download multibyte file name, the file name got corrupted.

Changed give force_download a header conforming to "RFC6266"

The phenomenon that the file name is garbled can be confirmed with Internet Explorer 11 older.

https://tools.ietf.org/html/rfc6266

```php
<?php

class Downloader extends CI_Controller
{
    public function __construct()
    {
         parent::__construct();
         $this->load->helper('download');
     }

     public function download()
     {
         force_download('マルチバイト.txt', 'test');
      }
}
```

### DIFF
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -152,7 +152,7 @@ if ( ! function_exists('force_download'))
 		if (strtoupper($encoding) !== 'UTF-8')
 		{
 			$converted_filename = get_instance()->utf8->convert_to_utf8($utf8_filename, $encoding);
-			if ($converted_filename)
+			if ($converted_filename !== FALSE)
 			{
 				$utf8_filename = $converted_filename;
 			}
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -148,10 +148,10 @@ if ( ! function_exists('force_download'))
 		}
 
 		$utf8_filename = $filename;
-		$encoding = config_item('charset');
-		if (strtoupper($encoding) !== 'UTF-8')
+		$charset = config_item('charset');
+		if (strtoupper($charset) !== 'UTF-8')
 		{
-			$converted_filename = get_instance()->utf8->convert_to_utf8($utf8_filename, $encoding);
+			$converted_filename = get_instance()->utf8->convert_to_utf8($utf8_filename, $charset);
 			if ($converted_filename !== FALSE)
 			{
 				$utf8_filename = $converted_filename;
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -151,13 +151,10 @@ if ( ! function_exists('force_download'))
 		$encoding = config_item('charset');
 		if (strtoupper($encoding) !== 'UTF-8')
 		{
-			if (MB_ENABLED)
+			$CI =& get_instance();
+			if ($converted_filename = $CI->utf8->convert_to_utf8($utf8_filename, $encoding) )
 			{
-				$utf8_filename = mb_convert_encoding($utf8_filename, 'UTF-8', $encoding);
-			}
-			elseif (ICONV_ENABLED)
-			{
-				$utf8_filename = @iconv($encoding, 'UTF-8', $utf8_filename);
+				$utf8_filename = $converted_filename;
 			}
 		}
 
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -147,21 +147,21 @@ if ( ! function_exists('force_download'))
 			@ob_clean();
 		}
 
-		$utf8_filename = $filename;
+		$disposition = 'attachment; filename="'.$filename.'";';
 		$charset = config_item('charset');
 		if (strtoupper($charset) !== 'UTF-8')
 		{
-			$converted_filename = get_instance()->utf8->convert_to_utf8($utf8_filename, $charset);
-			if ($converted_filename !== FALSE)
+			// charset attribute (RFC 6266 only allows UTF-8)
+			$utf8_filename = get_instance()->utf8->convert_to_utf8($filename, $charset);
+			if ($utf8_filename !== FALSE)
 			{
-				$utf8_filename = $converted_filename;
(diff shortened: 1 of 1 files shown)


kenjis:
> The phenomenon that the file name is garbled can be confirmed with Internet Explorer 10 older.

Also IE11.

narfbg:
The title "fix rfc6266" implies that something's not compliant with the RFC in question, which is not true. It's also not descriptive at all ... You seem to have noticed that and edited it, only now you wrote the wrong function name, and you're still using the word "fix" - there's nothing broken, this is a feature request.
Please be more careful with this in the future.

On the patch itself:

I like it, a lot.

But it assumes UTF-8 ... Yes, RFC 6266 leaves no other charset choice for the header param itself, but (while I wish it was the case) that doesn't mean every CI application is based on UTF-8.
We have an application-wide charset option, so we can convert from the one specified in it if necessary.

ytetsuro:
> The title "fix rfc6266" implies that something's not compliant with the RFC in question, which is not true. It's also not descriptive at all ... You seem to have noticed that and edited it, only now you wrote the wrong function name, and you're still using the word "fix" - there's nothing broken, this is a feature request.
> Please be more careful with this in the future.

I am sorry to create a bad description in PR.
Thank you for pointing out.
I will be careful from the next.

> But it assumes UTF-8 ... Yes, RFC 6266 leaves no other charset choice for the header param itself, but (while I wish it was the case) that doesn't mean every CI application is based on UTF-8.
> We have an application-wide charset option, so we can convert from the one specified in it if necessary.

Would you like me to make a modified commit to set `charset` using `confing_item('charset')` ?

narfbg:
> ytetsuro changed the title from fixed force_redirect according to rfc6266 to Give force_redirect a header conforming to RFC 6266

... it's `force_download()`.

> Would you like me to make a modified commit to set `charset` using `confing_item('charset')` ?

I was requesting that you make a change, yes.
But not to set the charset attribute (RFC 6266 only allows UTF-8); convert to UTF-8 from `config_item('charset')` if it's not already that.

ytetsuro:
> ... it's force_download().

I'm very very very sorry...

> I was requesting that you make a change, yes.
> But not to set the charset attribute (RFC 6266 only allows UTF-8); convert to UTF-8 from config_item('charset') if it's not already that.

I understand.
committed.

system/helpers/download_helper.php
```diff
@@ -147,9 +147,21 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
 			@ob_clean();
 		}
 
+		$utf8_filename = $filename;
```
narfbg: What if it's not UTF-8 *and* you can't convert it?

system/helpers/download_helper.php
```diff
@@ -147,9 +147,21 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
...
+		$encoding = config_item('charset');
+		if (strtoupper($encoding) !== 'UTF-8')
+		{
+			$CI =& get_instance();
```
narfbg: You don't need this; you can go directly with `get_instance()->utf8`

system/helpers/download_helper.php
```diff
@@ -147,9 +147,21 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
...
+			if ($converted_filename = $CI->utf8->convert_to_utf8($utf8_filename, $encoding) )
```
narfbg: What if the filename is '000' or something? That would falsely be treated as a failure here.
Also, there shouldn't be a space before the closing parenthesis.

system/helpers/download_helper.php
```diff
@@ -147,9 +147,21 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
...
+			{
+				$utf8_filename = $converted_filename;
+			}
+		}
+
 		// Generate the server headers
 		header('Content-Type: '.$mime);
-		header('Content-Disposition: attachment; filename="'.$filename.'"');
+		// charset attribute (RFC 6266 only allows UTF-8)
```
narfbg: Mentioning that RFC6266 only allows UTF-8 is worth it, but before all the charset conversion logic, not here ...
 here it can only be confusing, not least because there's no "charset" attribute in the header.

system/helpers/download_helper.php
```diff
@@ -147,9 +147,21 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
 			@ob_clean();
 		}
 
+		$utf8_filename = $filename;
+		$encoding = config_item('charset');
```
narfbg: While at the other stuff, might as well rename this one to `$charset` ...

system/helpers/download_helper.php
ytetsuro: Thank you for pointing out.
A character string that is not UTF-8 will be assigned.
I fixed it with the following commit.

https://github.com/bcit-ci/CodeIgniter/pull/5394/commits/8b49b1b17c9c313bcdc01d247854667b8eb35679

system/helpers/download_helper.php
ytetsuro: Thank you for pointing out.

> What if the filename is '000' or something?
Oops!
Since convert_to_utf8 may return false, we wrote like this, but we did not consider that `000` will come.
I fixed it with the following commit.

https://github.com/bcit-ci/CodeIgniter/pull/5394/commits/e5d719d822646a7c0bcd1a2732eb8fe19357444a

ytetsuro:
Thank you for pointing out.
All the corrections to the points pointed out have been completed.

system/helpers/download_helper.php
```diff
@@ -147,9 +147,21 @@ function force_download($filename = '', $data = '', $set_mime = FALSE)
 			@ob_clean();
 		}
 
+		$disposition = 'attachment; filename="'.$filename.'";';
+		$charset = config_item('charset');
+		if (strtoupper($charset) !== 'UTF-8')
+		{
+			// charset attribute (RFC 6266 only allows UTF-8)
```
narfbg: Again, there's no "charset" attribute.

But to avoid further back-and-forth, I'll correct this myself.

PR #5397, opened 2018-01-31
File name changing in dynamic output image_lib

mehdibo:
Allow changing file name (`Content-Disposition` header) when outputting the image by specifying the `new_name` config

### DIFF
--- a/system/libraries/Image_lib.php
+++ b/system/libraries/Image_lib.php
@@ -1548,7 +1548,8 @@ class CI_Image_lib {
 	 */
 	public function image_display_gd($resource)
 	{
-		header('Content-Disposition: filename='.( (empty($this->new_image)) ? $this->source_image : $this->new_image ).';');
+		$filename = basename(empty($this->new_image) ? $this->source_image : $this->new_image);
+	        header('Content-Disposition: filename='.$filename.'; filename*=UTF-8\'\''.rawurlencode(iconv(config_item('charset'), 'UTF-8', $filename)));
 		header('Content-Type: '.$this->mime_type);
 		header('Content-Transfer-Encoding: binary');
 		header('Last-Modified: '.gmdate('D, d M Y H:i:s', time()).' GMT');
--- a/system/libraries/Image_lib.php
+++ b/system/libraries/Image_lib.php
@@ -1548,7 +1548,7 @@ class CI_Image_lib {
 	 */
 	public function image_display_gd($resource)
 	{
-	        header('Content-Disposition: filename='.basename(empty($this->new_image) ? $this->source_image : $this->new_image).';');
+		header('Content-Disposition: filename='.basename(empty($this->new_image) ? $this->source_image : $this->new_image).';');
 		header('Content-Type: '.$this->mime_type);
 		header('Content-Transfer-Encoding: binary');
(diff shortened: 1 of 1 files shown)


system/libraries/Image_lib.php
```diff
@@ -1548,7 +1548,7 @@ public function image_save_gd($resource)
 	 */
 	public function image_display_gd($resource)
 	{
-		header('Content-Disposition: filename='.$this->source_image.';');
+		header('Content-Disposition: filename='.( (empty($this->new_image)) ? $this->source_image : $this->new_image ).';');
```
narfbg: There shouldn't be parentheses around the `empty()` call (nor a space before it).

Also, these properties are supposed to contain paths, not just names (which might mean it also doesn't currently work properly)

mehdibo: The function originally used `$this->source_image` which means it's supposed to work, and `$this->source_image` is not meant to be set when you are outputting the image, so I think It can be used to change the `filename`

narfbg: > The function originally used `$this->source_image` which means it's supposed to work

Supposed, but we don't really know if that's true; this entire class is a mess. And it may've been modified to be usable at this point and `new_image` not be.

Maybe use `basename()` to be sure.

> and `$this->source_image` is not meant to be set when you are outputting the image

Eh ... what?

---

On another note, see #5394 - it's not yet merged, but very relevant.

mehdibo: I meant `$this->new_image`, well I don't know then, I'm using it in a project and thought why not share it, do what you think is the best.

About #5394 no that's not what I was intending by this PR, the main reason for this edit is I'm storing a map of images in the database and the file names are 'encrypted' and long, when using the Image_lib to output the image after a resize it shows the full name, I added this so I can set a more friendly name.

But anyway this wasn't tested "that much" so I can't guarantee it.

narfbg: I know #5394 is not what you're trying to do here. What I'm saying is you might want to do that *as well*.

Just use `basename()` on the ternary result to ensure that only a filename is output, without possible directory components.

mehdibo:
I didn't understand #5394 that well, what I understood from your comment is that we should convert the charset to UTF-8.
I tested with **マルチバイト** but it outputs nothing (empty string), but when I add a character **before** the string (e.g. xマルチバイト) it worked fine.

narfbg:
That shouldn't be the case, but this also shouldn't explicitly rely on iconv either. On another thought ... forget that thing, #5394 itself is messy and I shouldn't have bothered you with it - sorry about that.

Revert it to just your originally intended new feature and it should be ready to go.

mehdibo:
Sure, no problem, don't worry about it.
I made the changes

system/libraries/Image_lib.php
```diff
@@ -1548,7 +1548,7 @@ public function image_save_gd($resource)
 	 */
 	public function image_display_gd($resource)
 	{
-		header('Content-Disposition: filename='.$this->source_image.';');
+	        header('Content-Disposition: filename='.basename(empty($this->new_image) ? $this->source_image : $this->new_image).';');
```
narfbg: You're replacing a tab with spaces here ...

mehdibo: I have no idea how I did that, I fixed it.

PR #5408, opened 2018-02-12
Allow cross join (join on true) in query builder

mmapes:
When upgrading to CodeIgniter 3, my cross joins stopped working. That's because the the join command assumes any JOIN condition without an operator must be a USING. But that is a faulty assumption, as JOIN ON TRUE is valid SQL for generating a CROSS JOIN.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -526,7 +526,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	public function join($table, $cond, $type = '', $escape = NULL)
 	{
 		$type = trim(strtoupper($type).' JOIN');
-		preg_match('#^(((NATURAL|CROSS)\s+)|((LEFT|RIGHT)\s+)?((INNER|OUTER))\s+)?JOIN$#', $type) OR $type = 'JOIN';
+		preg_match('#^((CROSS\s+)|((NATURAL\s+)?((LEFT|RIGHT)\s+)?((INNER|OUTER))\s+))?JOIN$#', $type) OR $type = 'JOIN';
 
 		// Extract any aliases that might exist. We use this information
 		// in the protect_identifiers to know whether to add a table prefix
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -538,6 +538,11 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		{
 			$cond = '';
 		}
+		// Check CROSS JOIN (JOIN ON TRUE) before checking !_has_operator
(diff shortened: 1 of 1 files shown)


narfbg:
Do all databases support the `ON TRUE` condition? And does that differ from `CROSS JOIN` without a condition?

mmapes:
Not all databases support ON TRUE. For example, MS SQL Server does not allow ON TRUE. However, Query Builder outputs JOIN USING, and that is not allowed on MS SQL Server either. 

Can you tell me which databases are officially supported by Query Builder so that I can research whether they all allow ON TRUE?

Mysql, Postgresql, SQLite all support it.

I could not find any information on whether Oracle supports it.

narfbg:
Anything that you see under system/database/drivers/ and system/database/drivers/pdo/subdrivers/ is relevant, but the major ones that you've already mentioned are what really matters ...

My question was rather whether this is part of the SQL standard. And again, does this differ from `CROSS JOIN` and how?

If it doesn't differ, then `CROSS JOIN` should be the better option to implement, as I'm pretty sure it's supported everywhere.

mmapes:
I understand what you are saying. Yes, CROSS JOIN is "standard." I have added a new commit to that effect. 

As to the join condition, it is almost always referred to as "boolean condition" and seems to be very widely allowed to be "true." The alternative interpretation is that [JOIN + condition: 'TRUE'] is intended to be a JOIN USING on a column whose name is TRUE. That seems much less likely an intention to me than a CROSS JOIN, but I admit I am biased.

On another note, I would have tested the join code if I had my tests set up, but I couldn't figure it out. If you are a forum guy and have a tutorial... see https://forum.codeigniter.com/forum-17.html.

narfbg:
You've linked to an entire forum section; I assume you mean [this thread](https://forum.codeigniter.com/thread-70063.html)? There's a README.md file in the tests/ dir, but it almost hasn't been touched since tests were introduced and doesn't contain instructions about DB tests, so it's kinda useless. Even I don't run the DB tests locally, they are very tailored to the Travis environment.  
You can click on the build results link at the end of the PR to check what fails though, and you've broken `LEFT` and `RIGHT` joins, perhaps because you tried to disable `NATURAL LEFT OUTER` and such ... I don't want to necessarily say don't fix what's not broken, but that's surely out of scope for this PR.  
The regexp should be: `#^(CROSS\s+)?|((NATURAL\s+)?((LEFT|RIGHT)\s+)?((INNER|OUTER)\s+))?JOIN$#`

However, you've preserved the condition check for `ON TRUE` as is, while my intention was to only support `CROSS` and potentially replace `JOIN ON TRUE` with `CROSS JOIN`. Or do you have an argument against that?

PR #5428, opened 2018-02-28
oci8 database driver: Added support of insert_id()

giterlizzi:
Added support of ``$db->insert_id()`` method in OCI8 database driver. By default this method return the ROWID Oracle pseudocolumn, but you can specify another column (eg. with a sequence) using ``$db->rowid_column`` public property.

This is a sample:

```.php
$this->db->insert('FOO', [
  'FOO_FIELD' => 'BAR',
]);

var_dump($this->db->insert_id());
// string(18) "AAAmpiAAEAAAHerAAE"
```

In alternative you can specify a column with a sequence:

```.php
$this->db->rowid_column = 'FOO_ID';
$this->db->insert('FOO', [
  'FOO_FIELD' => 'BAR',
]);

var_dump($this->db->insert_id());
// string(18) "1001"
```

### DIFF
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -99,6 +99,15 @@ class CI_DB_oci8_driver extends CI_DB {
 	 */
 	public $limit_used;
 
+	/**
+	 * ROWID column name or table-specific column name
+	 *
+	 * Used to fetch the value from ROWID (or table-specific) column
+	 *
+	 * @var string
+	 */
+	public $rowid_column = 'ROWID';
+
 	// --------------------------------------------------------------------
 
 	/**
@@ -134,6 +143,13 @@ class CI_DB_oci8_driver extends CI_DB {
 	 * @var	string
 	 */
 	protected $_count_string = 'SELECT COUNT(1) AS ';
+
+	/**
+	 * OCI8 ROWID
+	 *
+	 * Used by _execute() to fetch Oracle ROWID
+	 */
+	protected $_oci8_rowid  = -1;
 
 	// --------------------------------------------------------------------
 
@@ -280,6 +296,10 @@ class CI_DB_oci8_driver extends CI_DB {
 		if ($this->_reset_stmt_id === TRUE)
 		{
 			$this->stmt_id = oci_parse($this->conn_id, $sql);
+		}
+
+		if (strstr($sql, ':ci_oci8_rowid')) {
+			oci_bind_by_name($this->stmt_id, ":ci_oci8_rowid", $this->_oci8_rowid, 255);
 		}
 
 		oci_set_prefetch($this->stmt_id, 1000);
@@ -438,8 +458,7 @@ class CI_DB_oci8_driver extends CI_DB {
 	 */
 	public function insert_id()
 	{
-		// not supported in oracle
-		return $this->display_error('db_unsupported_function');
+		return $this->_oci8_rowid;
 	}
 
 	// --------------------------------------------------------------------
@@ -582,6 +601,24 @@ class CI_DB_oci8_driver extends CI_DB {
 		return is_array($error)
 			? $error
 			: array('code' => '', 'message' => '');
+	}
+
+	// --------------------------------------------------------------------
+
+	/**
+	 * Insert statement
+	 *
+	 * Generates a platform-specific insert string from the supplied data
+	 *
+	 * @param	string	the table name
+	 * @param	array	the insert keys
+	 * @param	array	the insert values
+	 * @return	string
+	 */
+	protected function _insert($table, $keys, $values)
+	{
+		return 'INSERT INTO '.$table.' ('.implode(', ', $keys).') VALUES ('.implode(', ', $values).')'.
+		       ' RETURNING ' . $this->rowid_column . ' INTO :ci_oci8_rowid';
 	}
 
 	// --------------------------------------------------------------------
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -463,9 +463,9 @@ class CI_DB_oci8_driver extends CI_DB {
 
 		if ($column !== NULL)
 		{
-			$sql =  'SELECT ' . strtoupper($column) . ' AS SEQ ' .
-					'FROM ' . strtoupper($this->_table) . ' ' .
-					'WHERE ROWID = ' . $this->escape($this->_rowid);
+			$sql = 'SELECT ' . strtoupper($column) . ' AS SEQ ' .
+			       'FROM ' . strtoupper($this->_table) . ' ' .
+			       'WHERE ROWID = ' . $this->escape($this->_rowid);
 			$seq = $this->query($sql)->row()->SEQ;
 			return $seq;
 		}
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -99,15 +99,6 @@ class CI_DB_oci8_driver extends CI_DB {
 	 */
 	public $limit_used;
 
-	/**
-	 * ROWID column name or table-specific column name
-	 *
-	 * Used to fetch the value from ROWID (or table-specific) column
-	 *
-	 * @var string
-	 */
-	public $rowid_column = 'ROWID';
-
 	// --------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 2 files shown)


jqr91:
Hi, thanks..I hope the codeigniter team will soon integrate this thing... great!

narfbg:
Modifying a system property to get results isn't acceptable. You can make that one an argument to `insert_id()` (we do that for PostgreSQL, for example).

I'll leave some inline comments for other necessary changes.

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -135,6 +144,13 @@ class CI_DB_oci8_driver extends CI_DB {
 	 */
 	protected $_count_string = 'SELECT COUNT(1) AS ';
 
+	/**
+	 * OCI8 ROWID
+	 *
+	 * Used by _execute() to fetch Oracle ROWID
+	 */
+	protected $_oci8_rowid  = -1;
```
narfbg: Why does this default to -1?

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -282,6 +298,10 @@ protected function _execute($sql)
 			$this->stmt_id = oci_parse($this->conn_id, $sql);
 		}
 
+		if (strstr($sql, ':ci_oci8_rowid')) {
```
narfbg: I assume the intention here is just to check if ':ci_oci8_rowid' exists in the SQL? If so, use `strpos()` instead.

Also, please read our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html); the opening brace must be on the next line.

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -586,6 +605,24 @@ public function error()
...
+	 *
+	 * @param	string	the table name
+	 * @param	array	the insert keys
+	 * @param	array	the insert values
+	 * @return	string
+	 */
+	protected function _insert($table, $keys, $values)
+	{
+		return 'INSERT INTO '.$table.' ('.implode(', ', $keys).') VALUES ('.implode(', ', $values).')'.
+		       ' RETURNING ' . $this->rowid_column . ' INTO :ci_oci8_rowid';
```
narfbg: No spaces around concatenation please.
Also, if this line is concatenated to the above one, then the dot must start this line and not end the previous one.

narfbg:
Also, same functionality should be added to the pdo/oci8 driver.

LotarProject:
Hi @narfbg , thanks for your review.

I have pushed a fix with your suggested improvements and the support of the same functionality in PDO OCI driver.

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -135,6 +135,20 @@ class CI_DB_oci8_driver extends CI_DB {
...
+	 * Used by _execute() to fetch Oracle ROWID
+	 */
+	protected $_rowid = NULL;
+
+	/**
+	 * Oracle table
+	 *
+	 * Used by insert_id() method
+	 */
+    protected $_table = NULL;
```
narfbg: In addition to ignoring my previous comment about following our [styleguide](https://codeigniter.com/userguide3/general/styleguide.html), you're now also using spaces for indentation.

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -438,8 +457,20 @@ public function affected_rows()
 	 */
 	public function insert_id()
 	{
-		// not supported in oracle
-		return $this->display_error('db_unsupported_function');
+		if (! $this->_rowid) return NULL;
```
narfbg: Also violating the [styleguide](https://codeigniter.com/userguide3/general/styleguide.html).

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -438,8 +457,20 @@ public function affected_rows()
...
+
+		$column	= (func_num_args() > 0) ? func_get_arg(0) : NULL;
```
narfbg: Why is this not declared as a regular function parameter?

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -438,8 +457,20 @@ public function affected_rows()
...
+
+		if ($column !== NULL)
+		{
+			$sql = 'SELECT ' . strtoupper($column) . ' AS SEQ ' .
```
narfbg: Why `strtoupper()`? Yet no identifier escaping?

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -438,8 +457,20 @@ public function affected_rows()
...
+			       'FROM ' . strtoupper($this->_table) . ' ' .
```
narfbg: Same as the previous line.
But also, again, why is the `$_table` property necessary at all? There's nothing preventing it from being a function parameter.

In addition, why select directly from the table? I'm very certain there's a way to simply select the current sequence value and not bother even knowing which the table is.

system/database/drivers/oci8/oci8_driver.php
```diff
@@ -438,8 +457,20 @@ public function affected_rows()
...
+			       'WHERE ROWID = ' . $this->escape($this->_rowid);
+			$seq = $this->query($sql)->row()->SEQ;
```
narfbg: This assignment is unnecessary; you can directly `return` this expression.

system/database/drivers/oci8/oci8_driver.php
LotarProject: Sorry. I have switched temporary the indentation style in my editor.

system/database/drivers/oci8/oci8_driver.php
LotarProject: :+1:

system/database/drivers/oci8/oci8_driver.php
LotarProject: This is the actual behavior used by Postgres driver. I have supposed this behavior is used to prevent eventually signature mismatch error or for compatibility

system/database/drivers/oci8/oci8_driver.php
LotarProject: The uppercase is a standard de-facto for Oracle. In this case i will be use ``escape_str()`` method

system/database/drivers/oci8/oci8_driver.php
LotarProject: Hi,
The ``_table`` property is used to temporally "save" the table specified by ``insert()`` method.

I select directly from the table because Oracle don't support natively the auto-increment column. The only way for Oracle is:
 - Create a primary key column
 - Create the sequence
 - Create the trigger on insert with the sequence

From Oracle 12c is introduced IDENTITY column attribute but don't solve the problem of ``insert_id()`` method.

My solution is:
 - Create on-the-fly a bind output to return the ROWID of inserted row data using ``RETURNING INTO`` clausole
 - Use this ROWID for fetch the user-defined record identifier (generally the primary key of the table) by a simple select (eg. ``insert_id("MY_COLUMN__ID")``)
 - In alternative if the user don't specify the column, return the Oracle ROWID by default

system/database/drivers/oci8/oci8_driver.php
LotarProject: :+1:

system/database/drivers/oci8/oci8_driver.php
narfbg: Right, the signature mismatch thing it is.

system/database/drivers/oci8/oci8_driver.php
narfbg: I know it's a de-facto standard, but not a universal one. The user may not want to comply with it and by enforcing it, you're getting in their way.

Also, `escape_str()` escapes string literals, not identifiers ... You need `escape_identifiers()`.

system/database/drivers/oci8/oci8_driver.php
narfbg: Yes, I am very aware of sequences in Oracle, hence my previous comment ... The following is all you need:

    SELECT <sequence_name>.CURRVAL FROM dual

In fact, just make `insert_id()` accept a `$sequence_name` parameter and nothing else; drop all other changes as they are necessary.

narfbg:
Too old, feedback wasn't addressed. If anyone has interest in this, please submit an up to date version.

PR #5429, opened 2018-03-04
Fixed bug in ZIP library

mehdibo:
the `read_dir` method ignores files that start with a dot (like .htaccess).
I looked through the documentation and it didn't mention that the method ignores hidden files so it must be a bug.

### DIFF
--- a/system/libraries/Zip.php
+++ b/system/libraries/Zip.php
@@ -366,7 +366,7 @@ class CI_Zip {
 
 		while (FALSE !== ($file = readdir($fp)))
 		{
-			if ($file[0] === '.')
+			if ($file === '.' OR $file === '..')
 			{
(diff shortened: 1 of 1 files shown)


narfbg:
If this is a bug, [it dates all the way back to 2006](https://github.com/bcit-ci/CodeIgniter/commit/5725a40301b3aceab1b838e33d003ecaf7ab113e). :D

But while you say it "must" be a bug, there's valid reasons for wanting to exclude hidden files from archives and so this is a breaking change. I agree with it, but it will have to wait for 3.2.0 and won't be released in a 3.1.x version.

mehdibo:
Maybe I should add a parameter for Excluding hidden files?

narfbg:
No.

PR #5431, opened 2018-03-05
Added parenthesis check around "document" elements and fixed non-existent document.window filter to window.document

CyberSecutor:
`document.cookie` and `document.write` are filtered but can easily be evaded by `(document).cookie` and `(document).write`.

Also the `$_never_allowed_regex` checked for `document.window` which is always undefined and won't work.
It should be `window.document`.

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -134,7 +134,9 @@ class CI_Security {
 	 */
 	protected $_never_allowed_str =	array(
 		'document.cookie' => '[removed]',
+		'(document).cookie' => '[removed]',
 		'document.write'  => '[removed]',
+		'(document).write'  => '[removed]',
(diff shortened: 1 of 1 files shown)


narfbg:
That's cool, but parts of your patch *only* add parentheses without preserving the old checks.

Also, we have private channels for security-related stuff.

CyberSecutor:
I'm not sure if i understand you correctly but the old checks are preserved.

I added the parenthesis `\(?document\)?` with the `?` modifier so they are optional, preserving the old check.

For this check:
```php
'(\(?document\)?|\(?window\)?(\.document)?)\.(location|on\w*)',
```
the old behavior is changed but that is the point, the old behavior was invalid.

narfbg:
My bad, I didn't notice the `?` modifier.

PR #5441, opened 2018-03-18
form validation issue

pgee70:
I was having an issue with form validation.  I use a my_form_validation extension instead of callbacks.
i have a field, that depending on context may or may not be valid as empty string.  so the $postdata ==='' needs to be evaluated by $this->$rule . This single-line mod makes sure that the code is evaluated if $this->$rule is present even if the $postdata is empty.

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -702,6 +702,7 @@ class CI_Form_validation {
 				&& $callback === FALSE
 				&& $callable === FALSE
 				&& ! in_array($rule, array('required', 'isset', 'matches'), TRUE)
+				&& method_exists($this, $rule) === FALSE
 			)
 			{
 				continue;


narfbg:
If you need this because you made an extension, then incorporate it in the extension ...

pgee70:
you dont get it. the extension does not run because the postdata is = ''

narfbg:
I do get it.

Still a thing that only you need for your own extension.

pgee70:
how can i make you understand then?
i am trying to communicate my extension does not get the chance to run.  consequently putting logic in it is of no use because it doesn't get to run!

i spent an hour tracing this, my extension set in $this->$rule did not run because the post data was empty.  adding this extra check in the PR makes it work.

if for example i had a callback it would have worked.  i don't like callbacks because they are linked to a particular controller, and its nicer to extend form_validation.

narfbg:
Again, I understand you perfectly well.  
And again, just *incorporate it in the extension* itself.

If there's any lack of understanding here, it's on your end ... The ability to make those extensions is the ability to entirely change the stock behavior; you can *override* any library method with your own, not just add new ones.

pgee70:
so perhaps you would have made yourself more clear by writing extend the _execute method.  you did not explicitly say which method you meant, and i had assumed you had meant to extend my $this->$rule method, which i couldn't understand because it was never executed.

i created this pull request because i think that the current behaviour of the CI validation is not what people would expect, hence i thought it would benefit the community if this PR was incorporated into source.  if you don't want to do that so be it.

narfbg:
> so perhaps you would have made yourself more clear by writing extend the _execute method. you did not explicitly say which method you meant, and i had assumed you had meant to extend my $this->$rule method, which i couldn't understand because it was never executed.

That would've been an objectively clearer message from me, yes. But if you look at my very first reply again:

> If you need this because you made an extension, then incorporate it in the extension ...

Assuming fluency in English and unless you misread it, it leaves no room to interpret it the way you did. The natural response to misunderstanding something is "Oh, I see now"; not blaming the other person.

> i created this pull request because i think that the current behaviour of the CI validation is not what people would expect, hence i thought it would benefit the community if this PR was incorporated into source. if you don't want to do that so be it.

Well, I can guarantee you that the current behavior is what people expect, because a) it's been working like that forever and you're the first one to want to change it' and b) we've gotten quite a lot of complaints about callbacks triggering on empty, non-required fields.

And I don't appreciate that little jab at the end, implying that it's simply me that doesn't want this change. You thinking what people would expect is based on your own isolated experience; my position is based on years of responding to everybody's issues.

Issue #5454, opened 2018-03-28
Security component xss_clean() logic imperfection.


bugexorcist:
When input received, sometimes valid symbol sequences like "% b" in "100 things 5% better" get captured.

How to reproduce: in my particular case a huge serialized string sent to CI-powered API backend got spoiled.

narfbg:
Duplicate of #5125.

PR #5455, opened 2018-03-28
Issue #5454: CI_Security::xss_clean() regexp fixed to avoid capturing valid symbol sequences

bugexorcist:
The CI_Security::xss_clean() is supposed to allow percent symbol and the following space, so that valid texts are not cleaned out.


bugexorcist:
I presumed no need for searching for closed issues given that problem persists.

PR #5457, opened 2018-03-29
Create cli-production.php for CLI Bootstrap

yidas:
Codeigniter 3 CLI Bootstrap for production environment

`$_SERVER['CI_ENV']` could not be determined when running on CLI, we could create a CLI bootstrap to solve production CLI problem such as running crontab.

### DIFF
--- a/cli-production.php
+++ b/cli-production.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * Codeigniter 3 CLI Bootstrap for production environment
+ */
+
+// CLI bootstrap only
+if (php_sapi_name() !== 'cli') {
+    
+    die('Access Denied');
+}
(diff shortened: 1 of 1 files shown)


nanndemoiikara:
Can you satisfy the request with the following command?

```shell
$ CI_ENV="production" php index.php welcome/index
```

```crontab
*/5 * * * * CI_ENV="production" php index.php welcome/index
```

yidas:
It's a good solution for Linux, but is not suited for other command environment such as Windows.

narfbg:
We don't need this and you don't need `$_SERVER['CI_ENV']` - it's a convenient default placeholder, but index.php is ultimately a configuration file and you are *expected* to alter it to change the environment.

PR #5460, opened 2018-04-01
Master

Mancy:
The is_referer function wasn't working because of the triple equal sign


Mancy:
sorry I will correct and resubmit

PR #5461, opened 2018-04-01
Fixing the is_referer function 

Mancy:
in User_agent library the is_referer function was always returning false


popcorner:
First, this function name is `is_referral`
And maybe you should read the user guide

> Returns TRUE/FALSE (boolean) if the user agent was referred from **ANOTHER SITE**.

Mancy:
I would then suggest amendment in the function name to is_{out/external/ext}_referral or whatever meaningful or addition of a parameter is_referral($external= true)

narfbg:
`isset($_SERVER['HTTP_REFERER'])` - there you go.

Issue #5462, opened 2018-04-02
Query Builder LIKE BEFORE doesnt work CI 3.1.8


romainrg:
There is an issue with LIKE method of query builder class when 'BEFORE' parameter is passed.

**This code generates database error :**
`$this->db->select('*')->like('id', 'zefzefezf', 'BEFORE')->get('email_v2');`

**Result :**
`SELECT * FROM `email_v2` WHERE id LIKE %'zefzefezf' ESCAPE '!'`

The 'BOTH' and 'AFTER' parameter are working properly.

romainrg:
![image](https://user-images.githubusercontent.com/14097222/38199799-54437d9c-3692-11e8-80b1-e474c83339d6.png)

narfbg:
Please don't post screenshots of code. It really doesn't help, and it's annoying.

PR #5464, opened 2018-04-03
Fixed issue #5462 Query Builder LIKE BEFORE doesnt work

ytetsuro:
#5462

### DIFF
--- a/tests/codeigniter/database/query_builder/like_test.php
+++ b/tests/codeigniter/database/query_builder/like_test.php
@@ -119,11 +119,11 @@ class Like_test extends CI_TestCase {
 
 	public function like_set_side_provider()
 	{
-		return [
-			['Developer', 'none', 'Developer'],
-			['tician', 'before', 'Politician'],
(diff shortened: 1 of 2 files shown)


tests/codeigniter/database/query_builder/like_test.php
```diff
@@ -103,4 +103,27 @@ public function test_like_spaces_and_tabs()
...
+	public function test_like_set_side($str, $side, $expected_name)
+	{
+		$actual = $this->db->like('name', $str, $side)->get('job')->result_array();
+		$this->assertCount(1, $actual);
+		$this->assertEquals($expected_name, $actual[0]['name']);
+	}
+
+	public function like_set_side_provider()
+	{
+		return [
```
narfbg: Please change to the old `array()` notation so this can be backported into 3.1-stable.

ytetsuro: changed.

PR #5466, opened 2018-04-06
Reset cached variables in load->view()

BathoryPeter:
Variable cache allows "views that are embedded within other views can have access to these variables" - according to the documentation.

But it has a side effect: if you load views in a row, the later loaded ones can access variables from the previous ones. E.g.:

`$this->load->view('first-view', ['firstVariable' => 'foo']);`
`$this->load->view('second-view');`

In second-view.php, you can access $firstVariable.

Resetting cache avoids this but keeps the embedding functionality.


narfbg:
See #5373 and please do search for existing issues in the future.

PR #5470, opened 2018-04-11
Bug line 1045

stefanobeccalli:
The char % is in wrong position. All db query with "like before" failed.


ytetsuro:
created.

@see: https://github.com/bcit-ci/CodeIgniter/pull/5464

PR #5471, opened 2018-04-12
The value of limit can be zero

toonitw:
The query builder tests if `$qb_limit` is the default value `FALSE`. So you can code like below;

    // $params->limit = 0;
    $this->db->limit($params->limit);

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -1469,7 +1469,7 @@ abstract class CI_DB_driver {
 		return 'UPDATE '.$table.' SET '.implode(', ', $valstr)
 			.$this->_compile_wh('qb_where')
 			.$this->_compile_order_by()
-			.($this->qb_limit ? ' LIMIT '.$this->qb_limit : '');
+			.(($this->qb_limit !== FALSE) ? ' LIMIT '.$this->qb_limit : '');
 	}
 
 	// --------------------------------------------------------------------
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -1469,7 +1469,7 @@ abstract class CI_DB_driver {
 		return 'UPDATE '.$table.' SET '.implode(', ', $valstr)
 			.$this->_compile_wh('qb_where')
 			.$this->_compile_order_by()
-			.(($this->qb_limit !== FALSE) ? ' LIMIT '.$this->qb_limit : '');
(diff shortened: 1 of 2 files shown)


narfbg:
Sure, you technically can have a `LIMIT 0` clause. But **what's the point?**

toonitw:
Currently I cannot query `limit(0)` because the query builder ignores. 

```
$sql = $this->db->limit(0)->get_compiled_select('posts');
// SELECT * FROM posts 
```

gxgpet:
@toonitw The question is why do you need `LIMIT 0`? Give an example/situation where this is needed.

toonitw:
I am in some legacy project. The project has a batch program that manipulates several db records. 

Administrator wants to control *batch size* the number of db records, on the dashboard. Also he wants to stop batch work by setting this value to zero.

I can do this with `if` syntax. But I thought it's more concise to use query builder naturally.

system/database/DB_driver.php
```diff
@@ -1469,7 +1469,7 @@ protected function _update($table, $values)
 		return 'UPDATE '.$table.' SET '.implode(', ', $valstr)
 			.$this->_compile_wh('qb_where')
 			.$this->_compile_order_by()
-			.($this->qb_limit ? ' LIMIT '.$this->qb_limit : '');
+			.(($this->qb_limit !== FALSE) ? ' LIMIT '.$this->qb_limit : '');
```
narfbg: Don't need the extra parentheses.

system/database/DB_query_builder.php
```diff
@@ -2287,7 +2287,7 @@ public function delete($table = '', $where = '', $limit = NULL, $reset_data = TR
 	protected function _delete($table)
 	{
 		return 'DELETE FROM '.$table.$this->_compile_wh('qb_where')
-			.($this->qb_limit ? ' LIMIT '.$this->qb_limit : '');
+			.(($this->qb_limit !== FALSE) ? ' LIMIT '.$this->qb_limit : '');
```
narfbg: Again, no need for the parentheses.

narfbg:
Ok, I can imagine this and similar use cases and I'll merge this after the requested changes.

You're still better off with an `if` check though.

toonitw:
@narfbg Ok, thanks.

PR #5488, opened 2018-05-01
Asset Versioning

roliod:
Overview
----------------------------------------
This update enables the addition of a version identifier to assets such as JS, CSS and Images 

Before
----------------------------------------
The **base_url()** or **site_url()** helper functions are usually used for assets but it makes it difficult for end users to see updates when an asset is cached and then updated. 

After
----------------------------------------
An introduction of a helper function called **asset_version()** for assets. This would give all assets version identifiers which would thereby force a browser to refresh the assets whenever an update is made. 

Usage
----------------------------------------
- Add **'asset'** to the helper autoload

---

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -469,3 +469,19 @@ $config['time_reference'] = 'local';
 | Array:		array('10.0.1.200', '192.168.5.0/24')
 */
 $config['proxy_ips'] = '';
+
+/*
+|--------------------------------------------------------------------------
+| Asset Version
+|--------------------------------------------------------------------------
+|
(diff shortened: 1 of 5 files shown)


alex2005git:
How about using filemtime on the asset, and not have to deal with changing the version manually?
I've used it on other projects without problems, it gets cached by PHP so no problems there.

narfbg:
I've always hated this feature wherever I've seen it - not only is it a hack around HTTP's core caching mechanics, but it does from the server side to solve a client-side problem.

You're free to publish this as a third-party extension/addon/whatever (composer too), but I'm not merging it into CI. Sorry.

PR #5490, opened 2018-05-02
Add exception handling

jim-parry:
Restart of the exception handling code, since the earlier feature branch was way too stale.


otg-chris (commented):
This appears to interfere with the existing error messaging and handling like the one of the things that ruined CI4. It may cause fatal errors unless there's a try-catch above it.

Issue #5329, opened 2017-11-07
DELETE doesn't allow JOINs


Ukuser36:
The following query is perfectly legit but CI query Builder will not allow it:

DELETE lc.*,lu.*
FROM users lu
JOIN companies lc ON lu.CompanyID = lc.LoginCompanyID
WHERE lu.UserID = 111

It will not work because of the JOIN. The fudge solution is to create it as a select and just str replace on the compiled_select with DELETE. Obviously not ideal.

narfbg:
Linking #2944.

This will probably disappoint you, but it's the way it is ... I have no time or desire to implement this, and there certainly are other things with way higher priority. So, the only way for this to happen is if somebody submits a pull request implementing it. *But*, that doesn't mean we'd merge any such PR - all quality requirements still apply and it has to be properly cross-platform (as I've said in #2944), which is no easy or small task.

Ukuser36:
Fair enough - amending the select for now it is then

robertmain:
Sounds like you might want a `CASCADE` on that column, no?

PR #5495, opened 2018-05-03
Update method _delete

taillis:
I did a change to a method called _delete, allowing it to delete with joins.


narfbg:
Read my comment on #5329 - this is neither platform-agnostic, nor is the code of particularly high quality (it's a quick hack).

PR #5499, opened 2018-05-14
Xss filter

qqqqiiiip:
%3Cscri%ffpt%3E  can not translate <removed> , so add "%ff filter" in common.php


narfbg:
I'm sorry but I don't at all understand what the goal is here. Please elaborate.

narfbg:
Also, seems that #5499 is a duplicate of this, but with better formatting.

Closing, please reply there.

PR #5500, opened 2018-05-16
'xss_clean-fixbug'

qqqqiiiip:
xss_clean()  %3Cscri%ffpt%3E3C%2fscrip%fft%3E  can not translate <removed> ,but <script></script>


narfbg:
#5499

qqqqiiiip:
...sorry,my fault....I previously thought this was an xss hole that would execute js. but not.. Tested wrong

PR #5502, opened 2018-05-17
Changed links to https:// when possible

mehdibo:
Changed all http:// links to the https:// version whenever possible (if the website has an HTTPS version)

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -88,7 +88,7 @@ $config['language']	= 'english';
 | This determines which character set is used by default in various methods
 | that require a character set to be provided.
 |
-| See http://php.net/htmlspecialchars for a list of supported charsets.
+| See https://secure.php.net/htmlspecialchars for a list of supported charsets.
 |
 */
 $config['charset'] = 'UTF-8';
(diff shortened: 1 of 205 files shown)


jim-parry:
Good idea, but a PR should address just one thing.
In addition to changing http:// links where appropriate, you
- change a browser user agent name
- add new HTTP status codes
- change config->item to config->slash_item
- change a security regex expressions
- correct some typos
- change a DB_driver return value
- added CLI test methods
and possibly many more.

These groups of changes need to be separate, to make it easier to accept or reject the groups, rather than an all or nothing approach. 
It would be easier for you if changes for a PR were in a separate feature branch in your repo, to isolate them from other changes.

To reiterate, each group of related changes needs to be proposed in a separate PR.

mehdibo:
I actually didn't change the things you mentioned... probably I messed up while cloning/forking the repo.

jim-parry:
oh dear, but that is the effect of your PR :(
If you fork/clone the repo, and then make a separate branch for a group of changes, you can then submit a PR to merge that branch into the develop branch of the repo.

mehdibo:
@jim-parry  It's okay I'll try again later, a question while you are here. I found a lot of dead links (404 not found and websites that are closed now) in the code. What should I do to them? should I remove them?

jim-parry:
That depends on what they are. Some might be removed, while others should have appropriate replacements dug up. If there are only a few, they could be mentioned/quesitoned in the initial comment for the PR. If there are a bunch, then I suggest one PR for the bulk of the changes, and a separate PR proposing replacements for the dead links.

PR #5507, opened 2018-05-23
MySQLI Server-side management of SSL encryption

romainrg:
Hello,

I think it's a good thing to have the possibility to choose server side encryption just by setting up the CI database.php `'encrypt' => TRUE,` but also keeping the base possibility to choose cert, cipher, etc.. (client-side encryption). Just by modifying some line of MySQLI driver, everything become possible!

Tested on all my projects, if I choose `'encrypt' => FALSE,` I have :

```
object(stdClass)[32]
  public 'Variable_name' => string 'Ssl_cipher' (length=10)
  public 'Value' => string '' (length=0)
```

Then if I choose `'encrypt' => TRUE,` I have :

```
object(stdClass)[34]
  public 'Variable_name' => string 'Ssl_cipher' (length=10)
  public 'Value' => string 'DHE-RSA-AES256-SHA' (length=18)
```

Available for further informations

Regards

CI ❤️

### DIFF
--- a/system/database/drivers/mysqli/mysqli_driver.php
+++ b/system/database/drivers/mysqli/mysqli_driver.php
@@ -158,7 +158,7 @@ class CI_DB_mysqli_driver extends CI_DB {
 			}
 		}
 
-		if (is_array($this->encrypt))
+		if (is_array($this->encrypt) or $this->encrypt === true)
 		{
 			$ssl = array();
 			empty($this->encrypt['ssl_key'])    OR $ssl['key']    = $this->encrypt['ssl_key'];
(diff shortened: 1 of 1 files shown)


narfbg:
Would you kindly explain how this works? I remember the originally implemented feature was full of quirks and I'm not confident of just merging something like that.

romainrg:
@narfbg By default in MySQL clients can supply a CA certificate matching the one used by the server and enable host name identity verification. In this way, the server and client place their trust in the same CA certificate and the client verifies that the host to which it connected is the one intended. If CA certificate is not present in client side, but SSL flag given at the moment of the MySQL connexion, PHP will etablish secured communication between client and server without verifying the identify of client (but server must be configured for) and actually it's not possible with this configuration.

Without this, it's very easy to sniff all TCP traffic..

Hope this will explain everything to you!

narfbg:
Not sure ... I understand you might mean that it's just a MySQL version of HTTPS.

But if that's the case, then disabling certificate verification is a bad idea.

romainrg:
@narfbg Sure but it's better to have encrypted traffic than nothing else ..

narfbg:
I don't disagree with that, but SSL isn't only about encryption.

What I'm saying is that `ssl_verify` should be enabled by default and being switchable. However, when you turn the array into a simple boolean, the switch disappears and there's no possibility to do that.

Also, what happens if the server doesn't support encryption? I'd assume silently continuing with unencrypted traffic, which isn't good.

narfbg:
Actually, that's my bad - I know we already do this, but assumed it wouldn't work for your feature.

But you're not addressing my comment on certificate verification.

PR #5510, opened 2018-05-25
Only use the output of idn_to_ascii() if it worked properly

johndekroon:
According to the manual, idn_to_ascii will return false if it encounters a problem. This happend on one of our systems (the function exists, but there was an issue in the underlaying libraries, which prevented that the function works). This resulted on that system in:

foo@example.com -> foo@[false] -> validate("foo@")

This will obviously fail. The solution is that when the function does not work properly, to treat it like it doesn't exists at all.

### DIFF
--- a/system/libraries/Email.php
+++ b/system/libraries/Email.php
@@ -1017,7 +1017,11 @@ class CI_Email {
 			$domain = defined('INTL_IDNA_VARIANT_UTS46')
 				? idn_to_ascii($matches[2], 0, INTL_IDNA_VARIANT_UTS46)
 				: idn_to_ascii($matches[2]);
-			$email = $matches[1].'@'.$domain;
+			//If idn_to_ascii() fails, treat it like it doesn't exists
+			if ($domain !== FALSE)
+			{
+				$email = $account.'@'.$domain;
+			}
 		}
 
 		return (bool) filter_var($email, FILTER_VALIDATE_EMAIL);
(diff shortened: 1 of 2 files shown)


narfbg:
Does it not trigger an error message when it fails?

narfbg:
Sorry to question that, but there's a known issue with this function with pre-4.3 (IIRC) versions of ICU and I *know* it raises a warning in that case, so ... double check with `error_reporting(E_ALL)` and `display_errors(1)` please?

johndekroon:
The intl.error_level is by default off. So by default there is no warning. 
Enabling it (set to E_ALL) yields the following error message:
Unknown error: idn_to_ascii(): idn_to_ascii: cannot convert to ASCII in Unknown on line 0

narfbg:
OK, that's not the same condition I was talking about.

Just one small modification is necessary to fit with our code style - add an empty line before the `if` block, and remove the comment (it's pretty self-explanatory).

johndekroon:
Alright, fixed that

system/libraries/Email.php
```diff
@@ -1017,7 +1017,11 @@ public function valid_email($email)
 			$domain = defined('INTL_IDNA_VARIANT_UTS46')
 				? idn_to_ascii($matches[2], 0, INTL_IDNA_VARIANT_UTS46)
 				: idn_to_ascii($matches[2]);
-			$email = $matches[1].'@'.$domain;
+
+			if ($domain !== FALSE)
+			{
+				$email = $account.'@'.$domain;
```
narfbg: Caught this during a merge conflict: `$account` doesn't exist here

I'm fixing it myself, but let this be a lesson - don't just copy-paste stuff. :)

PR #5511, opened 2018-05-26
Add support for multiple APPPATHS

alex2005git:
#### My scenario: I wanted to create a CMS.  
These changes are needed when creating a bundle like package.  
  
Each client will have a repo, so instead of having the CMS code and
framework in every clients' repo, I make the CMS a dependency, and the CMS
has codeigniter as a dependency.  
This way it's easy to update either the codeigniter system or the CMS code, as simple as a composer update (after changing the composer.json).
  
So we end up having client application(config, controllers, helpers, language, views, libraries, everything except core) -> cms application(all the good stuff) -> system.  
  
Without these changes I cant create custom models/controllers/configs or
anything else, per clients' demand, other than what can be extended by the CMS application.  
  
I've only tested it seeing if everything works, I haven't made any tests (my testing is kind of rusty).  
And from what I can see it works, I'm not sure if it breaks anything, I'm sure someone will let me know if that's the case.

### DIFF
--- a/index.php
+++ b/index.php
@@ -299,6 +299,15 @@ switch (ENVIRONMENT)
 	define('VIEWPATH', $view_folder.DIRECTORY_SEPARATOR);
 
 /*
+ * -----------------------------------------------------------------------------
+ *  Optionally add another APPPATH to your application, useful when
+ *  creating a package, and wanting to expand with new controllers/models/etc...
+ *  Except core classes, everything else can be added to this path
+ * -----------------------------------------------------------------------------
+ */
+	//$APPPATHS = array();
+
+/*
  * --------------------------------------------------------------------
  * LOAD THE BOOTSTRAP FILE
  * --------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


ivantcholakov:
The proposed change is big and I am concerned about breaking of third-party libraries (HMVC) and custom projects. If this idea gets accepted, let CI 3.1.x stay as it is.

gxgpet:
Why is this needed? CI already has [multiple applications](https://www.codeigniter.com/userguide3/general/managing_apps.html) support.

alex2005git:
@gxgpet because the multiple applications support you say it has, is different from what I want it to do.
I don't want each client to share codeigniter system, since they will be in different servers (possibly), what I describe in the description is not currently possible with the current implementation of the way CI loads files.

If anyone has ever coded in rails, it is very easy to bundle you app, then each client requires it, then you can overwrite the files or extend them in an easy way... This is what I want CI to do also, make it easier for bundled apps to be more easily extendable.

narfbg:
The things that you want to package as extensions, we consider application code, so we're clearly not on the same page here.

And @ivantcholakov's concerns are legitimate and not only to the extent he stated. This is bound to introduce suble bugs in unexpected places, with the benefit being very questionable and for only a handful of people.

Sorry, but CI is simply not built for the type of project you have.

PR #5512, opened 2018-05-26
mimes.php: 7z file types changes

gxgpet:
As this [Wikipedia article](https://en.wikipedia.org/wiki/7z) states, the 7-Zip generated archives have the `7z` extension and the MIME id `application/x-7z-compressed`.

### DIFF
--- a/application/config/mimes.php
+++ b/application/config/mimes.php
@@ -155,7 +155,7 @@ return array(
 	'ics'	=>	'text/calendar',
 	'ical'	=>	'text/calendar',
 	'zsh'	=>	'text/x-scriptzsh',
-	'7zip'	=>	array('application/x-compressed', 'application/x-zip-compressed', 'application/zip', 'multipart/x-zip'),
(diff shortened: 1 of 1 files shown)


narfbg:
That's cool but it should *add* '7z' not replace '7zip'.

gxgpet:
I know you are thinking about a potential BC break, I did too. But there won't be any: if we look [back in 2007](https://sourceforge.net/projects/sevenzip/files/7-Zip/4.57/), at the software itself, they were using `7z` archives, and not a `7zip`.

I don't really mind adding a new line instead of replacing, since I can't really tell what's the source of `7zip` extension in that file... but I'm pretty sure that if someone has a `7zip`, it's because of a renaming, not as 7-Zip direct output.

narfbg:
Yes, but renames do happen and we've had people complain about the weirdest BC breaks ever ... please just alter the patch.

gxgpet:
@narfbg You're right. Done in 8eeb1a8.

PR #5514, opened 2018-05-29
Solve error 2014 calling stored procedures mysqli

davicotico:
When calling multiple stored procedures, we can run into the following error: "Commands out of sync; you can't run this command now".
In MySQL, a procedure can return more than one result set, and it will always return one extra empty result set that carries some meta information on the procedure call itself, especially error information.
Only is fetched the results of the SELECT executed within the procedure though.
Source: 
https://bugs.mysql.com/bug.php?id=71044
https://dev.mysql.com/doc/refman/5.6/en/mysql-more-results.html
https://dev.mysql.com/doc/refman/8.0/en/commands-out-of-sync.html

### DIFF
--- a/system/database/drivers/mysqli/mysqli_driver.php
+++ b/system/database/drivers/mysqli/mysqli_driver.php
@@ -296,7 +296,12 @@ class CI_DB_mysqli_driver extends CI_DB {
 	 */
 	protected function _execute($sql)
 	{
-		return $this->conn_id->query($this->_prep_query($sql));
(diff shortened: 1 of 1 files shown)


narfbg:
The changes you're proposing don't match the (unclear) explanation you've given. Please elaborate.

narfbg:
That's not the clarification I was hoping for ... You just said the same thing but added 2 lines of code (and linked a random commit instead of a PR id).

PR #5516, opened 2018-05-30
Use Config::base_url() properly

wpyh:
Using the Config::slash_item() method, we get double forward slashes. For example, base_url = 'http://localhost:8080/' and asset is '/assets/my_asset.css', then the generated URL would be 'http://localhost:8080//assets/my_asset.css'.

This patch fixes that, so the generated URL would be 'http://localhost:8080/assets/my_asset.css'. There is actually already an analog example in the case if index_page === TRUE, where the site_url() function is used.

### DIFF
--- a/system/helpers/html_helper.php
+++ b/system/helpers/html_helper.php
@@ -200,7 +200,7 @@ if ( ! function_exists('img'))
 				}
 				else
 				{
-					$img .= ' src="'.get_instance()->config->slash_item('base_url').$v.'"';
+					$img .= ' src="'.get_instance()->config->base_url($v).'"';
 				}
 			}
 			else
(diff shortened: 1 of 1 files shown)


wpyh:
Hi, I think this would be suitable for the 3.1-stable branch as well. Should I make another PR for that branch? Or do I need to just create this one PR for both branches? Thanks.

jim-parry:
PRs are only proposed for the develop branch. We maintain the 3.1-stable branch separately.

wpyh:
I still don't really understand... please clarify, if I would like to see this patch incorporated into the next maintenance release of CodeIgniter (like 3.1.9), is this PR enough? Or do I need to do something else?

jim-parry:
Re build failures: travis is choking inconsistently on composer installs.

PR #5520, opened 2018-06-01
Set CSRF cookie empty if regenerate is activated

bsenyuva:
CSRF cookie was reusable till expire even regenerate is activated. It's a security vulnerability. A cookie should be use once if regenerate activated. Setting cookie to empty is guarantee to use a cookie only once.

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -238,6 +238,7 @@ class CI_Security {
 		if (config_item('csrf_regenerate'))
 		{
 			// Nothing should last forever
+			setcookie($this->_csrf_cookie_name);
 			unset($_COOKIE[$this->_csrf_cookie_name]);
 			$this->_csrf_hash = NULL;
 		}


narfbg:
There's 2 function calls immediately after the block you're changing - one to generate a new value, and another to send it.

So how does that make it reusable?

bsenyuva:
You're right, but older cookie is still in effect till expire.

You can regenerate the issue using "Replay XHR" feature of Chrome Dev Tools on a ajax post.

narfbg:
It's the *same* cookie name - you can't have 2 matching cookie names with different values being valid at the same time.

narfbg:
I get what you mean now, though I haven't tested it.

Whether you replay it with the old or new value - it will work; even past the expiry time, as long as you hard-code the cookie and form input value to be the same. The regeneration and expiry time only make sure that the client doesn't hold the value for too long.
It is simply the client that stores the value and nothing on the server side will change it.

Of course, if you trigger multiple AJAX requests with the same values at the same time, the value can be reused because it is already preset in the request queue with a fixed value, even after you receive the server response.

But this is only a vulnerability if a *third-party* is able to create those values and trigger a request on behalf of the client, which is impossible. The key here is that since only the client knows the value, a third-party can't create an input field that matches it.

bsenyuva:
It's possible via any XSS vulnerability. You can get csrf cookie value and reuse it. It's not impossible. A third party can create a replica of a request. It's not has to be same values but same csrf token. It's a vulnerability and the solution is just 36 chars of code :)

narfbg:
Look, if you need XSS to an obscure trigger an already hard to pull off CSRF replay, I don't consider this a vulnerability because that's almost like saying our CSRF protection as a whole is useless.

That being said, I don't mind hardening, regardless of how much code it adds. But what you're suggesting simply doesn't make sense - in the very same response that you want to send an empty (delete) cookie Set-Cookie header, it is followed by another one which sets a value. In some environments, that would even result in the empty one not being sent at all because it is overwritten before buffers are cleared.

The only way to convince me is with a PoC, and that means a ready-made script instead of suggesting I somehow replicate it on my own using specific tools (I don't use Chrome and I hate touching anything to do with the front-end anyway). No offsense, that's just how I am.

PR #5524, opened 2018-06-12
urldecode($slug)

imharrywu:
as $slug passed here for News::view($slug) is in url-encoded, in case of multi-bytes string, news_model->get_news($slug) will fail.


narfbg:
You *assume* it's URL-encoded, but it shouldn't be in any well-coded application.

PR #5525, opened 2018-06-12
Use multidimensional array as rule (field name)

edtsz:
Now it is possible validate something like this:
```JSON
{
    "name": "My custom template 00",
    "text": "Big custom text here",
    "vars": [
        {
            "name": "Birth",
            "position": 0,
            "step": 0
        },
        {
            "name": "Full name",
            "position": 1,
            "step": 0
        }
    ]
}
```

Or like this:
```HTML
<form action="" method="POST">
	<input name="name" value="My custom template 00"><br>
	<input name="text" value="Big custom text here"><br>
	<input name="vars[0][name]" value="Birth"><br>
	<input name="vars[0][position]" value="0"><br>
	<input name="vars[0][step]" value="0"><br>
	<input name="vars[1][name]" value="Full name"><br>
	<input name="vars[1][position]" value="1"><br>
	<input name="vars[1][step]" value="0"><br>
	<button type="submit">Send</button><br>
</form>
```

[The normal way to use array as field names](https://codeigniter.com/user_guide/libraries/form_validation.html#using-arrays-as-field-names)
```PHP
$this->form_validation->set_rules('options[]', 'Options', 'required');
```

The new way to use array as field name (the old one still working): 
```PHP
$validator->set_rules('var[][name]', 'Variable', 'required|min_length[4]|max_length[50]');
$validator->set_value('var[][name]');
```

Like this: https://forum.codeigniter.com/thread-63551-post-325434.html (also my problem)

### DIFF
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -221,7 +221,7 @@ class CI_Form_validation {
 		if (($is_array = (bool) preg_match_all('/\[(.*?)\]/', $field, $matches)) === TRUE)
 		{
 			sscanf($field, '%[^[][', $indexes[0]);
-			$indexes = array_merge($indexes, $matches[1]);
+			$indexes = array_merge($indexes, $matches[1][$i]);
 		}
 
 		// Build our master array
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but no. Supporting single-dimensional arrays is enough of a hassle already and that's where the line is drawn.

You can define a closure/callback/etc. rule to validate further sub-levels of arrays.

PR #5529, opened 2018-06-20
Add Input library functionality.

Chofoteddy:
In order to help enrich the functionality of CodeIgniter, I propose the implementation of the following methods; that will make it easier for the programmer to know the type of method that is used by the client.

```php
if ($this->input->is_post())
{
    // logic of update
}

$this->load->view('/example/form');
```

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -637,9 +637,93 @@ class CI_Input {
 			? strtoupper($this->server('REQUEST_METHOD'))
 			: strtolower($this->server('REQUEST_METHOD'));
 	}
+	
+	// ------------------------------------------------------------------------
+	
+	/**
+	 * Returns whether this is a GET request
+	 * 
+	 * @return bool
+	 */
+	public function is_get()
+	{
+		return $this->method(TRUE) === 'GET';
+	}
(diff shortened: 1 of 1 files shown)


vardash0131:
i think it can be useful in any case ...

ivantcholakov:
```php
if ($this->input->method() == 'post')
{
    // logic of update
}
```

Chofoteddy:
@ivantcholakov Yes, it is the current solution, the proposal can be a semantic improvement. In passing, it is useful to reduce the expression used to validate that action.

Chofoteddy:
@ivantcholakov @narfbg 

The most powerful frameworks have semantic improvement methods that are very helpful.

Does not CodeIgniter pretend to be like that? (helpful)

> CodeIgniter is a powerful PHP framework with a very small footprint, built for developers who need a simple and elegant toolkit to create full-featured web applications.

Reference:
- https://api.symfony.com/master/Symfony/Component/HttpFoundation/Request.html#method_isMethod
- https://www.yiiframework.com/doc/api/2.0/yii-web-request#getIsPost()-detail

PR #5531, opened 2018-06-22
Input: considering nested array keys in `get_post` and `post_get`

tianhe1986:
With URL such as `http://ci.mine.cn/test/test_get?a[b]=test` and the following code
```
$something = $this->input->get('a[b]');
$something2 = $this->input->get_post('a[b]');
var_dump($something, $something2);
```

`$something` gets 'test', and `$something2` turns out to be `NULL`.  However, they are expected to be the same.

So I try to fix it using null comparison instead of `isset` check.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -212,9 +212,8 @@ class CI_Input {
 	 */
 	public function post_get($index, $xss_clean = FALSE)
 	{
-		return isset($_POST[$index])
-			? $this->post($index, $xss_clean)
-			: $this->get($index, $xss_clean);
+		$output = $this->post($index, $xss_clean);
+		return ($output !== NULL) ? $output : $this->get($index, $xss_clean);
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


narfbg:
An `isset` check **is** a null comparison (that's not where the difference in results come from), and I insist that `isset()` still be used.

tianhe1986:
@narfbg No, it is just `isset` causing the difference, because in `get()` it doing more check than just null comparison.
With the URL I showed above, `isset($_GET['a[b]'])` would return false, resulting in `get_post` fallback to `post()` and get `NULL` at last.
It should fallback to array notation check with `$_GET` first, and then fallback to `post()` when still getting `NULL`.

tianhe1986:
I see. I'm sorry for misunderstanding what you said.
However, `?:` is not right for variables such as `0`, `'0'` and empty string for that it is a `TRUE` check， so `isset()` should also be used.

```
$output = $this->post($index, $xss_clean);
return isset($output) ? $output : $this->get($index, $xss_clean);
```
Is this OK?

narfbg:
You're right, I always mix up short ternary with null coalesce. :)

Yes, that'd be exactly what I wanted.

Issue #4461, opened 2016-02-15
wrong detected mobile browser with user_agent library - MALGJS;


widyou:
I used user_agent library for mobile browser like this:

```
// check mobile browser
$this->load->library('user_agent');`
$result = $this->agent->is_mobile();
```

but IE11 browser in LG OEM devices with Windows 10 agent value like this:
`Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; MALGJS; rv:11.0) like Gecko`

'MALGJS' means LG OEM Windows Devices. (like notebook)

in User_agent.php -> _set_mobile():

```
if (FALSE !== (stripos($this->agent, $key))) // stripos('... MALGJS; ...', 'lg')
{
        $this->is_mobile = TRUE;
        $this->mobile = $val;
        return TRUE;
}
```

then some IE in LG notebook redirect mobile page.

anovsiradj:
you make pull request here: https://github.com/bcit-ci/CodeIgniter/edit/develop/application/config/user_agents.php

flexsingit:
![firefox_wallpaper](https://cloud.githubusercontent.com/assets/25687998/25117512/1c2c5430-2430-11e7-86e6-7b48aab9bc8d.png)

PR #5536, opened 2018-06-25
Update User Agents

elminson:
'MALGJS' means LG OEM Windows Devices. (like a notebook) with Windows 10 but detected as mobile.
but IE11 browser in LG OEM devices with Windows 10 agent value like this:
Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; MALGJS; rv:11.0) like Gecko

### DIFF
--- a/application/config/user_agents.php
+++ b/application/config/user_agents.php
@@ -88,7 +88,8 @@ $browsers = array(
 	'IBrowse'		=> 'IBrowse',
 	'Maxthon'		=> 'Maxthon',
 	'Ubuntu'		=> 'Ubuntu Web Browser',
-	'Vivaldi'		=> 'Vivaldi'
+	'Vivaldi'		=> 'Vivaldi',
+	'MALGJS'		=> 'LG OEM Windows Devices'
 );
 
 $mobiles = array(


elminson:
Yes @gxgpet

elminson:
Can you check this PR @gxgpet?

gxgpet:
Well, there are some things here regarding this...

The reason that `user_agents.php` file is located under `application/config` it's because it's a _configuration_ file, meaning its aim is to be edited by the CI developer, if necessary. So, if you spot something which is targetting your audience, then there you go, you change this file.

It is also true that `user_agents.php` tries to add as many scenarios as possible. Yet, it should maintain a generic purpose list, not tailored to some specific cases, and to be honest, "LG OEM Windows Devices" looks a bit like a particular scenario, not a generic one.

However, since this is the second time this issue is reported, there could be some reflection for deciding whether or not this should be added.

Anyway, you should wait for some official feedback (mine isn't) here, for this PR.

narfbg:
Well, technically this isn't the "second time this issue is reported", but rather @elminson trying to resolve #4461 (and only with the info available in that issue, I would say).

"LG OEM Windows Devices" is not a browser brand name, so listing it under `$browsers` is definitely erroneous. At best it belongs under `$platforms` or `$mobiles`, but even then it would stand out as quite dissimilar compared to the types of other names listed.

PR #5540, opened 2018-07-02
Recursive translate_uri_dashes on each segment

tazoony:
Recursive translate_uri_dashes on each segment of the URL.
Because when your controllers are in folders, sub-folders, sub-sub-folders, etc., the translate_uri_dashes property is only applied on 2 firsts segments.


narfbg:
This feature exists because PHP class and method names cannot contain dashes, and it is only applied to those. File-system directories don't have such a limitation and don't need this.

AllenChong:
pull request

Issue #5382, opened 2018-01-17
Bad session key get_lock design


sskaje:
In both CI3 and CI4, memcached and redis session handler implementations, the session key lock designing are incorrect.

Take redis for example.

[CI4](https://github.com/bcit-ci/CodeIgniter4/blob/develop/system/Session/Handlers/RedisHandler.php), code were old, buggy but more compatible.
CI4 pseudo code inside loop is:
```
    if redis::ttl(lock_key) > 0 then 
        sleep 1s 
        continue loop

    redis::set(lock_key, ttl)
```

Yes, two concurrent requests would both bypass *ttl > 0* check, and then get the lock.
Concurrent session writing may be corrupted, but [1]concurrent session reading, or [2]one writing and one or more reading are working.

Since commit 569cf69 of [CI3](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Session/drivers/Session_redis_driver.php) , introduced from #5170 :
```
    if redis::ttl(lock_key) > 0 then 
        sleep 1s 
        continue loop

    if redis::ttl(lock_key) == key-not-exists then 
        redis::setNx(lock_key, ttl)
    else 
        redis::set(lock_key, ttl)
```
This does not help but gets more buggy.
Same as CI4 code, both requests can bypass *ttl > 0* and both hit ttl === -2 which is key not exists.
Request 1 hit redis::setNx and then success, but request 2 hit and then failed because 1 has made the key existed. Now request 2 wouldn't get session data. 

The solution is just simplified the code.

Redis:
```php

    protected function _get_lock($session_id)
    {
        // PHP 7 reuses the SessionHandler object on regeneration,
        // so we need to check here if the lock key is for the
        // correct session ID.
        if ($this->_lock_key === $this->_key_prefix.$session_id.':lock')
        {
            log_message('error', 'SESSION Redis _get_lock');
            return $this->_redis->setTimeout($this->_lock_key, 300);
        }

        // 30 attempts to obtain a lock, in case another request already has it
        $lock_key = $this->_key_prefix.$session_id.':lock';
        $attempt = 0;
        do
        {
            $result = $this->_redis->set($lock_key, time(), array('nx', 'ex'=>300));

            if (! $result)
            {
                sleep(1);
                continue;
            }

            $this->_lock_key = $lock_key;
            break;
        }
        while (++$attempt < 30);

        if ($attempt === 30) {
            // in case someone added a key without expiration
[...]

PR #5548, opened 2018-07-24
Fix redis session

kionez:
I'm trying to avoid the session lock issue with concurrent\parallel request (i.e.: multiple ajax calls): #5382 

If two concurrent request A and B try a setnx at the same time, the second will fail giving the "Session: Error while trying to obtain lock for..." error. So, I'm trying to test a different solution: if the ttl of a lock key is -2 (so lock doesn't exists) try to set with expiration (setnx) but if it fails, sleep a little bit and retry from start and fail only if there is >30 attempts. I'm using usleep to reduce waiting time.

I'm using rand() to avoid concurrency, but I don't think is a correct way.

(insipred from: https://redis.io/commands/setnx#design-pattern-locking-with-codesetnxcode )

### DIFF
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -360,17 +360,20 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 				continue;
 			}
 
-			if ($ttl === -2) {															   // key doesn't exists
-				$result = $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300)); // set key, if it doesn't exist, with a ttl of 300s
-				if (!$result) {															   // not set because already set? wait and retry
-					$sleep_random = (rand(1,5)) * 100000;								   // https://redis.io/commands/setnx#design-pattern-locking-with-codesetnxcode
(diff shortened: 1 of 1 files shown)


narfbg:
When I said that you shouldn't rely on `rand()`, I was suggesting that you should alter your patch, not submitting it while knowing it's not ok ...

But even with that aside, I don't see how this is a better solution than what @sskaje suggested in #5382.

narfbg:
@kionez Well, there was a potentially good solution in #5382, but the author tried to sneak in other changes that we don't want and wasn't cooperative. You're welcome to pick that one up.

narfbg:
@AllenChong Can you please stop with these random comments?

kionez:
I kept only the redis session related changes from the #5382, and you said it was not different but you give me the advice to pick that one up. 
I do not understand what I should do, the code I've upload in this PR seems working fine on my projects and this is enough for me, I hope that CodeIgniter team will fix this annoying issue in future releases, in the meanwhile I'll keep my patch.

PS: Sorry I don't speak english very well, therefore there are many misunderstandings.

narfbg:
I said this wasn't *better* than #5382.

    different != better

PR #5550, opened 2018-07-24
Fixed problem with session crash

dkowalskidev:
Site froze (100% CPU) when user open more than 2 pages in one moment in other tabs in browser. This happened after adding the method php5_validate_id (exactly validateId) in 3.1.9 My fix resolved this problem.

### DIFF
--- a/system/libraries/Session/drivers/Session_files_driver.php
+++ b/system/libraries/Session/drivers/Session_files_driver.php
@@ -206,16 +206,7 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 			rewind($this->_file_handle);
 		}
 
-		$session_data = '';
-		for ($read = 0, $length = filesize($this->_file_path.$session_id); $read < $length; $read += self::strlen($buffer))
-		{
-			if (($buffer = fread($this->_file_handle, $length - $read)) === FALSE)
-			{
-				break;
-			}
-
-			$session_data .= $buffer;
-		}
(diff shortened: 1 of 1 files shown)


system/libraries/Session/drivers/Session_files_driver.php
```diff
@@ -206,16 +206,7 @@ public function read($session_id)
...
-		for ($read = 0, $length = filesize($this->_file_path.$session_id); $read < $length; $read += self::strlen($buffer))
-		{
-			if (($buffer = fread($this->_file_handle, $length - $read)) === FALSE)
-			{
-				break;
-			}
-
-			$session_data .= $buffer;
-		}
+		$session_data = file_get_contents($this->_file_path.$session_id);
```
mrliptontea: Not sure this is the cause of the issue but anyway it won't work due to file being locked few lines above.
You can use `stream_get_contents` instead, benefit is it will also re-use already opened file handle:

```php
$session_data = stream_get_contents($this->_file_handle);
```

narfbg:
The code that you're changing has been working just fine for a long, long time. It is completely illogical to change it when the problem root cause isn't in there ... you're treating a symptom.

And @mrliptontea is correct on one of the reasons why `file_get_contents()` is a bad idea there, but there's more than one. It's not like we want to have a ton of code that basically does `file_get_contents()`, `file_put_contents()`, `unlink()`, etc. There are little details in there that you're completely ignoring.

PR #5552, opened 2018-07-26
Update DB_result.php

basit-adhi:
Adding public variable $dropdown_result_array (Dropdown Result Array, use @form_dropdown)
Adding public result_dropdown(), row_dropdown_array(), unbuffered_row()
Changing some function to accomodate  $dropdown_result_array: num_rows(), result(), row()

### DIFF
--- a/system/database/DB_result.php
+++ b/system/database/DB_result.php
@@ -79,6 +79,14 @@ class CI_DB_result {
 	public $dropdown_result_array		= array();
 
 	/**
+	 * Dropdown Result Array, use @form_dropdown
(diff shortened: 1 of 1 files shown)


narfbg:
Even if it was a desirable feature (which it isn't), this is a classic SRP violation.

basit-adhi:
Any suggestion? A new library?

narfbg:
A suggestion for what? How to do this in your apps? That's up to you, it just doesn't belong in CI core.

basit-adhi:
Ok, sorry for that. I make my own application/libraries then. Thank you

PR #5554, opened 2018-07-27
Language improvement

shawe:
If configured language file is not available, before to show and error try a last time looking for english version of this file.


narfbg:
Sorry, but no.

This just hides an underlying problem, making it harder to find.

Issue #2755, opened 2013-12-09
Table names with commas.


xmartinj:
Our SQL Server ERP software was installed with a table prefix that includes a comma. CI's Active Record automatically splits the FROM clause based on commas assuming they are separate tables ignoring quotes/brackets to keep the table name together.

i.e. "[Allen,LLC$Customer]" gets changed to "[Allen, LLC$Customer]" which fails.

p.s. The poor choice of table prefix is something we would like to change at the next opportunity, but yet this seems something the FROM function should handle better.

xmgcoyi:
MySQL
Database and table names cannot contain “/”, “\”, “.”, or characters that are not permitted in file names.
http://dev.mysql.com/doc/refman/5.0/en/identifiers.html

ivantcholakov:
@xmartinj
The closed issue #3764 gives an idea for a workaround. If you pass the selected fields as an array what would happen now?

alex7071:
If mysql accepts dots in schema names and fixes bugs related to it, why can't you? https://bugs.mysql.com/bug.php?id=61078 This IS a bug, as you're enforcing something that is not enforced by MySql. Dot names in schema name works fine everywhere else. This is why using third party software sucks.

narfbg:
If MySQL can do it, why can't we? Really?

MySQL is a huge project with countless contributors, a powerful corporation behind it, and on top of that only has to deal with itself.  
CodeIgniter is a 3MB framework that sparingly gets small contributions from outsiders, is largely maintained by me alone, in my spare time, and it touches on pretty much all elements of the web technology stack.

Even isolating the issue to databases alone - we try to maintain both compatibility and ease of use between more than 10 database platforms. And the issue isn't even closed yet.

That's why.

xmartinj:
No worries. We've moved to Laravel, and are moving away from this database
as well.

alex7071:
After 4 years since this has been reported without any attention, anyone can presume this "feature" will never be implemented, even though, it is a bug. Your "holier-than-thou" attitude doesn't do much to convince me you listen to your user-base, so you can abstain from snarky comments. I'm sure you're full of them.

> If MySQL can do it, why can't we? Really?
> 
> MySQL is a huge project with countless contributors, a powerful corporation behind it, and on top of that only has to deal with itself.
[...]

PR #5556, opened 2018-07-29
allow for escaping database and field names with non-standard characters

keplerg:
Signed-off-by: Kepler Gelotte <kepler@neighborwebmaster.com>

This pull request addresses feature request: Table names with commas. #2755

You can now escape database, table, and field names with the escape character(s) before submitting a request. For example:

```php
// Mysql example
$this->db->insert('`weird table name`', array('id' => 1, '`funky, name`' => 'some data'));
```

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -1341,7 +1341,7 @@ abstract class CI_DB_driver {
 			return $item;
 		}
 		// Avoid breaking functions and literal values inside queries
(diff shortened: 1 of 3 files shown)


narfbg:
This only works on MySQL (yet you've patched non-MySQL parts of the DB classes) and it doesn't really work as well as you think it does ... many parts of the code rely on very specific regular expressions that would still break.

If it was that easy, we'd have done this years ago.

PR #5576, opened 2018-08-26
Add MYSQLI_TYPE_NEWDECIMAL constant

1e90ff:


narfbg:
I'm very sorry for not responding to this earlier.

If you're still up for it, there's only a small issue to resolve - what's the difference between DECIMAL and NEWDECIMAL, and then why should both return 'decimal' as the type?

PR #5580, opened 2018-08-31
Fix insert_id() method in PDO to actually return integer

nazar-pc:


narfbg:
Not all IDs are integers ...

nazar-pc:
PhpDoc claims the opposite and this already caused big problems for me.

nazar-pc:
> Well, the phpdoc blocks are there more for an easily readable code and aren't always 100% accurate

If that is an official position of CI's developers, than it is A HORRIBLE one and you should seriously reconsider you point of view! PhpDoc sections, if present, should be either 100% accurate or fixed immediately as soon as someone points at inconsistencies.

IDEs use PhpDoc as source of truth and assist developers in writing WORKING code. In this case code that used mysqli driver and then switched to pdo (because rewritten mysqli driver leaks memory since version 3.0.0 when using persistent connections, which is a whole different story) STOPPED WORKING despite being 100% correct according to PhpDocs.

The issue here is not that `insert_id()` might return something other than an integer, but rather that for pdo it NEVER returns an integer, it only returns a string. And if people use IDEs and read documentation (which any good developer should) they don't expect anything other than an integer from `insert_id()` call.

narfbg:
You know, I could write an entire screen of text in reply to what you said, but the bottom line is I suggested a resolution and you opted for that tirade instead.

Have a good life.

PR #5581, opened 2018-09-01
Activate DB_driver reconnect method

yidas:
For common DB driver usage, this default reconnecting flow could support all drivers such as PDO.


narfbg:
That's not the same as the ping functionality provided by some drivers ... If a forced close/connect is what you want, then just do it with the 2 method calls.

Issue #5577, opened 2018-08-28
drivers/Cache_redis.php Call to a member function set() on null


Hanchengluo:
CI  V3.1.9
$config['host'] = '127.0.0.1';
$config['password'] = NULL;
$config['port'] = 6379;
$config['timeout'] = 0;	
$this->load->driver('cache');
$this->cache->redis->save('foo', 'bar', 10);

Fatal error: Call to a member function set() on null in D:\Vhosts\catv\system\libraries\Cache\drivers\Cache_redis.php on line 193

A PHP Error was encountered
Severity: Error
Message: Call to a member function set() on null
Filename: drivers/Cache_redis.php

Line Number: 193

fifthsegment:
This is most likely the php redis extension not loaded. CodeIgniter should have thrown an exception immediately when you loaded the cache driver, rather than allowing you to call the save function. I've pushed some changes that should fix that.

narfbg:
Please follow #5562.

ravikantgyanwave:
Getting this error:-

Type: Error

Message: Call to a member function sRemove() on null
Filename: C:\wamp64\www\services\system\libraries\Cache\drivers\Cache_redis.php
Line Number: 196

Backtrace:

C:\wamp64\www\services\application\controllers\Redis.php
Line: 6
Function: save

C:\wamp64\www\services\index.php
Line: 315
Function: require_once

Please help!

azuajece:
i have the same problems

PR #5586, opened 2018-09-05
Added exception when redis extn is not loaded

fifthsegment:
Fixes: https://github.com/bcit-ci/CodeIgniter/issues/5577 . Outputs a meaningful exception when the redis extension wasn't found and prevents the redis cache library from being loaded, instead of allowing the user to proceed loading the redis library without failure.


narfbg:
There's an easier way to do that - simply removing the existing `try ... catch` block.

Also, this won't work - anything after a `throw` is dead code.

PR #5602, opened 2018-09-19
Comma before "and" grammar fix

morlo1:
Comma is usually not being used before "and"


Tpojka:
It's called [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma).

jim-parry:
CodeIgniter is not trying to follow "Oxford English".
The closest would be "American English", as an international web convention.
The closest reference I can find is https://www.grammarly.com/blog/comma/, which suggests that such a comma is optional, and debatable, regardless of the "convention" one is adhering to.
This feels a bit too picky to get excited about!

Tpojka:
@jim-parry If answered to me, I was just pointing that it is not inappropriate how it was already written but rather as you say more debatable should use or shouldn't. I wasn't referencing to Oxford grammar rules per se to be used in CodeIgniter. Btw. thanks for link. :)

jim-parry:
@Tpojka Not to worry. My comment was directed more at the original poster, and explaining why the PR was not accepted.

PR #5604, opened 2018-09-21
ftp_no_source_file specifies local file name

MrWeb:
`ftp_no_source_file` returns a general "*Unable to locate the source file. Please check your path*" error making it difficult to see where the problem is.
The method has been updated to show the file is throwing the error so that it returns "*Unable to locate the source file. Please check your path* **(your/local/filename)**"

### DIFF
--- a/system/libraries/Ftp.php
+++ b/system/libraries/Ftp.php
@@ -663,7 +663,7 @@ class CI_FTP {
 		$CI =& get_instance();
 		$CI->lang->load('ftp');
 		$extra_message = ($file) ? " ($file)" : "";		
-		show_error($CI->lang->line($line).$file);
+		show_error($CI->lang->line($line).$extra_message);
 	}
 
 }
--- a/system/libraries/Ftp.php
+++ b/system/libraries/Ftp.php
@@ -309,7 +309,7 @@ class CI_FTP {
 
 		if ( ! file_exists($locpath))
 		{
-			$this->_error('ftp_no_source_file');
+			$this->_error('ftp_no_source_file', $locpath);
 			return FALSE;
 		}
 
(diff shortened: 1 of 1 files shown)


narfbg:
I've been sitting on this for a month and still can't decide on it.

On one hand easier debugging isn't a bad thing, but on the other I don't feel like this is worth it and this library doesn't really belong today. @jim-parry What do you think?

jim-parry:
I have no problem with this, regardless of whether the library belongs nowadays :)

narfbg:
I was rather looking for you to make the call on this, but I guess that's a "yes".

system/libraries/Ftp.php
```diff
@@ -655,13 +655,15 @@ public function close()
 	 * Display error message
 	 *
 	 * @param	string	$line
+	 * @param	string	$file
 	 * @return	void
 	 */
-	protected function _error($line)
+	protected function _error($line, $file = false)
```
narfbg: While you're handling only a single occurence right now, `$file` isn't a good name in terms of absraction. Some variation of `$extra` or `$var` would be better.

Also, default to `NULL` instead of `FALSE` (and do type that in all capitals, as per our [styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html)).

system/libraries/Ftp.php
```diff
@@ -655,13 +655,15 @@ public function close()
...
 	{
 		$CI =& get_instance();
 		$CI->lang->load('ftp');
-		show_error($CI->lang->line($line));
+		$extra_message = ($file) ? " ($file)" : "";
```
narfbg: [Styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html) again (please read it) - variables injected into strings should be written inside curly braces.

And you don't need a new variable for this, just modify the input one.

PR #5606, opened 2018-09-26
reset transaction depth before rollback call

mcorne:
The transaction depth must be reset before calling the rollback in case an exception is triggered during the rollback.


narfbg:
Nope. You're completely ignoring the intended behavior.

PR #5614, opened 2018-10-02
Update Readme.txt

BalajiJBcs:
fix the grammar error in the documentation.

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -3,7 +3,7 @@ What is CodeIgniter
 ###################
 
 CodeIgniter is an Application Development Framework - a toolkit - for people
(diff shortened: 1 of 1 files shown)


jim-parry:
- websites/web sites is debatable ... https://en.wikipedia.org/wiki/Website (American vs English)
- commas are debatable - that came up a couple of weeks ago, though in the context of a multi-item list. Your correction does look correct.
- acknowledgement/acknowledgment - both acceptable (English vs American)

I personally think it better to just stick with what we have. I suspect there are similar examples throughout the docs, and we don't need to be so picky.

PR #5640, opened 2018-11-23
Ability to delete DB cache files by sql

manojdotwork:
If the same query is used in multiple pages, CI DB cache will create multiple similar files in multiple folders. These can be deleted with this patch.

### DIFF
--- a/system/database/DB_cache.php
+++ b/system/database/DB_cache.php
@@ -206,6 +206,32 @@ class CI_DB_Cache {
 		delete_files($dir_path, TRUE);
 	}
 
+	/**
+	 * Deletes the cache files completely by sql query
+	 *
+	 * @param	string	$sql
+	 * @return	void
+	 */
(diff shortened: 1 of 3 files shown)


narfbg:
This is quite arbitrary and either limited in usage (to cases where there's only a single query per page) or inconvenient (in the case where you have to make N calls, where N is the unknown number of queries executed on every page).

`delete_all()` already exists and covers the cases where clearing per-page cache isn't sufficient.

manojdotwork:
using delete_all(), why should I clear all the cache, if I know the query which I'm caching. That will hit db again. Anyways, I dont see any issue, as user can know what to use and when to use. I'm currently facing an issue with this, so guess I'm not only the one.

narfbg:
> I'm currently facing an issue for not having this function, so guess I'm not only the one.

Completely unfounded speculation.

Issue #5641, opened 2018-11-24
Expanding the functionality of $this->db->like() and it's related methods without breaking anything.


Nudasoft:
Feature request 01
============
Now $this->db->like() method and it's related methods, e.i: this->db->or_like() and etc... support for 3 different wildcard (%) placements. I'm asking for another extra one. lets call it 'none' for now.

The expected result should be when using 'none' wildcard placement option:
$this->db->like('title', 'match', 'none'); 
Should produces: WHERE `title` LIKE 'match' ESCAPE '!'

Reason to request this feature
----------------------------------
MySQL return rows that seemingly do not match the WHERE clause. Please refer this url to get a good idea about what I'm taling here:
[https://stackoverflow.com/questions/15715719/why-does-mysql-return-rows-that-seemingly-do-not-match-the-where-clause](https://stackoverflow.com/questions/15715719/why-does-mysql-return-rows-that-seemingly-do-not-match-the-where-clause)

By using LIKE operator without wildcard character we can get the exact result set that we want.

This is how I do it for now. but dont know if this way of doing this intorduce any security issues. 
$this->db->where('title LIKE ', 'match);

And I think that introducing extra option like 'none' will not break anthing or affect anyone who use codeigniter in their projects.

Feature request 02
============
Currently $this->db-like() and it's related methods using 'WHERE' clause to generate the query. for example:

$this->db->like('title', 'match', 'before');
Will produce: WHERE `title` LIKE '%match' ESCAPE '!'

I'm asking for 'HAVING' clause option for $this->db-like() and it's related methods as a another optional parameter (4th parameter)

for example:
$this->db->like('title', 'match', 'before', 'having');
Should produce: HAVING `title` LIKE '%match' ESCAPE '!'

Reason to request this feature
----------------------------------
'WHERE' clause is not support for aggregate functions like COUNT().

If I need performe search against a aliased field, we have to concatinate text strings here and their to get the job done. may be doing that can introduce security issues.

This is how I do it now:
$this->db->having('title LIKE ', '%match%'); 

But I wish I had a clean way of achieving this without concatinate wildard character around.

narfbg:
01:

> MySQL return rows that seemingly do not match the WHERE clause. Please refer this url to get a good idea about what I'm taling here:
> https://stackoverflow.com/questions/15715719/why-does-mysql-return-rows-that-seemingly-do-not-match-the-where-clause

[...]

PR #5643, opened 2018-11-27
Add optional parameter to like(), or_like(), not_like() and or_not_like() methods.

nalakapws:
This PR address the 2nd feature request from this #5641 issue.

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -924,152 +924,175 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	// --------------------------------------------------------------------
 
 	/**
-	 * LIKE
-	 *
-	 * Generates a %LIKE% portion of the query.
-	 * Separates multiple calls with 'AND'.
-	 *
-	 * @param	mixed	$field
-	 * @param	string	$match
-	 * @param	string	$side
-	 * @param	bool	$escape
-	 * @return	CI_DB_query_builder
-	 */
-	public function like($field, $match = '', $side = 'both', $escape = NULL)
-	{
-		return $this->_like($field, $match, 'AND ', $side, '', $escape);
-	}
+     * LIKE
+     *
+     * Generates a %LIKE% portion of the query.
+     * Separates multiple calls with 'AND'.
+     *
+     * @param	mixed	$field
+     * @param	string	$match
+     * @param	string	$side
+     * @param	bool	$escape
+     * @param	string	$clause
+     * @return	CI_DB_query_builder
+     */
+    public function like($field, $match = '', $side = 'both', $escape = NULL, $clause = 'where')
+    {
+        return $this->_like($field, $match, 'AND ', $side, '', $escape, $clause);
+    }
 
-	// --------------------------------------------------------------------
+    // --------------------------------------------------------------------
 
-	/**
-	 * NOT LIKE
-	 *
-	 * Generates a NOT LIKE portion of the query.
-	 * Separates multiple calls with 'AND'.
-	 *
-	 * @param	mixed	$field
-	 * @param	string	$match
-	 * @param	string	$side
-	 * @param	bool	$escape
-	 * @return	CI_DB_query_builder
-	 */
-	public function not_like($field, $match = '', $side = 'both', $escape = NULL)
-	{
-		return $this->_like($field, $match, 'AND ', $side, 'NOT', $escape);
-	}
+    /**
+     * NOT LIKE
+     *
+     * Generates a NOT LIKE portion of the query.
+     * Separates multiple calls with 'AND'.
+     *
+     * @param	mixed	$field
+     * @param	string	$match
+     * @param	string	$side
+     * @param	bool	$escape
+     * @param	string	$clause
+     * @return	CI_DB_query_builder
+     */
+    public function not_like($field, $match = '', $side = 'both', $escape = NULL, $clause = 'where')
+    {
+        return $this->_like($field, $match, 'AND ', $side, 'NOT', $escape, $clause);
+    }
 
-	// --------------------------------------------------------------------
+    // --------------------------------------------------------------------
 
-	/**
-	 * OR LIKE
-	 *
-	 * Generates a %LIKE% portion of the query.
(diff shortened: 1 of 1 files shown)


narfbg:
This is parameter creep ... It may look inconsistent, but a better option would be to just have `having_like()`, `having_not_like()`, etc.

And you must follow the [styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html).

Nudasoft:
Is there anything I have to do in order to make this PR get merged?

narfbg:
Yes, there are still some spaces left in on lines 1100, 1171 (shown in diff with darker green).

And there's more significant stuff code-wise, but it will probably result in a lot of back and forth if I asked you to do all of it, so I can make those changes myself after merging. It's decent enough as it is, I just want less repetition.

system/database/DB_query_builder.php
```diff
@@ -999,24 +999,105 @@ public function or_not_like($field, $match = '', $side = 'both', $escape = NULL)
...
+	 * Generates a %LIKE% portion of the query.
+	 * Separates multiple calls with 'AND'.
+	 *
+	 * @param	mixed	$field
+	 * @param	string	$match
+	 * @param	string	$side
+	 * @param	bool	$escape
+	 * @return	CI_DB_query_builder
+	 */
+	public function having_like($field, $match = '', $side = 'both', $escape = NULL)
```
NallDev: this line doesn't match PSR-2 rules

system/database/DB_query_builder.php
```diff
@@ -999,24 +999,105 @@ public function or_not_like($field, $match = '', $side = 'both', $escape = NULL)
...
+	{
+		return $this->_like($field, $match, 'AND ', $side, '', $escape, 'having');
+	}
+
+	// --------------------------------------------------------------------
```
NallDev: bad comment

system/database/DB_query_builder.php
```diff
@@ -999,24 +999,105 @@ public function or_not_like($field, $match = '', $side = 'both', $escape = NULL)
...
+	 * Generates a NOT LIKE portion of the query.
+	 * Separates multiple calls with 'AND'.
+	 *
+	 * @param	mixed	$field
+	 * @param	string	$match
+	 * @param	string	$side
+	 * @param	bool	$escape
+	 * @return	CI_DB_query_builder
+	 */
+	public function not_having_like($field, $match = '', $side = 'both', $escape = NULL)
```
NallDev: this line doesn't match PSR-2 rules

system/database/DB_query_builder.php
```diff
@@ -999,24 +999,105 @@ public function or_not_like($field, $match = '', $side = 'both', $escape = NULL)
...
+	 * Generates a %LIKE% portion of the query.
+	 * Separates multiple calls with 'OR'.
+	 *
+	 * @param	mixed	$field
+	 * @param	string	$match
+	 * @param	string	$side
+	 * @param	bool	$escape
+	 * @return	CI_DB_query_builder
+	 */
+	public function or_having_like($field, $match = '', $side = 'both', $escape = NULL)
```
NallDev: this line doesn't match PSR-2 rules

system/database/DB_query_builder.php
```diff
@@ -999,24 +999,105 @@ public function or_not_like($field, $match = '', $side = 'both', $escape = NULL)
...
+	 * Generates a NOT LIKE portion of the query.
+	 * Separates multiple calls with 'OR'.
+	 *
+	 * @param	mixed	$field
+	 * @param	string	$match
+	 * @param	string	$side
+	 * @param	bool	$escape
+	 * @return	CI_DB_query_builder
+	 */
+	public function or_not_having_like($field, $match = '', $side = 'both', $escape = NULL)
```
NallDev: this line doesn't match PSR-2 rules

system/database/DB_query_builder.php
```diff
@@ -1164,6 +1263,97 @@ public function group_end()
...
 	// --------------------------------------------------------------------
 
+	/**
+	 * Starts a query group for HAVING clause.
+	 *
+	 * @param	string	$not	(Internal use only)
+	 * @param	string	$type	(Internal use only)
+	 * @return	CI_DB_query_builder
+	 */
+	public function having_group_start($not = '', $type = 'AND ')
```
NallDev: this line doesn't match PSR-2 rules

system/database/DB_query_builder.php
```diff
@@ -1164,6 +1263,97 @@ public function group_end()
...
+	{
+		$type = $this->_group_get_type($type);
+
+		$this->qb_where_group_started = TRUE;
+		$prefix = (count($this->qb_having) === 0 && count($this->qb_cache_having) === 0) ? '' : $type;
+		$having = array(
```
NallDev: You can use `[]`

error56 (changes requested):
All the code doesn't match PSR-2 rules

system/database/DB_query_builder.php
narfbg: We don't subscribe to PSR-2.

system/database/DB_query_builder.php
narfbg: It's not a "bad comment"; we have these in between all functions.

system/database/DB_query_builder.php
narfbg: ... and I'd force him to change it back to `array()`, because that's what we use.

narfbg:
Yes, all of CodeIgniter doesn't match PSR-2 rules. We have our own [styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html), referenced in the [contributing guide](https://github.com/bcit-ci/CodeIgniter/blob/develop/contributing.md), which you should read before reviewing others' work.

system/database/DB_query_builder.php
narfbg, line 1033: This should be `having_not_like` instead.
Same thing applies to all other methods below - `not` should always be after `having` in the name.

narfbg (changes requested):

PR #5651, opened 2018-12-09
Test if cache file is writable before unlink it

pawelkg:
Make sure that file exists when unlinking it. Concurrent process run on same time may unlink in the meantime this file.


narfbg:
Did that actually happen?

pawelkg:
Yes. I had this issue few time. When I boot up my development server (Vagrant), there is outdated cache and I run two tabs at start, I have this error on one of tabs. Fact is that in real production cases that is highly unlikely, but still, I can happen if all condition are valid.

PR #5652, opened 2018-12-09
Add px to width and height for inline style

Khuthaily:
The image's width and height are always more than 0; so, adding "px" after the width and height is necessary for HTML validation.

### DIFF
--- a/system/helpers/captcha_helper.php
+++ b/system/helpers/captcha_helper.php
@@ -352,7 +352,7 @@ if ( ! function_exists('create_captcha'))
 			$img_src = 'data:image/png;base64,'.base64_encode($img_src);
 		}
 
-		$img = '<img '.($img_id === '' ? '' : 'id="'.$img_id.'"').' src="'.$img_src.'" style="width: '.$img_width.'; height: '.$img_height .'; border: 0;" alt="'.$img_alt.'" />';
+		$img = '<img '.($img_id === '' ? '' : 'id="'.$img_id.'"').' src="'.$img_src.'" style="width: '.$img_width.'px; height: '.$img_height .'px; border: 0;" alt="'.$img_alt.'" />';
 		ImageDestroy($im);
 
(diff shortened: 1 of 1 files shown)


narfbg:
HTML validation? This is CSS ...

And to be completely honest, I have no idea what the official specs say about this. Can you point us to the relevant part?

Khuthaily:
> HTML validation? This is CSS ...
> 

**Example of Current `$img` Value:**
`alt-text-here`
**Result:** Fails HTML Validation

**Pull request adds `px` to `width `and `height`:**
`alt-text-here`
**Result:** Passes HTML Validation

**Explanation:** If the HTML attributes "width" and "height" are used with the "img" tag, there is no need for "px". Since the width and height are added using the "style" attribute, a unit needs to be declared (px, in this case).

> And to be completely honest, I have no idea what the official specs say about this. Can you point us to the relevant part?

I will come back to this later; but anyone can add their two-cents ;)

pawelkg:
This is CSS not HTML matter. 'width' (https://www.w3.org/wiki/CSS/Properties/width) and 'height' (https://www.w3.org/wiki/CSS/Properties/height) property, in this case, use 'length' value: https://www.w3.org/TR/CSS2/syndata.html#value-def-length: 
> The format of a length value (denoted by <length> in this specification) is a <number> (with or without a decimal point) immediately followed by a unit identifier (e.g., px, em, etc.). After a zero length, the unit identifier is optional.

In this case image width and height will not be 0, so we should unit identifier.

Khuthaily:
@narfbg, below are two screen shots from the book "Web Standards: Mastering HTML5, CSS3, and XML", showing that 
> attribute value has the same syntax as the contents of a CSS declaration block except that the delimiting braces are omitted

html-1
html-0

narfbg:
@Khuthaily Please, please, don't ever post a screenshot of text ever again, to anyone. It's one of the most annoying things ever. And in this case quite useless as well.

PR #5654, opened 2018-12-12
Fixes docblock and small code fixes.

pawelkg:
Fixes for some small docblock issues and small fixes for code.

### DIFF
--- a/system/helpers/html_helper.php
+++ b/system/helpers/html_helper.php
@@ -370,7 +370,7 @@ if ( ! function_exists('meta'))
 			{
 				if ($meta['type'] === 'equiv')
 				{
-					$meta['type'] === 'http-equiv';
+					$meta['type'] = 'http-equiv';
 				}
 				elseif ( ! in_array($meta['type'], $allowed_types, TRUE))
 				{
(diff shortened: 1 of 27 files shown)


system/helpers/html_helper.php
```diff
@@ -370,7 +370,7 @@ function meta($name = '', $content = '', $type = 'name', $newline = "\n")
 			{
 				if ($meta['type'] === 'equiv')
 				{
-					$meta['type'] === 'http-equiv';
+					$meta['type'] = 'http-equiv';
```
narfbg: This is the only actual fix and valuable change, out of everything you've done in 27 files.

But I can't accept the entire thing just for this single line ... This is why it's a bad idea to submit large PRs with changes unrelated to one another.

pawelkg: OK. I can prepare separate PR, just for this one line. By the way, can You explain me what exactly is wrong with correction in docblocks? I do not clearly understand, why other changes in this PR can't be merged.

narfbg: Well, I didn't say that all of them are wrong. Some are mergeable, but they still don't belong together with an unrelated bugfix.

Some are simply redundant (e.g. inferred through type declarations in code), spacing looks weird on others, and we simply don't use the `type|another_type` notation.

Also, those `break`s that you wanted to add aren't omitted by mistake.

PR #5655, opened 2018-12-15
add string _ci_prepare_view_vars verification

Zhucola:
if params $vars is string，now is verification by function "_ci_prepare_view_vars"

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -526,10 +526,10 @@ class CI_Loader {
 	 */
 	public function vars($vars, $val = '')
 	{
-		$vars = is_string($vars)
-			? array($vars => $val)
-			: $this->_ci_prepare_view_vars($vars);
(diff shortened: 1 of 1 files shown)


Zhucola:
example:

$this->load->vars("_ci_path","abc");
$this->load->view("test");

narfbg:
Sorry, but I have no idea what you're trying to say or do.

Zhucola:
> Sorry, but I have no idea what you're trying to say or do.



> Sorry, but I have no idea what you're trying to say or do.


public function vars($vars, $val = ''){
	$vars = is_string($vars)
		? array($vars => $val)
		: $this->_ci_prepare_view_vars($vars);
	foreach ($vars as $key => $val){
		$this->_ci_cached_vars[$key] = $val;
	}

	return $this;
}

PR #5663, opened 2018-12-28
Update news_section.rst

hansdez:
Fix a mistake in the URL viewing the individual slug from the index.


hansdez:
My mistake. I tried it before I changed the route file.

PR #5664, opened 2018-12-28
use local.php values in get_config helper function

Siphon098:
This looks for a file at `APPPATH.'config/local.php'` and loads the values, if found. An example of where this can currently be a problem can be found at:

https://github.com/Siphon098/CodeIgniter/blob/develop/system/core/Security.php#L275-L283

The creation of the cookie may fail even when `cookie_domain` is valid within the `local.php` config.

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -260,6 +260,12 @@ if ( ! function_exists('get_config'))
 				exit(3); // EXIT_CONFIG
 			}
 
+            // Do we have a local.php configuration file?
+            if (file_exists($file_path = APPPATH.'config/local.php'))
+            {
+                require($file_path);
+            }
+
 			// Does the $config array exist in the file?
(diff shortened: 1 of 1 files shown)


narfbg:
We already have a way more flexible, environment-based configuration system.

Siphon098:
Perhaps this should have been an issue before a pull request. The current issue is that the environment-based configurations would require each developer working on a project to have their own environment config, which causes tons of headaches in managing updates to the environment files. We use a local.php that is setup in the autoloader config to load person-specific values such as domains, endpoints and third-party credentials for developing (that does not get committed to Git), but that autoloader has no effect on the get_config function. We can set it up that way, but it's less than ideal due to the way we need the cascading values to work on 3 different levels without duplication within files.

Having that 3rd level on the get_config, or the ability to add custom configs would be a lifesaver.

PR #5668, opened 2019-01-03
Remove NULL check from protected _wh_in function

derekyle:
The NULL check that has been removed was quietly skipping the where_in statement if a NULL was passed in the $value parameter. By not checking for NULL, this allows an inadvertent NULL to get pushed into the query built by the query builder and avoid any accidental updates or selects of the entire table. I couldn't find any issues where this might cause a problem. The _wh protected function that builds the 'where' portions of a query does not seem to have a NULL check like it's where_in counterpart.


narfbg:
You're also removing the NULL check for `$key` though ... This would be fine if you also removed the `= NULL` input defaults.

PR #5678, opened 2019-01-18
Update user_agents.php

Lablnet:
Added Meizu  agent in Manufacturers 

https://www.meizu.com/


application/config/user_agents.php
```diff
@@ -152,7 +152,8 @@
 	'wii'			=> 'Nintendo Wii',
 	'open web'		=> 'Open Web',
 	'openweb'		=> 'OpenWeb',
-
+	'meizu'                 => 'Meizu',
+
```
narfbg: There's whitespace here that needs to be removed.

Issue #5648, opened 2018-12-05
Feature Request - Customize the log file name


BrandonNoad:
Hi,

Is it possible to add a config setting to customize the name of the log file?

In the `write_log` method (Log.php), you can customize the log file's directory and extension, but the `'log-'.date('Y-m-d')` portion of `$filepath` cannot be changed.

 Thanks!

BrandonNoad:
We are making some changes to our dev environment and need the log filename to be static.

shahrushabh1996:
Search this line on system/core/log.php just replace 'log-'.date('Y-m-d').' Replace this name with your custom name.

$filepath = $this->_log_path.'log-'.date('Y-m-d').'.'.$this->_file_ext;

narfbg:
@BrandonNoad That doesn't really answer my question. What do you mean by static?

@shahrushabh1996 Please don't suggest manual edits of the stock framework code. And especially not here - this is our official repository, not a random forum board for people who want quick hacks.

BrandonNoad:
Instead of having a new log file each day, we want to have a single log file, which is actually a symlink that points to /dev/stderr. 

Ideally, we could log directly to stderr. But I don't think that is possible, so this is a workaround.

narfbg:
I have to say, I don't think that's a particularly good idea.

To be honest, when I asked why you'd need the feature I expected you to say something like per-vhost logfiles. And I've slowly convinced myself that we do need such a feature, so this does go into the TODO list.

BrandonNoad:
Thanks.

We are moving our project to Kubernetes/Docker, and this feature will help with that.

Cheers.

Lablnet:
https://github.com/bcit-ci/CodeIgniter/pull/5679

jonsmith1982:
@narfbg I might have a crack at this, what do you propose to do with $config['log_file_extension'] to avoid it not adding to $filepath making it FALSE? and checking against it, if its not then adding .php?

jonsmith1982:
[patch-#5648.txt](https://github.com/bcit-ci/CodeIgniter/files/2802784/patch-.5648.txt)

@narfbg This should do the trick

jonsmith1982:
[issue-#5648.patch.txt](https://github.com/bcit-ci/CodeIgniter/files/2802907/issue-.5648.patch.txt)

@narfbg Right, did it properly this time. Would I be able to push these changes myself or is that restricted?

shyamjos:
[...]

PR #5679, opened 2019-01-18
Update Log.php

Lablnet:
https://github.com/bcit-ci/CodeIgniter/issues/5648

### DIFF
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -111,6 +111,12 @@ class CI_Log {
 	 */
 	protected static $func_overload;
 
+	/**
+	 * Custum file name
+	 *
+	 * @var	string
+	 */
+	protected $_c_file;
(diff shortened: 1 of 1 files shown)


system/core/Log.php
```diff
@@ -158,6 +165,18 @@ public function __construct()
...
 	}
 
+	/**
+	 * Set the custum file
+	 *
+	 * @param	string	$file name of file
+	 *
+	 * @return	void
+	 */
+	public function setCustumFile($file)
```
dmamills: `setCustomFile`

system/core/Log.php
```diff
@@ -158,6 +165,18 @@ public function __construct()
 		}
 	}
 
+	/**
+	 * Set the custum file
```
dmamills: typo

PR #5684, opened 2019-01-28
fixes #5648

### LINKED ISSUE #5648: Feature Request - Customize the log file name

jonsmith1982:
Signed-off-by: jonsmith1982  - Sorry, I sent a patch to fix this in the original post. Just getting used to git and how best to do things.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -228,6 +228,17 @@ $config['log_path'] = '';
 
 /*
 |--------------------------------------------------------------------------
+| Error Logging FILENAME
+|--------------------------------------------------------------------------
+|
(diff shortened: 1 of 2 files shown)


narfbg:
Hmm, I had something different in mind; i.e. incorporating the whole thing into `log_path` with the possibility to parse date formats, but this is decent too. Not sure right now.

But you did raise a good point in #5648 about `log_file_extension` and I don't think leaving it as is would be OK.

PR #5699, opened 2019-02-21
Added unbuffered_get() for fetch larger dataset from MySQL

giriannamalai:
Added unbuffered_get() in Query Builder for MySQL. Default get() with unbufferd_row() is giving memory allocation error for larger dataset. Simply, Its not working.
So added _execute_unbuffered() method in Mysqli_driver.php and it execute the query with MYSQLI_USE_RESULT flag.
So the cursor will be returned, with that we can use unbuffered_row() to fetch row by row to prevent the memory alocation issue.
Ex:
$i = 0;
$qry = $this->db->select("id")->from("table")->unbuffered_get();
while ($row = $qry->unbuffered_row('array')) {
	$i+=1;
}
print($i);

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -569,14 +569,12 @@ abstract class CI_DB_driver {
 	 * @param	string	$sql
 	 * @param	array	$binds = FALSE		An array of binding data
 	 * @param	bool	$return_object = NULL
-	 * @param	bool	$_is_unbuffered = FALSE		Set the query execution as buffered or unbuffered
+	 * @param	bool	$is_unbuffered = FALSE		Set the query execution as buffered or unbuffered
 	 * @return	mixed
 	 */
-	public function query($sql, $binds = FALSE, $return_object = NULL, $_is_unbuffered=FALSE)
+	public function query($sql, $binds = FALSE, $return_object = NULL, $is_unbuffered=FALSE)
 	{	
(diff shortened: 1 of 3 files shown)


narfbg:
All other considerations aside, we don't do MySQL-only features.

giriannamalai:
Ok. Let's add this for MysQL, PgSQL and Sqlite..!

narfbg:
That's not a good idea in the first place. Even if it works for you now, you'll still reach the memory limit with a larger dataset and you'll have the same problem, so this feature isn't a solution.

The whole point of `unbuffered_row()` is that you fetch a row, use it immediately and then discard it (thus freeing memory as soon as possible), so that you always have more memory available. It's the only technique that guarantees success with large datasets and it won't work if you try to fetch the entire result set at once.

giriannamalai:
> That's not a good idea in the first place. Even if it works for you now, you'll still reach the memory limit with a larger dataset and you'll have the same problem, so this feature isn't a solution.
> 
> The whole point of `unbuffered_row()` is that you fetch a row, use it immediately and then discard it (thus freeing memory as soon as possible), so that you always have more memory available. It's the only technique that guarantees success with large datasets and it won't work if you try to fetch the entire result set at once.

Yes. When we fetch and write a CSV large dataset, this will be completely useful. Fetched records will discarded once we write into file.

PR #5704, opened 2019-02-28
Update form_textarea function form_helper.php

kuwayasu:
When $extra has 'cols' or 'rows', $default should be changed. Now these extra attributes are ignored.


narfbg:
Why would you put those in `$extra` instead of `$data` (as is intended)? And what if `$extra` is a string? Also not complying with the [styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html).

PR #5708, opened 2019-03-02
Resolve race condition in redis driven session key get_lock

mchobbylong:
This PR tries to provide a better design of session key get_lock design, mentioned in #5382 and by @sskaje .

The previous design may fail when multiple threads are trying the access the same session at nearly the same time. Most of the threads will be killed.

Although this behavior may be corrected by adding checks in custom CI controllers or custom CI libraries, in my understanding session operations in CI controllers should never fail because of one issue come from the driver. In other words, race conditions should be handled in the driver.

BTW, I have completely read that issue and related discussions. As for >300s / no-TTL problem, every lock is going to have TTL, except someone tries to operate sessions by himself, not by the driver IMO.
If this does happen, then he totally deserves the long time waiting for the lock releases.

### DIFF
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -389,7 +389,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 				sleep(1);
 				continue;
 			}
-			elseif ($ttl === -1 && ! $this->_redis->setex($lock_key, 300, time()))
+			elseif ( ! $this->_redis->setex($lock_key, 300, time()))
 			{
 				log_message('error', 'Session: Error while trying to obtain lock for '.$this->_key_prefix.$session_id);
 				return FALSE;
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -389,8 +389,9 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 
 			if ( ! $result)
 			{
-				log_message('error', 'Session: Error while trying to obtain lock for '.$this->_key_prefix.$session_id);
-				return FALSE;
+				// Sleep for 0.1s to wait for lock releases.
+				usleep(100000);
+				continue;
 			}
 
 			$this->_lock_key = $lock_key;
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -389,7 +389,8 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 				sleep(1);
 				continue;
 			}
-			elseif ($ttl === -1 && ! $this->_redis->setex($lock_key, 300, time())) {
+			elseif ($ttl === -1 && ! $this->_redis->setex($lock_key, 300, time()))
+			{
 				log_message('error', 'Session: Error while trying to obtain lock for '.$this->_key_prefix.$session_id);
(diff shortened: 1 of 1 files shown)


narfbg:
As I just said in #5382, I can't rely on anything in that thread anymore. If you want me to merge this, you'll have to completely detatch yourself from that issue and explain the problem and solution completely from the start here, and unfortunately I'll have to question every argument again, even if I already did previously.

mchobbylong:
You're right about that thread. Much thanks for your time on that mess.

Now I will try to make this PR clear.

The problem happens when **clients make multiple AJAX requests**. When multiple threads are accessing the same session at nearly the same time, only ONE of them will get the session_lock. Others will be killed.

This is because, according to the code in `Session_redis_driver.php` at line 386~394:
```php
$result = ($ttl === -2)
	? $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300))
	: $this->_redis->setex($lock_key, 300, time());
if ( ! $result)
{
	log_message('error', 'Session: Error while trying to obtain lock for '.$this->_key_prefix.$session_id);
	return FALSE;
}
```
When multiple threads come here, only one of them will successfully execute `setnx`. Others then will be terminated.

The point is, every session request *should* not get an immediate failure when attempts to get the lock, but wait for the lock releases. So, instead of `return FALSE`, let's just make this request wait for a short while and have a try again.

narfbg:
OK, that makes sense. However, isn't the concurrency problem only triggered by `set()` (when `$ttl` is `-2`)? Why should we still retry if `setex()` fails?

In addition, you sleep for 0.1 seconds (and each attempt is still counted as a full one up to the max of 30), while other wait-to-release-lock scenarios sleep for a full second.

mchobbylong:
@narfbg Updated according to your suggestions 😉 

> In addition, you sleep for 0.1 seconds (and each attempt is still counted as a full one up to the max of 30), while other wait-to-release-lock scenarios sleep for a full second.

Initially I sleep for only 0.1s just because it will retry more quickly then shorten the overall response time when handling multiple AJAX requests. On my test case, sleeping for 0.1s will be nearly 50% faster when executing `ab -n 1000 -c 20`.

But I just realize that this sleep-period configuration may vary due to different workload. Maybe sleeping 1s is more "common" :-)

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -383,11 +383,13 @@ protected function _get_lock($session_id)
...
-				: $this->_redis->setex($lock_key, 300, time());
-
-			if ( ! $result)
+			if ($ttl === -2 && ! $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300)))
+			{
+				// Sleep for 1s to wait for lock releases.
+				sleep(1);
+				continue;
+			}
+			elseif ($ttl === -1 && ! $this->_redis->setex($lock_key, 300, time()))
```
narfbg: The `$ttl === -1` condition should go away, it changes logic ... Even if it's correct, it's redundant at best.

mchobbylong:
@narfbg Corrected. That is surely my mistake.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Corrected.

mchobbylong:
It seems that travis-ci has reported a failure which is not truly related to my commit. Wonder how to fix it...

narfbg:
Yes, Travis sometimes errors with silly stuff like failures to download Composer or something ... I restarted that particular job to fix it, but in general you don't need to worry about that.

Thanks for solving what has been a nightmare of a problem for so long!

PR #5712, opened 2019-03-08
Adding pt-br language

bersuc:
Adding pt-br language to system. Get the info from english version and translated to Brazilian portuguese.

Need to check special characters and / or add encoding.


jim-parry:
translations should be submitted to https://github.com/bcit-ci/codeigniter3-translations, and it already has Portugese-Brazilian, thanks.

PR #5716, opened 2019-03-12
Corrected example mistake

achraf52:
Corrected an example mistake where the input script tries to echo `<p>fieldset content here</p>\n` but the result in "Produces" is `<p>form content here</p>` which is clearly an inconsistant result.

### DIFF
--- a/user_guide_src/source/helpers/form_helper.rst
+++ b/user_guide_src/source/helpers/form_helper.rst
@@ -380,7 +380,7 @@ The following functions are available:
 
 				<fieldset>
 					<legend>Address Information</legend>
-						<p>form content here</p>
+						<p>fieldset content here</p>
 				</fieldset>
 		*/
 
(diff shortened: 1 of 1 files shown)


jim-parry:
A "fieldset" is an HTML element, which groups several form elements together. The original text looks correct.

achraf52:
@jim-parry You clearly misunderstood what I'm trying to correct. I'm well aware what the "fieldset" element is, the mistake is not in the element itself, but the example text (mentioned above) which is NOT THE SAME as the PHP script that produced it.

Source script:

```
$attributes = array(
        'id'    => 'address_info',
        'class' => 'address_info'
);

echo form_fieldset('Address Information', $attributes);
echo "<p>fieldset content here</p>\n";
echo form_fieldset_close();
```

Incorrect result:

```
<fieldset id="address_info" class="address_info">
	<legend>Address Information</legend>
	<p>form content here</p>
</fieldset>
```

Correct result:

```
<fieldset id="address_info" class="address_info">
	<legend>Address Information</legend>
	<p>fieldset content here</p>
</fieldset>
```

If you still didn't catch it, use a diff tool to compare.

jim-parry:
Aha - I see the difference now! Sorry!!

PR #5618, opened 2018-10-11
Add new date validation rules to form_validation library

ChadTiffin:
This PR includes 3 new Form_validation rules:

- valid_date
- date_is_past
- date_is_future

Rule reference in docs are updated as well

PR #5060, opened 2017-03-20
More form validation Options

gahse:
I have added two validation methods to validate date and time for a given or default format

PR #5723, opened 2019-03-16
Validate dates in form_validation

Goodnato:
Now in _form_validation_ it is possible **to validate dates** using the **valid_date** method in set_rules(), this is amazing!

Using one or more parameters, the date validation will be based on the formats that the developer reports. Check out this [commit ](https://github.com/bcit-ci/CodeIgniter/pull/5723/commits/0f7e2cde187b69d354a617ff7e8ac778c68d41d8)**_Form_validation_test_**

### DIFF
--- a/system/language/english/form_validation_lang.php
+++ b/system/language/english/form_validation_lang.php
@@ -44,6 +44,7 @@ $lang['form_validation_valid_emails']		= 'The {field} field must contain all val
 $lang['form_validation_valid_url']		= 'The {field} field must contain a valid URL.';
 $lang['form_validation_valid_ip']		= 'The {field} field must contain a valid IP.';
 $lang['form_validation_valid_mac']		= 'The {field} field must contain a valid MAC.';
+$lang['form_validation_valid_date']		= 'The {field} field must contain a valid DATE.';
 $lang['form_validation_valid_base64']		= 'The {field} field must contain a valid Base64 string.';
 $lang['form_validation_min_length']		= 'The {field} field must be at least {param} characters in length.';
 $lang['form_validation_max_length']		= 'The {field} field cannot exceed {param} characters in length.';
--- a/system/libraries/Form_validation.php
+++ b/system/libraries/Form_validation.php
@@ -1327,7 +1327,8 @@ class CI_Form_validation {
 	{
                 if (strpos($val, ',') === FALSE)
 		{
-                        if($val !== "week"){
+                        if($val !== "week")
+                        {
                                 return DateTime::createFromFormat($val, $str)
                                         ? (DateTime::createFromFormat($val, $str)->format($val) === $str)
                                         : FALSE;
@@ -1346,14 +1347,15 @@ class CI_Form_validation {
                                 catch (Exception $ex) 
                                 {
                                         return FALSE;
-                               }
+                                }
                         }
 		}
 
 		foreach (explode(',', $val) as $format)
 		{
-			if(DateTime::createFromFormat($format, $str) && DateTime::createFromFormat($format, $str)->format($format) === $str){
-                            return TRUE;
+			if(DateTime::createFromFormat($format, $str) && DateTime::createFromFormat($format, $str)->format($format) === $str)
+                        {
+                                return TRUE;
                         }
 		}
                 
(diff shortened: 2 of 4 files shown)


Goodnato:
@narfbg Hey dude, im very happy because its my first time here. I love codeigniter and i love your work here with the others dev. Do you like the form_validation method? Its a good idea? I'm waiting your answer 🗡

narfbg:
Yes, I know you're waiting and I've responded to other PRs/issues in that time - sorry about that.

This one just needs more a thorough review, including going back through the multiple times it's been proposed in different forms already (you probably should do that too) ... and rejected. I'll get back to it when I have more than just a few spare minutes.

Goodnato:
I'm really happy! You are awesome. I'll be here all the time ready to help. My family will know this day that I helped codeigniter. See u soon :D

narfbg:
You're getting a bit too excited, prematurely ... I said it needs a thorough review, I didn't promise it will be accepted.

Anyway, these are the past PRs proposing a *valid_date* rule: #4346, #5060, #5618 

The main problem has always been that validating dates isn't nearly as easy as it seems at first glance. Difference between them and yours, is that they've been too naive while you've put much more effort into covering edge cases. But it's still too ... messy.

Here are the main problems to solve for this to have a chance of getting merged:

- Flip the arguments; first param is always the value to validate and you've inverted that order
- Make the code (way more) easier to read/understand; one shouldn't have to look at the test cases in order to get a clue of what's going on
- Make it dates-only; no hours, minutes, seconds, unix timestamps, etc ... Should probably blacklist month *names* and thus support only numbers. I wouldn't even bother with supporting weeks, but if you can provide good argumentation for it - maybe it's ok

That last one will involve a whitelist of format tokens, but here's a tip for after you (presumably) solve that - use `check_date()`, don't rely on `DateTime::createFromFormat()`.

After that, it will almost certainly need some polishing and naturally documentation, but the main roadblock is the above.

Goodnato:
I am very honored to see your work and Jim thank you very much for everything, you guys are incredible.

I apologize if I am giving work, but this validation has always been difficult to really work for the whole community (much discussion on this subject)

- I really apologize for having placed the order of the wrong arguments, thank you for alerting me and for your patience to verify.

- Once again you forgive me for not commenting on the tasks, in which case I am very ashamed

- I'm not just going to leave the date, the main idea is to follow the HTML5 standards (calm, I'll explain myself, I've seen all the discussions on the subject and I'll say why mine will work).

My function will work because if the developer really wants to receive the 1995 June 28, informs in the parameter valid_date [Y M d], it does not need lists. 
Or he wants to know a date and time to generate a report, he puts the default HTML5 as a parameter and a default for his region. What matters most is the date being valid and not which region the date is `valid_date [Y-m-d H:i,d/m/Y H:i]`. In the form can be `type="date"` or `type="text"`, the developer chooses the formats that will be accepted. That's amazing will work.

The only defect of `DateTime::createFromFormat` is that it does not say that this example is false 2019-02-31. What it does is add up the days that return 2019-03-02. That is why our problems are solved by comparing the value that came from the form 2019-02-31 === 2019-03-02 FALSE! Stay calm with this topic or tell me what else can give trouble (nothing in this world is perfect, is it?)

Goodnato:
Regarding the week, how are we working to have this date validation, let's make it complete? Since there is a `type="week"` attribute, we will not leave a developer sad because it will be the only attribute that will not have validation.

narfbg:
> My function will work because if the developer really wants to receive the 1995 June 28, informs in the parameter `valid_date[Y M d]`, it does not need lists.

What? This doesn't say anything.

> Or he wants to know a date and time to generate a report, he puts the default HTML5 as a parameter and a default for his region. What matters most is the date being valid and not which region the date is `valid_date[Y-m-d H:i,d/m/Y H:i]`. In the form can be `type="date"` or `type="text"`, the developer chooses the formats that will be accepted. That's amazing will work.

I don't care about HTML5, regions or form input types. I care that `valid_date` shoult only accept dates.

> The only defect of `DateTime::createFromFormat` is that it does not say that this example is false 2019-02-31. What it does is add up the days that return 2019-03-02. That is why our problems are solved by comparing the value that came from the form 2019-02-31 === 2019-03-02 FALSE! Stay calm with this topic or tell me what else can give trouble (nothing in this world is perfect, is it?)

`check_date()` is perfect ...

PR #5729, opened 2019-03-28
NOT BETWEEN syntax error DB where() issue solved

emrulmax2:
While creating a SQL query string using query builder,  NOT BETWEEN operator adds a syntax error, Solution added to _has_operator() and _get_operator() method.

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -1476,7 +1476,7 @@ abstract class CI_DB_driver {
 	 */
 	protected function _has_operator($str)
 	{
-		return (bool) preg_match('/(<|>|!|=|\sIS NULL|\sIS NOT NULL|\sEXISTS|\sBETWEEN|\s NOT BETWEEN|\sLIKE|\sIN\s*\(|\s)/i', trim($str));
+		return (bool) preg_match('/(<|>|!|=|\sIS NULL|\sIS NOT NULL|\sEXISTS|\sBETWEEN|\sNOT BETWEEN|\sLIKE|\sIN\s*\(|\s)/i', trim($str));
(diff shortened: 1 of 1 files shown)


emrulmax2:
Thank you for your accpetance, Hope it will merge with the branch because it will help others to avoid the unnecessary syntax error in query building process.

system/database/DB_driver.php
```diff
@@ -1476,7 +1476,7 @@ protected function _update($table, $values)
 	 */
 	protected function _has_operator($str)
 	{
-		return (bool) preg_match('/(<|>|!|=|\sIS NULL|\sIS NOT NULL|\sEXISTS|\sBETWEEN|\sLIKE|\sIN\s*\(|\s)/i', trim($str));
+		return (bool) preg_match('/(<|>|!|=|\sIS NULL|\sIS NOT NULL|\sEXISTS|\sBETWEEN|\sNOT BETWEEN|\sLIKE|\sIN\s*\(|\s)/i', trim($str));
```
tianhe1986: No need to modify this function.

If it has 'NOT BETWEEN' in `$str`, it must already has 'BETWEEN' and can be matched .

emrulmax2: Ok, you are right _has_operator() method doesn't need the 'NOT BETWEEN' match. but _get_operator() method must need to add the change. otherwise syntax query error shows.

narfbg:
Not that I can't imagine how a syntax error would be produced, but still - would you provide an example of when that happens? It's common sense to demonstrate a problem before trying to solve it ... :)

@tianhe1986 Thanks for the spot-on review.

PR #5730, opened 2019-03-29
Handle 404 errors in Router

JavaFluids:
We can move 404 errors handling to the Router class in order to gain in code readability as 404 errors handling can be the responsability of the Router.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -328,121 +328,22 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 
 /*
  * ------------------------------------------------------
- *  Sanity checks
+ *  Handle 404 errors
  * ------------------------------------------------------
  *
- *  The Router class has already validated the request,
- *  leaving us with 3 options here:
- *
- *	1) an empty class name, if we reached the default
- *	   controller, but it didn't exist;
- *	2) a query string which doesn't go through a
- *	   file_exists() check
- *	3) a regular request for a non-existing page
- *
- *  We handle all of these as a 404 error.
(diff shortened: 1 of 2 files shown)


narfbg:
So, we can move some code from outside class X to inside it, because we can ... Sure, we can, but why should we?

JavaFluids:
I propose this change for three reasons :

#### Increase the readability of the CodeIgniter.php file :

CodeIgniter.php is pretty easy to read and allow a quick understanding of the application flow, however 404 errors handling take an important part the class.
Here is a quick pastebin ( https://pastebin.com/LhUi4eUK ), 404 errors handling is from line 100 to 175 so we have (75 / 191) * 100 = 39.2% of the code.

#### Decrease in code complexity :

The current code handling 404 errors can be divided in two parts : the first one check if a controller for the request exists, then the second one check if a custom 404 controller exists.
I propose to divide these two parts in two separate functions in the Router component to get shorter units of code, respectively controller_exists() and handle_404_override().

Also I noticed a strange behaviour with the sharing of the '$e404' boolean variable between these two parts.
In the current code we have $e404 = FALSE when a custom 404 controller is loaded which can be confusing since we are still loading a 404 error page.
I replaced this $e404 boolean respectively with '$controller_exists' and '$overriden_404' boolean variables in controller_exists() and handle_404_override() functions.

#### Gain in encapsulation and cohesion : 

All the routing responsibilities are grouped in the Router component.

Let me know what you think regarding this proposal

narfbg:
Eh ... Sorry, but really you're just moving stuff around.

PR #5732, opened 2019-03-31
Code style fix

error56:


narfbg:
We don't want or need this. Strike [two](https://github.com/bcit-ci/CodeIgniter/pull/5731).

PR #5739, opened 2019-04-14
Stop using a group to manage serialised items

### LINKED ISSUE #5738: Redis driver performance issues
CodeIgnter tries to keep a cache of all objects it knows is serialized and handle it PHP side. This causes 2 issues.

1 - On Redis this is a o(1) operation, which is very quick, but populating the array makes it o(n) so large amounts of keys in this array slowly slow the application down. 

2- This value is stored in another cache key with a TTL of 0, and slowly grows over time too due to naturally expiring keys not losing their corresponding cache array.

Under a default instance of redis issue 2 the key will grow large enough to stop redis working, but if you change the eviction policy to `allkeys-lru` to make redis behave a little more like memcache then the large cache key can be evicted leaving everything in cache effectively broken as it returns serialized strings rather then the arrays / objects expected 

Just wondering what the thoughts are on removing the the `_ci_redis_serialized`. An alternative option would be using `$this->_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);` - I assume this wasnt used due to suspected performance issues?

Could use IGBinary but that has issues with incr and decr which would mean doing that client side, which could be breaking (splitting an atomic operation) ?

Happy to submit a PR with that I have now which is basically just adding the serliize option for all data types but having the redis library handle it rather than doing it manually

exussum12:
fixes #5738

When using a group to manage which objects should be serialised the
performance of the application suffers with larger amounts of keys. We
originally moved this to be a set / check on each key use (eg first get
the item, then check if its a member of the group). This made things
much faster when managing lots of keys.

We also use the non default but often used eviction policy allkeys-lru
Having this setting on meant that it is possible for redis to remove
this item and when this occurs the driver returns strings for
everything.

There is a very slight overhead when caching primitive values but also
does not allow any risks of cache corruption

Incr and Decr have had to be recreated so its no longer an atomic
operation, meaning there is a potential case where these will not work
as expected. the only fix I could see here is to check is numeric on way
in and out of the cache and manually serialise instead of the driver
handling it.

The membership group also continues to grow over time if keys expire
naturally, meaning that key consumes more and more of the available
cache as time goes on increasing the likelihood of an eviction of the
item

### DIFF
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -125,7 +125,7 @@ class CI_Cache_redis extends CI_Driver
 			{
 				log_message('error', 'Cache: Redis authentication failed.');
 			}
-			$obj_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);
+			this->_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);
 		}
 		catch (RedisException $e)
 		{
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -125,7 +125,7 @@ class CI_Cache_redis extends CI_Driver
 			{
 				log_message('error', 'Cache: Redis authentication failed.');
 			}
-			this->_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);
+			$this->_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);
 		}
 		catch (RedisException $e)
 		{
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -186,7 +186,7 @@ class CI_Cache_redis extends CI_Driver
 	 */
 	public function increment($id, $offset = 1)
 	{
-		$oldValue = (int) $this->-redis->get($id);
+		$oldValue = (int) $this->_redis->get($id);
 
 		$ttl = $this->_redis->ttl($id);
 		if ($ttl < 0) {
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -193,7 +193,7 @@ class CI_Cache_redis extends CI_Driver
 			$ttl = self::DEFAULT_TTL;
 		}
 
-		return $this->_redis->set($id, $oldValue += $offset, $ttl);
+		return $this->_redis->set($id, $oldValue + $offset, $ttl);
 
 	}
 
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -48,6 +48,8 @@ defined('BASEPATH') OR exit('No direct script access allowed');
  */
 class CI_Cache_redis extends CI_Driver
 {
+
+	const DEFAULT_TTL = 60;
 	/**
 	 * Default config
 	 *
@@ -68,13 +70,6 @@ class CI_Cache_redis extends CI_Driver
 	 * @var	Redis
 	 */
 	protected $_redis;
-
-	/**
-	 * An internal cache for storing keys of serialized values.
-	 *
-	 * @var	array
-	 */
-	protected $_serialized = array();
 
 	// ------------------------------------------------------------------------
 
@@ -130,15 +125,12 @@ class CI_Cache_redis extends CI_Driver
 			{
 				log_message('error', 'Cache: Redis authentication failed.');
 			}
+			$obj_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);
 		}
 		catch (RedisException $e)
 		{
 			log_message('error', 'Cache: Redis connection refused ('.$e->getMessage().')');
 		}
-
-		// Initialize the index of serialized values.
-		$serialized = $this->_redis->sMembers('_ci_redis_serialized');
-		empty($serialized) OR $this->_serialized = array_flip($serialized);
 	}
 
 	// ------------------------------------------------------------------------
@@ -151,14 +143,7 @@ class CI_Cache_redis extends CI_Driver
 	 */
 	public function get($key)
 	{
-		$value = $this->_redis->get($key);
-
-		if ($value !== FALSE && isset($this->_serialized[$key]))
-		{
-			return unserialize($value);
-		}
-
-		return $value;
+		return $this->_redis->get($key);
 	}
 
 	// ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -48,6 +48,8 @@
  */
 class CI_Cache_redis extends CI_Driver
 {
+
+	const DEFAULT_TTL = 60;
```
narfbg: Why do we need this?

system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -130,15 +125,12 @@ public function __construct()
 			{
 				log_message('error', 'Cache: Redis authentication failed.');
 			}
+			this->_redis->setOption(Redis::OPT_SERIALIZER, Redis::SERIALIZER_PHP);
```
narfbg: Syntax error ...

narfbg:
There's room for discussing whether what we currently have should be changed, but at the expense of breaking increment/decrement? Not keen on that.

Also, any such change should definitely target develop and not 3.1-stable.

system/libraries/Cache/drivers/Cache_redis.php
exussum12: Just to remove the magic number which currently exists

system/libraries/Cache/drivers/Cache_redis.php
exussum12: Apologies I copy pasted this badly

system/libraries/Cache/drivers/Cache_redis.php
narfbg: Replacing it with a magic constant ... :)

It currently only appears in one place and it's a variable; in your changeset you use it in another, but that seems rather arbitrary.

exussum12:
Happy to discuss this. The driver as it stands took 45 minutes to cause issues for us on live. Making a change to keep the group membership but check each time took about 2 weeks to cause an issue. 

Not sure of others experience with this driver and the amount of traffic they receive.

system/libraries/Cache/drivers/Cache_redis.php
exussum12: It's in two. Once there and once as the third Param of the set.

narfbg: Once "there" ... where? In `increment()`? That's what I referred to was arbitrary ... What has incrementing got to do with TTLs?

narfbg:
OK, well, for starters why the changes to increment/decrement? Especially since Redis supports them as atomic operations out of the box.

exussum12:
That needed to change unfortunately due to the value stored in redis being something like "i:3" now running increment on that will not work as expected.  The only other way I can think to check this is doing is numeric checks on the way in and out of the cache and then manually serialising everything else

system/libraries/Cache/drivers/Cache_redis.php
exussum12: If the value does not exist. Increment will make a new value with a TTL of the default TTL. So `incr(something)` if `something` doesn't exist `something` is set to 1 with a TTL of the default TTL. Other place the default is used is in the case of a missing Param of set. This makes both places use the same value

narfbg:
I see ... Well, that's probably why ``Redis::SERIALIZER_PHP`` isn't currently used - there's no reason why scalar values should be serialized, and I'm not really keen on changing that. Unless you have some quite convincing arguments?

exussum12:
Ok so your happy if I change this to do numeric checks on way in and out of the cache ? This will leave incr / decr as atomic but be slightly slower on each set / get ?

narfbg:
Numeric checks would solve the increment/decrement issue, but the root problem isn't just numerics, it's scalars. You'd have to do `is_scalar()` checks and at that point we're really looking at just replacing `_ci_redis_serialized` with another way of keeping track what needed serializing and what not ...

exussum12:
You can infer that from the what comes back, Just try and unserialize. if that is successful use the value as is, if it fails then use it as the type which comes out.

The issue is saving a string which is also a valid serialization format. Personally I would just treat numbers differently

exussum12:
The issue with _ci_redis_serialized is that it is completely separate and redis has no way of linking the two.

narfbg:
Blind `unserialize()` is an absolute no go; it can lead to RCE vulnerabilities. In fact, PHP itself has had to deal with multiple `deserialize()` vulnerabilities in recent years.

Maybe if we used JSON ... but that would break native PHP objects.

exussum12:
Yeah, which leaves it with just treating numbers differently ?

narfbg:
No? As I said, treating numbers differently only solves one issue.

I'd much rather switch to JSON instead.

exussum12:
Json will break much more, stdobjects vs arrays will be impossible to tell for example

https://3v4l.org/PUmXq

exussum12:
I can change this to do a getset which will ensure the updated value hasnt been touched in the mean time, It wont use the incr / decr commands though.

narfbg:
Oh, yes I forgot about that in JSON.

Still, I don't get why you continuously insist on treating only numbers differently. Forget the numbers and think of scalars as a whole. We either serialize all, serialize only non-scalars or serialize nothing.

Leaving aside the fact that serializing scalars is just stupid because it's unnecessary, serializing strings is the most dangerous of all, so I still prefer serializing only arrays & objects even with a performance hit. So, to me the issue becomes how much can we reduce that hit.

exussum12:
Treating numbers differently is because they are the only things treated differently. incr / decr is really the only missing functionality. Storing a string in a very slightly less efficient format for me is not a deal breaker. The current way keeps an entire separate list of items, so is not space effective either.

Some solutions that I can think of are

1. Keep this and replace the current incr / decr with getset  this would mean we can 100% accurately keep the values in sync . This way has the least BC impact.

2. Use a database to manage the current  _ci_redis_serialized https://redis.io/commands/select   This way would tie the value to the db type. Would need to check both (probably scalar first, then serialized). This would keep the behavior close to how it is now, but be much more robust.  
A read would be SELECT -> GET -> SELECT -> GET worst case, but would mean that scalars are stored as is. Sets would be SELECT -> SET -> DELETE  to delete the other versions value, so one key can only have 1 value (ie key abc could only appear in scalar or serialized, not both) This would break BC - people upgrading would need to flush their redis instances to continue to work.

3. Implement incr / decr with watch. Its slightly more complex code but will give an error if the update fails to happen.  This would still mean doing incr / decr in php. but would make the update completely safe. This will have little BC impact but will add load to redis (from what I have seen)


Would like to hear your thoughts on any of these ?

narfbg:
It's not about space efficiency, it's more about security and to some extent about keeping things simple and not reinventing the wheel. `Redis::SERIALIZE_PHP` is a PHP feature, not a Redis one; if Redis itself doesn't support storing non-scalars then it makes perfect sense for us to only make a workaround for that instead of reinvent what Redis already does natively for scalars.

Anyway, onto your points:

1. I think you already know what I think of this - I'd avoid serializing scalars.
2. I'm not really familiar with this database feature, so I'd have to read up on it ... why would we need 3 operations to perform a set? Not concerned about people having to flush their caches.
3. How's that different from 1 in terms of dealing scalars? If it's just a different way of special treatment for numerics, I don't like it.

exussum12:
For security serialising everything makes sense. For keeping things simple it's also the least amount of code. 

For the set imagine key ABC. We originally set ABC to be 123. This would be a scaler. So put in the scalar section. If we then use set ABC ['some other value'] this would set the object version also in cache. So both values would be there. To keep it consist with now after the set we should delete the value which would normally have been overwritten. 

On point 2 flushing in redis is actually much more expensive than flushing in memcache for large values of cache as it needs to remove all the data on disk too. It causes us about 1 minute if downtime. Upgrading the framework may be worth that 1 minute of downtime though ? 

I can put a proof of concept for 2 up of that helps

exussum12:
having a bit more of a play with redis, passing int "1" returns string "1". Personally I see this as a better reason to keep serialization. For example passing a phone number starting with 0. coming out if you were to treat that as an int it would be cast as octal.

The current code here would not have that issue.

narfbg:
Sorry about not replying earlier, I was travelling ...

Whatever (if any) change we make, it's going to be in 3.2.0 and not in a patch version - this is why I'm not bothered by potential downtime; it's natural to have some work to do in a more major upgrade.

On that example with the phone number - doesn't Redis recognize if you pass it an int or string? Passing `0123456` (as opposed to `'0123456'`) would make PHP itself treat that as an octal and cast it to integer, so this shouldn't be our problem.

But other than that, I kind of got lost in your replies and I still hold the opinion that it is best to just optimize what we currently have.

exussum12:
opened #5752 which is what I think you mean by the "optimize what we currently have" comment, we actually did that first, it made things better but still had issues with it

exussum12:
Also jusst noticed the incr and decr are currently broken also. the command it means to use is incrBy and decrBy 

The second param is currently ignored

narfbg:
Merged both of your other PRs, thanks a lot for the patience and cooperation. I'll leave this one open for you in case you come up with something else, but I suspect we've exhausted all options.

PR #5741, opened 2019-04-17
Added config option to disable view variable caching

AllanJeremy:
This is an enhancement to the [fix here](https://github.com/bcit-ci/CodeIgniter/pull/2684)

## The problem
Variables from one view are loaded into the next instance of the same view loaded. Sometimes this behavior may be undesirable. Added a config item & option to disable the caching of the view variables.

**If the config item hasn't been added:**
Defaults to TRUE (even if the variable is not declared in the config file - to allow for upgrades using only system directory copy)

> Config variable is currently set to `FALSE` for testing. Forgot to change it back to `TRUE` before sending the PR

The caching currently occurs in the `Loader` class & the `view()` method to be specific

## Current solution
Clear the cache on every load if we have disabled variable caching

## Optimal solution
This would mean preventing the actual variable caching from happening. Since I wasn't sure whether this would break things, I used the first approach.

### Will the optimal solution break the usability in future?
No. In future, the actual usability **won't be affected**, the only thing that will change is where the check for caching/clearing cache occurs.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -469,3 +469,19 @@ $config['time_reference'] = 'local';
 | Array:		array('10.0.1.200', '192.168.5.0/24')
 */
 $config['proxy_ips'] = '';
+
+/* 
+|--------------------------------------------------------------------------
+| Cache view vars
+|--------------------------------------------------------------------------
+|
+| Variable caching allows for view variables to be passed down from one view to another
+| This means that the variables set in one view are passed into the next call to the same view loading
+| This option allows you to disable variable caching in views, meaning each view load will NOT inherit variables belonging to the previous load
+| Enable/disable view variable caching. 
+|
+| Defaults to TRUE.
+|
+| Accepted values are TRUE or FALSE
+*/
+$config['cache_view_vars'] = FALSE;
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -493,6 +493,11 @@ class CI_Loader {
 	 */
 	public function view($view, $vars = array(), $return = FALSE)
 	{
+		if(config_item('cache_view_vars') === FALSE)
+		{
+			$this->clear_vars();
+		}
+
 		return $this->_ci_load(array('_ci_view' => $view, '_ci_vars' => $this->_ci_prepare_view_vars($vars), '_ci_return' => $return));
 	}
 


narfbg:
Preventing caching in the first place would be the desired behavior if this is to be merged. However ...

Adding an entire config option for something that's rarely used seems like overkill. Also the way you implemented it is very counter-intuitive - it says `clear_cache_vars`, but you have to set it to `TRUE` not to clear them *and* the default is `FALSE`, changing current behavior.

There's another subtle, but very important BC break - this prevents vars from being carried down into nested views unless they're explicitly passed there as well. This would happen A LOT more often than the case you're trying to solve.

So, you can see how I just have to say No for this as it is. Maybe if you make substantial changes that resolve all standing problems with it, but ... Well, that won't be easy.

AllanJeremy:
Thanks for the reply @narfbg.
>
> Preventing caching in the first place would be the desired behavior if this is to be merged. However ...
> Adding an entire config option for something that's rarely used seems like overkill.

Will give this a shot as soon as I can.  Also, with the config option being optional, wouldn't that balance things out? In any case, is there any other way to approach globally setting of the "config variable" without using an actual `config` item?

>There's another subtle, but very important BC break - this prevents vars from being carried down into nested views unless they're explicitly passed there as well. 

Indeed it does. Regarding this particular case, that would be the time to clear cached variables manually? The current fix assumed a scenario where views may be used as "components", in which case, we would want values to be automatically passed. Probably an edge-case at this point but picked up the habit of using views as "components" when working with Vue. 

As such, how would you propose this "break" be avoided?

narfbg:
> Also, with the config option being optional, wouldn't that balance things out? In any case, is there any other way to approach globally setting of the "config variable" without using an actual `config` item?

Well, that may look perfectly fine on paper, but that's quite deceptive. An option that completely changes how code works may look convenient at first if (as in this case) spares you from writing more code, but then it also turns any edge case scenario into a major need for refactoring. That's why config vars are generally reserved for "must haves" such as your character set and other environment-specific things.
If you're hell-bent on having this for your own usage, I suppose you could create a `MY_Loader` extension to change the default, but for everybody's use global feature switches are not a good thing.

AllanJeremy:
> 
> 
> > Also, with the config option being optional, wouldn't that balance things out? In any case, is there any other way to approach globally setting of the "config variable" without using an actual `config` item?
> 
> Well, that may look perfectly fine on paper, but that's quite deceptive. An option that completely changes how code works may look convenient at first if (as in this case) spares you from writing more code, but then it also turns any edge case scenario into a major need for refactoring. That's why config vars are generally reserved for "must haves" such as your character set and other environment-specific things.
> If you're hell-bent on having this for your own usage, I suppose you could create a `MY_Loader` extension to change the default, but for everybody's use global feature switches are not a good thing.

Makes perfect sense. Thanks. 

That's what I'm currently working with (MY_Loader).
I'll check out some of the issues sometime and see if there's something I can help out with

PR #5742, opened 2019-04-17
Fix an "$this->load->vars()"  security bug

l4yn3:
```
class Welcome extends CI_Controller {
    public function index()
    {
        $this->load->vars("_ci_path", "C:/windows/win.ini");
	$this->load->view('welcome_message');
    }
}
```
The demo code could include file `C:/windows/win.ini`.
 If the first param can be controlled in  `$this->load->vars("_ci_path", "C:/windows/win.ini");`
  like this:
 `$this->load->vars($_GET['field'], $_GET['condition']);` 
The hacker can read everything or run command seriously.

### DIFF
--- a/system/core/Loader.php
+++ b/system/core/Loader.php
@@ -529,7 +529,7 @@ class CI_Loader {
 		$vars = is_string($vars)
 			? array($vars => $val)
 			: $vars;
-        $vars = $this->_ci_prepare_view_vars($vars);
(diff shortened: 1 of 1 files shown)


narfbg:
Yea, this isn't a bug. Very reckless coding is necessary to allow such things, at which point countless other vulnerabilities appear and a lot of them probably aren't preventable.

PR #5744, opened 2019-04-22
Adding support for proper CRLF

sharadrsoni:
Window's Notepad doesn't recognize `\n` linebreak it has to be `\r\n`

### DIFF
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -253,7 +253,7 @@ class CI_Log {
 	 */
 	protected function _format_line($level, $date, $message)
 	{
-		return $level.' - '.$date.' --> '.$message. PHP_EOL;
+		return $level.' - '.$date.' --> '.$message.PHP_EOL;
 	}
 
 	// --------------------------------------------------------------------
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -253,7 +253,7 @@ class CI_Log {
 	 */
 	protected function _format_line($level, $date, $message)
 	{
-		return $level.' - '.$date.' --> '.$message."\r\n";
+		return $level.' - '.$date.' --> '.$message. PHP_EOL;
 	}
 
 	// --------------------------------------------------------------------
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -249,7 +249,7 @@ class CI_Log {
 	 * @param	string	$level 	The error level
 	 * @param	string	$date 	Formatted date string
 	 * @param	string	$message 	The log message
-	 * @return	string	Formatted log line with a new line character '\r\n' at the end
+	 * @return	string	Formatted log line with a new line character at the end
 	 */
 	protected function _format_line($level, $date, $message)
 	{
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -249,11 +249,11 @@ class CI_Log {
 	 * @param	string	$level 	The error level
 	 * @param	string	$date 	Formatted date string
 	 * @param	string	$message 	The log message
-	 * @return	string	Formatted log line with a new line character '\n' at the end
+	 * @return	string	Formatted log line with a new line character '\r\n' at the end
 	 */
 	protected function _format_line($level, $date, $message)
 	{
-		return $level.' - '.$date.' --> '.$message."\n";
+		return $level.' - '.$date.' --> '.$message."\r\n";
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jim-parry:
This is going to mess things up for all non-notepad usage, isn't it?
The better strategy could be to use notepad++ instead of notepad.

gxgpet:
Or to use the existing `PHP_EOL`.

sharadrsoni:
> This is going to mess things up for all non-notepad usage, isn't it?
> The better strategy could be to use notepad++ instead of notepad.

I have checked on notepad++ and other editors, it is working fine.

sharadrsoni:
> Or to use the existing `PHP_EOL`. But yeah, Notepad++ should do the trick, too.

PHP_EOL is defined as "\n"

gxgpet:
> PHP_EOL is defined as "\n"

Well, if you are running two different types of OS on your server and your own machine, then nothing can help, but a more advanced text editor/viewer.

sharadrsoni:
> > PHP_EOL is defined as "\n"
> 
> Well, if you are running two different types of OS on your server and your own machine, then nothing can help, but a more advanced text editor/viewer.

No, I am using windows only, using WAMP stack.
My thought is we have a default application for notepad by Windows OS, we need to support them, we can't just tell to use any other notepad editor.
This is a log file, ideally shared across multiple OS and can be viewed by layman too, for them they don't know what is notepadd++.

I hope you are understanding my point.

gxgpet:
Just made a test on my local env:

```
C:\xampp\php>php.exe -r "echo strlen(PHP_EOL);"
2
```

Are you sure about that `PHP_EOL` being `\n`?

sharadrsoni:
> PHP_EOL

Correct because there are 2 characters only. i.e. `\n`

narfbg:
> Correct because there are 2 characters only. i.e. `\n`

`\n` is a single character.

Also: https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/

I would just close this, but `PHP_EOL` seems like a decent suggestion.

sharadrsoni:
> > Correct because there are 2 characters only. i.e. `\n`
> 
> `\n` is a single character.
> 
> Also: https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/
> 
> I would just close this, but `PHP_EOL` seems like a decent suggestion.

My Bad, reason: On ctrl + click on `PHP_EOL` PhpStrom shows `Core_d.php` where it only shows `\n` but while evaluating an expression runtime, I got the value of '\r\n'

I have made necessary changes, but still, for my project, I am going to use `/r/n` as it is harmless for Windows and Linux [to my understanding].

My use-case is, clients, download the logs and open in notepad in Windows, for them it looks like a mess, hence have to use `/r/n` as it is not harming in Linux too.

Thanks for sharing the Link of notepad, but that feature is not yet enabled, and to so one need to change the registry which is not a cup of tea for every layman.

gxgpet:
That is because PhpStorm's internal documentation relies on some hardcoded stubs (which are available on the Github as well for editing), and not on the configured PHP interpreter.

narfbg:
> Thanks for sharing the Link of notepad, but that feature is not yet enabled, and to so one need to change the registry which is not a cup of tea for every layman.

The thing has worked with just `\n` for the entirety of CodeIgniter's existence. We're not just now going to change it because a few people don't have a third-party software's patch yet ... :)

system/core/Log.php
```diff
@@ -249,11 +249,11 @@ public function write_log($level, $msg)
 	 * @param	string	$level 	The error level
 	 * @param	string	$date 	Formatted date string
 	 * @param	string	$message 	The log message
-	 * @return	string	Formatted log line with a new line character '\n' at the end
+	 * @return	string	Formatted log line with a new line character '\r\n' at the end
 	 */
 	protected function _format_line($level, $date, $message)
 	{
-		return $level.' - '.$date.' --> '.$message."\n";
+		return $level.' - '.$date.' --> '.$message. PHP_EOL;
```
narfbg: No spaces around the concatenation operator please, same applies for the other file as well.

system/core/Log.php
```diff
@@ -249,11 +249,11 @@ public function write_log($level, $msg)
 	 * @param	string	$level 	The error level
 	 * @param	string	$date 	Formatted date string
 	 * @param	string	$message 	The log message
-	 * @return	string	Formatted log line with a new line character '\n' at the end
+	 * @return	string	Formatted log line with a new line character '\r\n' at the end
```
narfbg: ... and now this isn't necessarily true.

I guess it's better just to remove it, so it only says newline character.

PR #5751, opened 2019-05-07
reset query after completing a multi-table delete if reset_data parameter is true

bengennaria:
Delete queries were not properly reset if multiple tables were passed causing subsequent queries to retain the query attributes passed for the delete query

Signed-off-by: Ben Gennaria <ben@gennaria.com>

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -2236,11 +2236,16 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		}
 		elseif (is_array($table))
 		{
-			empty($where) && $reset_data = FALSE;
+			$single_table_reset_data = $reset_data;
+			empty($where) && $single_table_reset_data = FALSE;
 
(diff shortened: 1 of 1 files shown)


narfbg:
Can you provide an example scenario of when/how this happens? Because, with `delete()` being called in that loop, it should actually reset on each iteration ...

bengennaria:
> Can you provide an example scenario of when/how this happens? Because, with `delete()` being called in that loop, it should actually reset on each iteration ...

If an empty where clause is provided then the original code changes the $reset_data parameter to false which means $this->_reset_write() is never called after finishing the individual table deletes.

narfbg:
Ugh, trying to imagine the scenario you're trying to do and whatever I can come up with isn't pretty. Can you please provide an example?

The patch that you've provided needs work anyways; the question is whether it's going to be about code quality or actually restricting such deletes, because that seems dangerous.

PR #5752, opened 2019-05-08
Alterntive redis fix

exussum12:
We ran this for about 2 weeks without issue. It keeps the group and works completely with whats there currently.

This may be the "optimized" the version you were talking about.

Its major drawback is that is only works correctly in the default set up for redis. the other PR correctly fixes this for all options in redis.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -135,10 +135,6 @@ class CI_Cache_redis extends CI_Driver
 		{
 			log_message('error', 'Cache: Redis connection refused ('.$e->getMessage().')');
 		}
-
-		// Initialize the index of serialized values.
-		$serialized = $this->_redis->sMembers('_ci_redis_serialized');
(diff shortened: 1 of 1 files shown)


narfbg:
Looks OK. What do you mean by default set up and other options?

exussum12:
Redis out of the box will only consider items with a TTL > 0 for LRU evictions. This works as the group has no TTL. Other modes make redis consider all keys (more similar to memcache) 

In the other mode which is more similar to memcache the group can be removed resulting in code which expects objects or arrays to actually get serialised strings. This shows in logs as can not call method x on string. 

This is the reason the group needs to go as it can be expired separate from the data.

narfbg:
Well, we'll have to live with that for now. :)

narfbg:
Eh, I just noticed that you're again targeting 3.1-stable instead of develop ... Is this BC-free?

exussum12:
This is BC. It works identically to the current menthod but instead of front loading the group it checks on each use. It's o(1) (and fast) in redis. 

The only reason I stopped using this is the issue mentioned above. We dont run redis with the default options

PR #5753, opened 2019-05-08
Fix increment / decrement in redis

exussum12:
also fixed a value incorrectly returned docblock says array but bool returned

### DIFF
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -228,7 +228,7 @@ class CI_Cache_redis extends CI_Driver
 	 */
 	public function increment($id, $offset = 1)
 	{
-		return $this->_redis->incr($id, $offset);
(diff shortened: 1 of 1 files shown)


system/libraries/Cache/drivers/Cache_redis.php
```diff
@@ -294,7 +294,7 @@ public function get_metadata($key)
 			);
 		}
 
-		return FALSE;
+		return array();
```
narfbg: Please revert this, it's supposed to return FALSE.

exussum12: Update the doc block then ? It's currently saying it only returns array

narfbg: Nope. I know it says only array, we simply don't bother with the possibility for `FALSE` being returned.

exussum12: Ok. Seems odd. I've reverted this

PR #5759, opened 2019-05-15
Encode Url

sedhossein:
Encode the special sanitize chars


narfbg:
XSS protection isn't just adding as much encoding/sanitization/whatever you can come up with on top of the code. In fact, double escaping can backfire and sometimes result in vulnerabilities.

PR #5774, opened 2019-06-10
fixed coutable warn when not found row index.

ytetsuro:
I using PHP7.3.

I occurred error following when not found row index.

```
Severity:    Warning
Message:     count(): Parameter must be an array or an object that implements Countable
Filename:    /var/www/html/vendor/codeigniter/framework/system/database/DB_result.php
Line Number: 386

Backtrace:
	File: /var/www/html/application/models/Group_model.php
	Line: 60
	Function: row
```

The following code can be reproduced.

```php
<?php

class Group_model extends CI_Model
{
   public function get($id) {
       $query = $this->db->get_where('group', ['id' => 1], 1);
       $not_found_index = 1000;

       return $query->row($not_found_index, Group::class);
   }
}
```

### DIFF
--- a/system/database/DB_result.php
+++ b/system/database/DB_result.php
@@ -383,7 +383,7 @@ class CI_DB_result {
 	{
 		isset($this->custom_result_object[$type]) OR $this->custom_result_object($type);
 
-		if (count($this->custom_result_object[$type]) === 0)
+		if (isset($this->custom_result_object[$type]) === false)
 		{
 			return NULL;
 		}


narfbg:
This is really weird ... if you look at `custom_result_object()`, you'd see that there's supposed to be an empty array there in case of no results.
Which driver are you using?

ytetsuro:
@narfbg 

> Which driver are you using?

I using mysqli.

> This is really weird ... if you look at custom_result_object(), you'd see that there's supposed to be an empty array there in case of no results.

I think don't weird.

`custom_result_object()` makes sure that the result is 0 to return an empty array.
such, `custom_result_object` has no data.

https://github.com/bcit-ci/CodeIgniter/blob/8d0e535c406490ea0fead8ed3389da249aab925a/system/database/DB_result.php#L184

> I occurred error following when not found row index.

My explanation is confusing you.
I'm very sorry.
I think that occurs when trying to get data for the empty table.

narfbg:
> `custom_result_object()` makes sure that the result is 0 to return an empty array.

Exactly. `count(array())` returns 0; it doesn't trigger an error, that's why I said it's weird.

However, I think I did find where the issue is - that empty array is never actually assigned on line 384. Will commit the fix, linking it here. Please confirm that it works.

ytetsuro:
@narfbg 

I confirmed.
Thanks fixed.

PR #5778, opened 2019-06-14
Strict mode db option is not required

toonitw:
You might want to don't set sql_mode on every connection.


narfbg:
Sorry, but I don't think this belongs in there.

If you really know what you're doing, you'll find your way to it. And if not ... you very likely don't know what it does.

PR #5793, opened 2019-07-03
Redis driver: Continue waiting for the lock instead of failling

nicolasduranl:
Right now, one of two concurrent requests will fail if one of them sets the lock first in Redis driver, what I propose is to continue waiting for the lock to be free instead of failing.

The only downside is that the request will be delayed at least a second (obviously because of the sleep time) but, as I myself prefer that in my application, maybe everyone else having requests failing prefer this scenario, specially when calling multiple ajax request on a page load.


narfbg:
You should take a look at #5708 first.

PR #5798, opened 2019-07-11
Speed up xss_clean for numerics

karololszacki:
You can't have XSS in numerics ("123", "0.54321" etc).
`is_numeric()` is very fast and when `xss_clean()` is used very often (to check all POST data, etc),
this change can be a huge benefit to the application - it just skips over all costly regex executions

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -354,6 +354,12 @@ class CI_Security {
 	 */
 	public function xss_clean($str, $is_image = FALSE)
 	{
+		// Is the string a numeric? you can't have XSS in a numeric
+		if (is_numeric($str))
+		{
+			return $str;
+		}
+
 		// Is the string an array?
 		if (is_array($str))
 		{


narfbg:
You also can't have XSS in an empty string (or one that's just spaces), nulls, booleans, objects and resources (i.e. because you passed the wrong variable), but that's aside the reason why I suspect you're having any complaints about speed.

You're not supposed to pass every single input to `xss_clean()`; what you're supposed to do is use `xss_clean()` only when *printing* inputs that you know may contain XSS and you don't do that when you output data from fields that you know are always numeric, so ... what you're suggesting here as a performance improvement actually becomes a (neglible indeed) performance regression for the actual proper use cases.

karololszacki:
We're using `xss_clean()` when storing the inputs to prevent "stored XSS" / "persistent XSS" attacks.
Our API accepts data in a form of "arrays of arrays of data", and we're accessing it in a manner similar to `$data = $CI->input->post("data", true)` - because in the end we want to prevent any xss possible, and we wanted easy and simple access to post-processed data.

It was for our convenience to do it this way and modify `xss_clean` with a form of a whitelist (and indeed we added check for `empty()` too, _and one simple whitelist-regex for ISO date-time format, since that is what we expect in 80% of our api fields, but that was not suitable for upstream at all_).

Otherwise we'd have to loop over all the input fields and filter them with `xss_clean`, and that would be a second walk over all of them (since internally, input uses `_fetch_from_array`, which does a `foreach` on all of them too). 

From a philosophy standpoint, I understand that you should do any XSS cleaning when you actually send data in response, and that you should do it "when printing inputs". 
But what you are saying is inconsistent with the framework, because it has a param to do xss cleaning on the `input` class, not on the `view` or `output` classes. CodeIgniter guides you to use `xss_clean` this way (on the inputs), both in code and in documentation. 
There's even an old, deprecated setting `global_xss_filtering` which runs this function on all inputs and I suppose there are still people using that. 

There is a very big performance penalty for "wrong usage" of `xss_clean`, and it is very easy to go this "wrong way", as my team now has gone the wrong way, and even the whole CodeIgniter project have gone it in the past with `global_xss_filtering`.

If we apply this patch (and another for eg. `empty($str)` case) the performance penalty for "right usage" will be neglible, as you said, and also the number of right usages (from the right philosophy standpoint) will be lower compared to wrong usages (right: you do it on some data fields, only ones you know you have to; wrong: you do it on all).

However, if one uses `xss_clean` the wrong way, either via `global_xss_filtering` or by just doing it via `$data = $CI->input->post("data", true)` on large inputs, they might get very nice benefits from this patch.

I don't have the actual numbers for the project, but it's a change that gives big gains to some, while introducing neglible losses to other. I understand the opposition to introduce this change, but I'd advocate for more clarification on the subject of what you're "supposed to do" or "not do to" with this function in the documentation, and at least some information on how costly it might be (this runs a number of regexes on the input data, and you can pass arrays of arrays to it, which it will all parse). It amounted to more than 50% of cpu time for some special requests with large digits-only inputs (profiler info). This patch looked to us like a low hanging fruit and that's why I decided to create this Pull Request.

narfbg:
> We're using `xss_clean()` when storing the inputs to prevent "stored XSS" / "persistent XSS" attacks.

Well, that is one almost surefire way to be vulnerable to stored XSS, since any new attack vector that gets handled by `xss_clean()` in the future will not have been covered for your past stored data. The mangling it does is also often irreversible.

> From a philosophy standpoint, I understand that you should do any XSS cleaning when you actually send data in response, and that you should do it "when printing inputs".

This has nothing to do with philosophy, as I just explained.

> But what you are saying is inconsistent with the framework, because it has a param to do xss cleaning on the `input` class, not on the `view` or `output` classes.

That's convenience, not inconsistency. For one, even though you're supposed to do it in the output phase, you're still filtering inputs, so calling this an inconsistency is just plain false; but also there's no more convenient way to provide this functionality by other means.

> CodeIgniter guides you to use `xss_clean` this way (on the inputs), both in code and in documentation.

I'm not aware of an instance where the documentation guides you to do XSS sanitization the way you're doing it, but if you can point me to one that exists, I'll be happy to correct it.

> There's even an old, deprecated setting `global_xss_filtering` which runs this function on all inputs and I suppose there are still people using that.

Yes, and it's deprecated for a good reason ...

---

Sorry, we're not going to make it easier to do things the wrong way.

Issue #5801, opened 2019-07-15
Redis session driver now returns TRUE instead of '+PONG'


darvanen:
The recommended phpredis package has had this change recently:
https://github.com/phpredis/phpredis/issues/1563

This means that CI fails to close redis sessions because it is expecting a string of '+PONG' rather than the new boolean TRUE.

Line 252 of system/libraries/Session/drivers/Session_redis_driver.php

darvanen:
PR at https://github.com/bradtraversy/ciblog/pull/15

narfbg:
Please don't create issues just to reference them in pull requests, you can just explain the problem in the PR itself ...

darvanen:
Noted thank you.

Got to the issue before I realised how easy a PR could be, won't happen again.

mchobbylong:
Hi there! Since my PR is merged, I believe this issue has been solved. Please check that out for the latest solution 😉

narfbg:
Fixed by #5816.

PR #5802, opened 2019-07-15
Redis php driver now returns TRUE

darvanen:
Issue:
https://github.com/bcit-ci/CodeIgniter/issues/5801


narfbg:
This would break compatibility with older phpredis versions that still do return '+PONG'.

Also, please target the develop branch and not 3.1-stable. We handle the merges ourselves, don't worry about that.

darvanen:
Noted, thank you.

There was something in the phpredis issue about returning a given string, perhaps passed as a parameter, I'll look into it tomorrow.

PR #5808, opened 2019-07-30
Correct / Added characters

harrysrevis:
Corrected and added some characters in "foreign_chars.php" config file.


narfbg:
Looks fine, but I have to ask first - have you checked for duplicates in the file?

harrysrevis:
@narfbg Yes. I've double checked it just to be sure!

Issue #5125, opened 2017-05-19
xss_clean security library


colethorsen:
There is an issue in the xss_clean function where it might be a little over aggressive in stripping content.

when you run the string "100% becomes" through the xss_clean function the section on lines 377-387 strip out "% be" and the string becomes " 100�comes" this happens in any instance where the 2 characters following a percent are a-f 0-9.

I understand what this is preventing, and I'm assuming that it considers the space for good reason. So I'm not overly sure what the solution here would be.

narfbg:
Inconvenient, indeed. But I don't see a way around #4877 without this ...

colethorsen:
Well if that is the case wouldn't limiting the replacement to within a tag, or within a parameter inside a tag solve the issue?

narfbg:
That's easier said than done, and while the example in #4877 uses an attribute value, it is not really clear if that's the only vector ...

wrmond:
#4877  hot-fix is terrible idea. should check ahead if in attribute value first

Zae:
Hi guys and gals,

I know this issue was closed a while back already, but is there anyway you could maybe look into this again?

I'm hitting this bug because I'm trying to sanitize a rich text field, but now people can't type % in their text, because it creates weird utf-8 characters.

see my ticket at [graham campbells security package](https://github.com/GrahamCampbell/Laravel-Security/issues/36)

narfbg:
@Zae Refer to my previous comment ... Until it is proven that this is only limited to attribute values, it's not changing.

Agimovel:
This sucks...I'm making work arounds on this but don't know if will affect security.

No solutions yet?

mckaygerhard:
well seems guys from opensourcepos received same response .... "works for me" -> wont-fix!

skorp:
@francoisjacquet  think the first preg_replace should be changed to this not?
 ```
$str = preg_replace('#%((?:[a-f]){2,})#i', '%25$1', $str);
 ```
because else it captures only the latest iteration


i think i will use this modification to allow strings like this: '... 60% about ....'

```
....
if (stripos($str, '%') !== false)
		{
			do
			{
				$oldstr = $str;
				$str = rawurldecode($str);
				$str = preg_replace_callback('#%(?:\s*[0-9]\s*[0-9a-f])#i', array($this, '_urldecodespaces'), $str);

			}
			while ($oldstr !== $str);
			unset($oldstr);
		}
		....
```

[...]

PR #5813, opened 2019-08-08
Check decoding result before returning

xinhus:
The issue #5125 (which is closed) is a problem which could be solved by this pull request.
I've faced this issue when the user post some data with `% ca`.
The method `$this->input->post( 'mycustomfield', $xss_clean = true);` will convert `% ca` to a invalid character.

Signed-off-by: Rafael Cardoso <xinhus@gmail.com>

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -782,9 +782,12 @@ class CI_Security {
 	{
 		$input    = $matches[0];
 		$nospaces = preg_replace('#\s+#', '', $input);
-		return ($nospaces === $input)
+		$is_the_same_character = $nospaces === $input;
+		$decoded_input = rawurldecode($nospaces);
+		$is_invalid_character = !(mb_check_encoding($decoded_input));
+		return ($is_the_same_character || $is_invalid_character)
 			? $input
-			: rawurldecode($nospaces);
+			: $decoded_input;
 	}
 
 	// ----------------------------------------------------------------
--- a/tests/codeigniter/core/Security_test.php
+++ b/tests/codeigniter/core/Security_test.php
@@ -74,6 +74,17 @@ class Security_test extends CI_TestCase {
 
 	// --------------------------------------------------------------------
 
+	public function test_xss_clean_with_invalid_unicode()
+	{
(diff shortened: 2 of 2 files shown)


narfbg:
I fail to see how this is safe. Please explain in detail.

xinhus:
> I fail to see how this is safe. Please explain in detail.
Thank you for um reply @narfbg 

The method convert any string with % sign followed by two hex value with literal characters
BUT, sometimes these values are invalids.
Like in my test case `% ca`
This will add an invalid character into the string. Which is a bad behavior.
So before return the decoded value, it's check if it's a valid character.

Current behavior is:
The string `They have 100 animals, 10% cat and also 90% dogs`
Will be converter to `They have 100 animals, 10�t and also 90% dogs`

narfbg:
Of course I can see the result of your test case and of course I can see how the current behavior is undesirable to you; that's not the explanation I was looking for ...

I need to know that you're a subject expert on XSS and that this is 100% safe.

xinhus:
Hello @narfbg thank you again for your reply! 

It's safe because before return I'm asking is this is a valid character.

I'm not an  XSS EXPERT.
But you only can attack with a Unicode of a valid character like `%22` which means `"` or some like this.
I'm justing asking if this is a valid character before convert.
The current behavior will send an invalid character ahead which will cause some issues in database and things which contains a restricted charset.

But again, it's a bad behavior
If you don't fell secure about it fell free to close.
I'm just trying to help the open source ^^.

xinhus:
@narfbg I made some research about xss and invalid unicode and don't found anything.
Do you have any `tip` to find something??
Since we don't modify none of existing tests, it's safe?

narfbg:
No.

Sorry, but I've had enough headaches with XSS issues in the past and I'm taking 0 risk with it.

mckaygerhard:
> Current behavior is:
> The string `They have 100 animals, 10% cat and also 90% dogs`
> Will be converter to `They have 100 animals, 10�t and also 90% dogs`

sybase DBMS has some trick in this way.. seems fair segure to me!.. now CI does nothign to able to made life easier to the developer!

PR #5816, opened 2019-08-12
Adapt to new version of php-redis

mchobbylong:
In the new version of php-redis since `5.0.0`, some of the methods have been deprecated, like `Redis::setTimeout` and `Redis::delete`. Also, `Redis::ping` now returns `TRUE` (#5801).

### DIFF
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -37,6 +37,8 @@
  */
 defined('BASEPATH') OR exit('No direct script access allowed');
 
+define('PHPREDIS_VERSION', phpversion('redis'));
+
 /**
  * CodeIgniter Session Redis Driver
  *
@@ -453,7 +455,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 	 */
 	protected function _expire_key($key, $ttl)
 	{
-		if (version_compare(phpversion('redis'), '5', '>='))
+		if (version_compare(PHPREDIS_VERSION, '5', '>='))
 			return $this->_redis->expire($key, $ttl);
 		return $this->_redis->setTimeout($key, $ttl);
 	}
@@ -470,7 +472,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 	 */
 	protected function _delete_key($key)
 	{
-		if (version_compare(phpversion('redis'), '5', '>='))
+		if (version_compare(PHPREDIS_VERSION, '5', '>='))
 			return $this->_redis->del($key);
 		return $this->_redis->delete($key);
 	}
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -37,8 +37,6 @@
  */
 defined('BASEPATH') OR exit('No direct script access allowed');
 
-define('PHPREDIS_VERSION', phpversion('redis'));
-
 /**
  * CodeIgniter Session Redis Driver
  *
@@ -78,6 +76,33 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 	 */
 	protected $_key_exists = FALSE;
 
+	/**
+	 * Name of setTimeout() method in phpRedis
+	 * 
+	 * Due to some deprecated methods in phpRedis, we need to call the
+	 * specific methods depending on the version of phpRedis.
+	 * 
+	 * @var string
+	 */
+	protected $_setTimeout_name;
+
+	/**
+	 * Name of delete() method in phpRedis
+	 * 
+	 * Due to some deprecated methods in phpRedis, we need to call the
+	 * specific methods depending on the version of phpRedis.
+	 * 
+	 * @var string
+	 */
+	protected $_delete_name;
+
+	/**
+	 * Response of ping() method in phpRedis
+	 * 
+	 * @var mixed
+	 */
+	protected $_ping_response;
+
 	// ------------------------------------------------------------------------
 
 	/**
@@ -89,6 +114,20 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 	public function __construct(&$params)
 	{
 		parent::__construct($params);
+
+		// Detect the names of some methods in phpRedis instance
+		if (version_compare(phpversion('redis'), '5', '>='))
+		{
+			$this->_setTimeout_name = 'expire';
+			$this->_delete_name = 'del';
+			$this->_ping_response = TRUE;
+		}
+		else
+		{
+			$this->_setTimeout_name = 'setTimeout';
+			$this->_delete_name = 'delete';
+			$this->_ping_response = '+PONG';
+		}
 
 		if (empty($this->_config['save_path']))
 		{
@@ -170,8 +209,8 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 			}
 			else
 			{
-				$this->php5_validate_id();
 				$this->_redis = $redis;
+				$this->php5_validate_id();
 				return $this->_success;
 			}
 		}
@@ -242,7 +281,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 			$this->_session_id = $session_id;
 		}
 
-		$this->_expire_key($this->_lock_key, 300);
+		$this->_setTimeout($this->_lock_key, 300);
 		if ($this->_fingerprint !== ($fingerprint = md5($session_data)) OR $this->_key_exists === FALSE)
 		{
 			if ($this->_redis->set($this->_key_prefix.$session_id, $session_data, $this->_config['expiration']))
@@ -255,7 +294,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 			return $this->_fail();
 		}
 
-		return ($this->_expire_key($this->_key_prefix.$session_id, $this->_config['expiration']))
+		return ($this->_setTimeout($this->_key_prefix.$session_id, $this->_config['expiration']))
 			? $this->_success
 			: $this->_fail();
 	}
@@ -275,7 +314,7 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 		{
 			try {
 				$ping = $this->_redis->ping();
-				if ($ping === '+PONG' OR $ping === TRUE)
+				if ($ping === $this->_ping_response)
 				{
 					$this->_release_lock();
 					if ($this->_redis->close() === FALSE)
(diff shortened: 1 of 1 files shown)


narfbg:
Is there no way to check the phpredis version so we can avoid the dynamic checks?

Also, patch needs to comply with our [styleguide](https://www.codeigniter.com/user_guide/general/styleguide.html).

mchobbylong:
> Is there no way to check the phpredis version so we can avoid the dynamic checks?

phpredis deprecated the two functions in version `5.0.0` (phpredis/phpredis@a81b4f2d). It seems that `phpversion('redis')` is the right one to use.

I planned to add a static variable into the redis driver class, which stands for the version of phpredis. Unfortunately, no matter how I try, `phpversion('redis')` returns `NULL` in the constructor of the class, but returns the correct version string in other methods. So I add `phpversion('redis')` in those two protected functions.

> Also, patch needs to comply with our styleguide.

Corrected.

mchobbylong:
I just discovered that `define()` works. Use constants should be a better choice.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -37,6 +37,8 @@
  */
 defined('BASEPATH') OR exit('No direct script access allowed');
 
+define('PHPREDIS_VERSION', phpversion('redis'));
```
narfbg: I don't like this ...

What if this file is included more than once? (it shouldn't, but still ...)
What if some other library tries to define this constant? (we do also have a caching library that has a redis driver)
What if phpredis itself defines it at some point in the future? (I'm surprised it still hasn't)

This should be a property.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -272,7 +274,8 @@ public function close()
 		if (isset($this->_redis))
 		{
 			try {
-				if ($this->_redis->ping() === '+PONG')
+				$ping = $this->_redis->ping();
+				if ($ping === '+PONG' OR $ping === TRUE)
```
narfbg: Why would we silently accept both if we know the version?

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -307,9 +310,9 @@ public function destroy($session_id)
 	{
 		if (isset($this->_redis, $this->_lock_key))
 		{
-			if (($result = $this->_redis->delete($this->_key_prefix.$session_id)) !== 1)
+			if (($result = $this->_delete_key($this->_key_prefix.$session_id)) !== 1)
 			{
-				log_message('debug', 'Session: Redis::delete() expected to return 1, got '.var_export($result, TRUE).' instead.');
+				log_message('debug', 'Session: Redis::del() expected to return 1, got '.var_export($result, TRUE).' instead.');
```
narfbg: We don't know if `del()` or `delete()` was called ...

Also, why is `delete()` replaced by `del()` anyways?
That makes no sense if they work in the same way, and if they do work differently I'm very suspicious of just renaming the call without any other changes.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -439,4 +442,39 @@ protected function _release_lock()
...
+	/**
+	 * Expire key
+	 *
+	 * Sets expiration for a key
+	 *
+	 * @param	string	$key	Redis key
+	 * @param	int	$ttl	The key's remaining Time To Live, in seconds.
+	 * @return	bool
+	 */
+	protected function _expire_key($key, $ttl)
```
narfbg: I don't like this method name; I don't like that phpredis changed `setTimeout` to `expire` either ...

"expire" is a verb, implying immediate action and that's not what the function does.
Also the "key" part is redundant.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -439,4 +442,39 @@ protected function _release_lock()
...
+	{
+		if (version_compare(PHPREDIS_VERSION, '5', '>='))
```
narfbg: I asked if there's a way to check the version so we can *avoid the dynamic checks*. Now we're checking the version, but still doing it on every step ... that's no better than the previous solution.

See how we handle a similar situation here: https://github.com/bcit-ci/CodeIgniter/blob/3.1.10/system/libraries/Session/Session_driver.php#L77 (also a possible solution for `'+PONG'` vs. `TRUE`)

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -439,4 +442,39 @@ protected function _release_lock()
...
+			return $this->_redis->expire($key, $ttl);
```
narfbg: Still not complying with the styleguide, we don't have `if`s without braces.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -439,4 +442,39 @@ protected function _release_lock()
...
+
+	/**
+	 * Delete key
+	 *
+	 * Deletes a key
+	 *
+	 * @param	string	$key		Redis key
+	 * @return	bool
+	 */
+	protected function _delete_key($key)
```
narfbg: Same as with the previous method name, "key" is redundant.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Yes, you're right. I've tried to implement this as a property, but it was always `NULL` when I tested in my project.

Now it turns out that my custom redis driver in the project covers this driver, so I cannot see how I am doing after modifying the driver... I removed my custom redis driver, and now my code works 😉

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Solved in the latest commit.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: > We don't know if `del()` or `delete()` was called ...

Solved in the latest commit.

> Also, why is delete() replaced by del() anyways?

Here is what the author said in [https://github.com/phpredis/phpredis/commit/a81b4f2d](https://github.com/phpredis/phpredis/commit/a81b4f2d): 

> For unknown reasons some methods named differently to Redis
> commands (for example setTimeout instead of expire).

Looks like they just want to follow the name of commands in Redis.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Solved in the latest commit.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Learned it. That IS good implementation.

Solved in the latest commit.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Solved in the latest commit.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Solved in the latest commit.

mchobbylong:
By the way:

https://github.com/bcit-ci/CodeIgniter/blob/a66c71e131c2a7f05d57b03c8619d9bdd542b5bf/system/libraries/Session/drivers/Session_redis_driver.php#L169-L174

I am really confused about why `php5_validate_id()` should be called before setting `$this->_redis`.

Although I am using php7.2, this function will still be called, then calls `$this->_redis->exists()` in method `validateSessionId()`:

https://github.com/bcit-ci/CodeIgniter/blob/a66c71e131c2a7f05d57b03c8619d9bdd542b5bf/system/libraries/Session/drivers/Session_redis_driver.php#L349-L352

That will cause an exception, because `$this->_redis` is NULL.

To solve it, I altered L171 in the latest commit.

gxgpet:
Looks like that bug was introduced by me in #5781... :)

mchobbylong:
@gxgpet Aha! Turns out that this IS a bug and has been fixed. No more worries now 😉

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -76,6 +76,33 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
...
+	 * @var string
+	 */
+	protected $_delete_name;
+
+	/**
+	 * Response of ping() method in phpRedis
+	 * 
+	 * @var mixed
+	 */
+	protected $_ping_response;
```
narfbg: Should rather be `$_ping_success` or something like that.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -272,7 +313,8 @@ public function close()
 		if (isset($this->_redis))
 		{
 			try {
-				if ($this->_redis->ping() === '+PONG')
+				$ping = $this->_redis->ping();
```
narfbg: This is no longer necessary.

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -439,4 +481,35 @@ protected function _release_lock()
...
+	/**
+	 * Set timeout
+	 *
+	 * Sets expiration for a key
+	 *
+	 * @param	string	$key	Redis key
+	 * @param	int	$ttl	The key's remaining Time To Live, in seconds.
+	 * @return	bool
+	 */
+	protected function _setTimeout($key, $ttl)
```
narfbg: This and `_delete()` aren't necessary if you're just proxying to a variable method ... just use the variable method where necessary.

narfbg:
Good catch on the #5781 regression, but please submit a separate pull request for that ... it's a different issue.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Yes, that makes sense. Solved.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Solved.

system/libraries/Session/drivers/Session_redis_driver.php
mchobbylong: Solved. The two functions are now removed.

mchobbylong:
> Good catch on the #5781 regression, but please submit a separate pull request for that ... it's a different issue.

New PR is here #5828.

narfbg:
Thank you for your contribution and patience. And I mean that, few people have been as collaborative as you.

As I noted earlier, we also have a Cache library that utilizes phpRedis and it is likely to need similar changes. Would you like to tackle that too or should I go ahead myself? Either way is fine; I just wanted to give you the chance if you enjoy it. :)

system/libraries/Session/drivers/Session_redis_driver.php
```diff
@@ -272,7 +313,7 @@ public function close()
 		if (isset($this->_redis))
 		{
 			try {
-				if ($this->_redis->ping() === '+PONG')
+				if ($this->_redis->ping() === $this->_ping_success)
```
nebkat: Glad to see this change has made it in, gave me a lot of trouble until I realized today what was happening before seeing the pull request.

Should there be a log message if the ping does not return as expected? It becomes a silent failure that's very hard to track down, and causes most sessions to stay locked.

mchobbylong: That is right. It should be improved.

mchobbylong:
@narfbg Thanks for your appreciation. However, I don't know which issue you are mentioning because there are two issues currently active.

I think I could subscribe to that issue for discussion and further contribution. But you could go ahead but not just wait for me 😉

narfbg:
Sorry, there was no open issue ... see the last commit referenced above.

PR #5826, opened 2019-08-27
Cleanly handle array values for CSRF tokens

### LINKED ISSUE #5825: Error from array CSRF token
If the user POSTs an array value for the CSRF token, then CodeIgniter generates a PHP warning while trying to validate the CSRF value:

```
Severity: Warning
Message: hash_equals(): Expected known_string to be a string, array given
Filename: core/Security.php
Line Number: 231
```

Since the POSTed CSRF token is under the control of the user, CodeIgniter should cleanly handle this, instead of generating a warning.

joshkel:
If the user POSTs an array value for the CSRF token, then CodeIgniter
generates a PHP warning while trying to validate the CSRF value:

```
Severity: Warning
Message: hash_equals(): Expected known_string to be a string, array given
Filename: core/Security.php
Line Number: 231
```

Since the POSTed CSRF token is under the control of the user,
CodeIgniter should cleanly handle this, instead of generating a warning.

Fixes #5825

### DIFF
--- a/system/core/Security.php
+++ b/system/core/Security.php
@@ -229,6 +229,7 @@ class CI_Security {
 
 		// Check CSRF token validity, but don't error on mismatch just yet - we'll want to regenerate
 		$valid = isset($_POST[$this->_csrf_token_name], $_COOKIE[$this->_csrf_cookie_name])
+			&& is_string($_POST[$this->_csrf_token_name])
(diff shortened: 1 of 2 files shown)


narfbg:
As noted in #5825, this is already fixed in ba67b297d1e2baa281eed2cc7bf1fb16742cde87 ... it's just not merged in develop yet. As a side note, please don't open an issue just to link a pull request to it; PRs can serve as issues by themselves.

joshkel:
Thank you. I'll keep that in mind for the future.

PR #5832, opened 2019-09-08
Improve create_captcha()'s logging

gxgpet:
While moving to a different server, I noticed that `create_captcha()` was returning `false`, and since there was no logging made, I had to change manually the captcha helper to do some step-by-step debugging. It turned out that php-gd wasn't installed.

Since I'm already touching it, I added some logging for the scenario where img_path wouldn't be writable, as well.

### DIFF
--- a/system/helpers/captcha_helper.php
+++ b/system/helpers/captcha_helper.php
@@ -109,7 +109,9 @@ if ( ! function_exists('create_captcha'))
 			{
 				log_message('error', 'create_captcha(): "img_path" and "img_url" are required.');
 				return FALSE;
-			} elseif ( ! is_dir($img_path) OR ! is_really_writable($img_path))
+			}
+
+			if ( ! is_dir($img_path) OR ! is_really_writable($img_path))
 			{
 				log_message('error', "create_captcha(): '$img_path' is not a dir, nor it's writable.");
 				return FALSE;
--- a/system/helpers/captcha_helper.php
+++ b/system/helpers/captcha_helper.php
@@ -107,7 +107,7 @@ if ( ! function_exists('create_captcha'))
 		{
 			if ($img_path === '' OR $img_url === '')
 			{
-				log_message('error', 'create_captcha(): "$img_path" and "$img_url" are required.');
+				log_message('error', 'create_captcha(): $img_path and $img_url are required.');
 				return FALSE;
 			}
 
--- a/system/helpers/captcha_helper.php
+++ b/system/helpers/captcha_helper.php
@@ -113,7 +113,7 @@ if ( ! function_exists('create_captcha'))
 
 			if ( ! is_dir($img_path) OR ! is_really_writable($img_path))
 			{
-				log_message('error', "create_captcha(): '$img_path' is not a dir, nor it's writable.");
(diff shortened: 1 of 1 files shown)


system/helpers/captcha_helper.php
```diff
@@ -99,13 +99,19 @@ function create_captcha($data = '', $img_path = '', $img_url = '', $font_path =
...
 		}
 
 		if ($img_url !== '' OR $img_path !== '')
 		{
-			if ($img_path === '' OR $img_url === '' OR ! is_dir($img_path) OR ! is_really_writable($img_path))
+			if ($img_path === '' OR $img_url === '')
 			{
+				log_message('error', 'create_captcha(): "img_path" and "img_url" are required.');
+				return FALSE;
+			} elseif ( ! is_dir($img_path) OR ! is_really_writable($img_path))
```
narfbg: Per our styleguide, `elseif` must be on the next line (i.e. never on the same line as the closing brace before it).

An easier way to avoid this is to always make it an `if` instead, unless strictly necessary; it won't make a difference in this case.

system/helpers/captcha_helper.php
```diff
@@ -99,13 +99,19 @@ function create_captcha($data = '', $img_path = '', $img_url = '', $font_path =
...
+			log_message('error', 'create_captcha(): GD extension is not loaded.');
 			return FALSE;
 		}
 
 		if ($img_url !== '' OR $img_path !== '')
 		{
-			if ($img_path === '' OR $img_url === '' OR ! is_dir($img_path) OR ! is_really_writable($img_path))
+			if ($img_path === '' OR $img_url === '')
 			{
+				log_message('error', 'create_captcha(): "img_path" and "img_url" are required.');
```
narfbg: Might be better for it to say `$img_path` and `$img_url` instead.

system/helpers/captcha_helper.php
```diff
@@ -99,13 +99,19 @@ function create_captcha($data = '', $img_path = '', $img_url = '', $font_path =
...
+				return FALSE;
+			} elseif ( ! is_dir($img_path) OR ! is_really_writable($img_path))
+			{
+				log_message('error', "create_captcha(): '$img_path' is not a dir, nor it's writable.");
```
narfbg: Please put braces around the variable to make it more obvious that it's not a literal string.
I'm also fairly certain that the correct way would be to write "nor is it writable" instead, but English is not my native tongue so maybe someone can correct us?

system/helpers/captcha_helper.php
gxgpet: Fixed in 25931b0.

system/helpers/captcha_helper.php
gxgpet: Yes, you are right. See 24ff922.

system/helpers/captcha_helper.php
gxgpet: Fixed in 25931b0.

system/helpers/captcha_helper.php
narfbg: Sorry about the delayed response ...

It's certainly better now, but I meant for those quotes to go away. :)

gxgpet: I see. Alright, it's done now.

PR #5833, opened 2019-09-09
Update delete_files() in file helper.

curiousteam:
Hi,
Existing delete_files() function of Codeigniter file helper deletes sub-directories and files which are inside the supplied root path. But it does not delete the directory. Sometimes, you may want to delete a directory including any sub-directories and files. 

Imagine: we create a folder (i.e. 12_companyName) in the upload directory when a company registered by the user so that all upload files goes there. User has the ability to delete the registered company. Now, we want our script to delete concern folder of the company including its sub-directories and files, then delete from the database. Can we do that using delete_files() function?

Of course No.

Here is the solution:

delete_files($path[, $del_dir = FALSE[, $del_subdir = FALSE[, $htdocs = FALSE]]])

:param	string	$path: Directory path
:param	bool	$del_dir: Whether to delete the self directory including any sub-directories and files
:param	bool	$del_subdir: Whether to delete any sub-directories found in the path
:param	bool	$htdocs: Whether to skip deleting .htaccess and index page files
:returns:	TRUE on success, FALSE in case of an error
:rtype:	bool

If the second parameter is set to TRUE, the directory will be deleted including any directories or files contained within the supplied root path.

	Example::

		delete_files('./path/to/directory/', TRUE);

If the second parameter is set to FALSE and the third parameter is set to TRUE, any directories or files contained within the supplied root path will be deleted only.

	Example::

		delete_files('./path/to/directory/', FALSE, TRUE);

### DIFF
--- a/system/helpers/file_helper.php
+++ b/system/helpers/file_helper.php
@@ -99,12 +99,13 @@ if ( ! function_exists('delete_files'))
 	 * within the supplied base directory will be nuked as well.
 	 *
 	 * @param	string	$path		File path
-	 * @param	bool	$del_dir	Whether to delete any directories found in the path
+	 * @param	bool	$del_dir	Whether to delete the self directory including any sub-directories and files
+	 * @param	bool	$del_subdir	Whether to delete any sub-directories found in the path
 	 * @param	bool	$htdocs		Whether to skip deleting .htaccess and index page files
-	 * @param	int	$_level		Current directory depth level (default: 0; internal use only)
+	 * @param	int		$_level		Current directory depth level (default: 0; internal use only)
 	 * @return	bool
 	 */
-	function delete_files($path, $del_dir = FALSE, $htdocs = FALSE, $_level = 0)
+	function delete_files($path, $del_dir = FALSE, $del_subdir = FALSE, $htdocs = FALSE, $_level = 0)
 	{
 		// Trim the trailing slash
 		$path = rtrim($path, '/\\');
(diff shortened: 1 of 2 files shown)


system/helpers/file_helper.php
```diff
@@ -99,12 +99,13 @@ function write_file($path, $data, $mode = 'wb')
 	 * within the supplied base directory will be nuked as well.
 	 *
 	 * @param	string	$path		File path
-	 * @param	bool	$del_dir	Whether to delete any directories found in the path
+	 * @param	bool	$del_dir	Whether to delete the self directory including any sub-directories and files
+	 * @param	bool	$del_subdir	Whether to delete any sub-directories found in the path
 	 * @param	bool	$htdocs		Whether to skip deleting .htaccess and index page files
-	 * @param	int	$_level		Current directory depth level (default: 0; internal use only)
+	 * @param	int		$_level		Current directory depth level (default: 0; internal use only)
```
curiousteam: Tab was not showing correctly in my Sublime Text 3.

jim-parry: The style checker/fixer will fix tab alignment eventually, if you have the vommit pre-hook installed. I wouldn't worry about this.

curiousteam: Unfortunately, I don't have installed it. I have fixed the indentation in another commit.
Thanks.

narfbg:
Just use `rmdir()` after `delete_files()`.

PR #5834, opened 2019-09-13
Fix count_all_result if QB query contains Having

najdanovicivan:
There was an issues when using count_all_results with query builder. The select portion of query gets overwritten making the fields referenced in having not existent 

Signed-off-by: Najdanovic Ivan <najdanovicivan@gmail.com>


narfbg:
Please provide an example QB code coupled with both expected and current output.

PR #5836, opened 2019-09-19
Update Ubuntu version from Precise to Trusty on Travis

sapics:
Travis warns about old ubuntu version, now.
https://travis-ci.org/bcit-ci/CodeIgniter/jobs/584604656
Actually, the support of Precise(12.04) is ended at April 2017.
It might be better to update Xenial(16.04), but php5.4 and php5.5
are not supported in Xenial, thus, this PR updates to Trusty(14.04)
which is supported until April 2022.
https://docs.travis-ci.com/user/reference/xenial/#php-support
https://wiki.ubuntu.com/Releases.

By updating to Trusty, we could test 7.4snapshot version and 
8.0.0-dev version easily, thus I added them.

This pr also replace `php:hhvm` to `php:hhvm-3.30`.
Because composer is not support hhvm any more as
https://github.com/composer/composer/issues/8127,
thus, composer cause the error in hhvm-4.
It would be better to test in hhvm-3.30 which is the
last php support version of hhvm.
https://hhvm.com/blog/2018/12/17/hhvm-3.30.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ php:
   - 7.3
   - 7.4snapshot
   - master
-  - hhvm
+  - hhvm-3.30
   
 env:
   - DB=mysql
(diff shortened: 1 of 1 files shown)


narfbg:
Looks solid :+1: 

Does trusty have 5.3 though, so I can merge this into 3.1-stable as well?

sapics:
Thanks!

Oh sorry, I did not know that stable-3.1 branch supports PHP5.3.

PHP5.3 is not supported on Trusty in travis, unfortunately.
For supporting PHP5.3, it would be better to add precise by matrix method.
https://docs.travis-ci.com/user/reference/trusty#php-images

I have created #5839 for supporting PHP5.3 to `3.1-stable` branch.

PR #5846, opened 2019-09-26
cache: redis: Check version of phpredis and redis

nebkat:
Some distributions use redis rather than phpredis, so check for both. phpversion() returns false if the extension is not enabled.

### DIFF
--- a/system/libraries/Cache/drivers/Cache_redis.php
+++ b/system/libraries/Cache/drivers/Cache_redis.php
@@ -98,7 +98,7 @@ class CI_Cache_redis extends CI_Driver
 			return;
 		}
 
-		isset(static::$_delete_name) OR static::$_delete_name = version_compare(phpversion('phpredis'), '5', '>=')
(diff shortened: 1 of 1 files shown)


narfbg:
Please elaborate ... If there are Redis libraries with different names out there, I would expect that they are not one and the same.

nebkat:
I have only encountered "redis" on my installations but I see that it is different between https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Session/drivers/Session_redis_driver.php#L119 and 
https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Cache/drivers/Cache_redis.php#L101. I could add the same check to both, or one of them should be removed?

narfbg:
Hmm, you know what? The 'phpredis' one might actually be wrong altogether and my mistake too - that check in the Cache driver was supposed to replicate similar changes to the Session driver submitted by another contributor.

So, no need for the ternary operation, just rename 'phpredis' to 'redis'.

nebkat:
See #5861

PR #5852, opened 2019-10-05
fix issue on session_regenerate with PHP 7.0 or greater

pjsde:
If the PHP version in use is 7 or greater, then there is a problem when doing session_regenerate(), because when we call read() it always affect $this->sessionID = $sessionID which means that in write() function you can't detect a $sessionID regenerate, because when we get there the $this->sessionID is always equal to $sessionID.

This happen because in PHP 7 when session_regenerate() is call, this function executes internally and in the following order:
- write() (with the initial session;
- close();
- open();
- read() (already with the new session).

With this PR the session will be prepared for PHP 7 or greater and for PHP 5

### DIFF
--- a/system/libraries/Session/drivers/Session_database_driver.php
+++ b/system/libraries/Session/drivers/Session_database_driver.php
@@ -159,7 +159,17 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 		$this->_db->reset_query();
 
 		// Needed by write() to detect session_regenerate_id() calls
-		$this->_session_id = $session_id;
+		if (is_php('7.0'))
+		{
+			if(is_null($this->_session_id))
+			{
+				$this->_session_id = $session_id;
+			}
+		}
+		else
+		{
+			$this->_session_id = $session_id;
+		}
 
 		$this->_db
 			->select('data')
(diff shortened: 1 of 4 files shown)


narfbg:
How did you determine that there's such a difference? I see no reason why `close()` and `open()` would only be called in on 7+ in that sequence.

pjsde:
Tha's because as you can see in the [manual](https://www.php.net/manual/en/function.session-regenerate-id.php) the behavior change since version 7.0 to saves old session data before closing.

**Changelog**
> 
Version | Description
-- | --
7.0.0 | session_regenerate_id() saves old session data         before closing.
5.1.0 | Added the delete_old_session parameter.

So if you use the PHP 7.0 or greater than it will have the behavior as described above

narfbg:
OK, but ... if `read()` is called beforehand, that would mean the `$this->_session_id !== $session_id` in `write()` should never succeed with this sequence, so I'm still puzzled as to how this fixes anything.

pjsde:
This fix that because it will only affect $this->_session_id to $session_id if $this->_session_id is null in read() which means that when we get the write() the difference  will be catched

narfbg:
Sorry, but I still don't get it. Here's how I see it currently working:

- PHP 5
    - `read()` sets `$this->_session_id`.
    - `write()` catches the difference between `$session_id` and `$this->_session_id` and thus determines that we have regeneration.
- PHP 7:
    - `read()` sets `$this->_session_id`.
    - `write()` is called with the *old* `$session_id` and thus catches no difference, BUT according to the changelog you referenced, this is exactly what's supposed to happen as the old session needs to be saved before regeneration.
    - `close()` on the *old* session and `open()` on the *new* one are subsequently called.
    - `read()` is called with the *new* `$session_id` and overwrites `$this->_session_id` with it.
    - `write()` is called with the *new* `$session_id` and catches no difference, due to the overwrite mentioned in the last step. At this point we're already operating on the already regenerated session "pointer", so this makes no difference.

So ... I suppose you are correct in saying that the regeneration isn't detected by `write()` on PHP 7, but that by itself means practically nothing as it doesn't need to do anything different. Or am I missing something?

Issue #5843, opened 2019-09-25
Exclude user_guide if CodeIgniter is installed via Composer.


indrakaw:
This is probably similar to #3831.

If possible, I'd to exclude `user_guide` directory to keep the file size when using composer down. It's quite big, compared to core directories:

**Installed with `--prefer-dist`**
```
$ du -h --max-depth=1 | sort -h
280K    ./application
2.4M    ./system
11M     ./user_guide
13M     .
```

**Installed with `--prefer-source`**
```
$ du -h --max-depth=1 | sort -h
280K    ./application
2.4M    ./system
11M     ./user_guide
55M     ./.git
67M     .
```

I'm in favor of ignoring it, The idea is that when using the library as is on a prod server, you don't end up with files that can be used for full path disclosure.
https://www.owasp.org/index.php/Full_Path_Disclosure

Addendum:
My thoughts were that when you are using this (or any other package), I would typically look through the code and the documentation online instead of reading through a docs folder in the vendor directory.

narfbg:
Well, how do you propose to do that? The `user_guide` dir is included in the tag ...

indrakaw:
I'm not familiar with composer packages. Normally, it can be excluded by add this line:
```
user_guide export-ignore
```
Into `.gitattributes` file.

djunehor:
I agree with @indrakaw, installation via composer should be fast and light-weight, which means non-required files should be excluded. As suggested, the folder should be included in `gitignore` and an update pushed so that subsequent composer installations don't down user_guider folder.

indrakaw:
Apparently all `3.X` tags are updated with the current `.gitattributes`. I was trying to test by installing older version before one from issue I mentioned then compare it with the current one.

This might take while, I'm installing the package via composer as commit as version pointer.

gphg:
Hi. Is it possible to get `user_guide_src` excluded too?

PR #5859, opened 2019-10-18
Exclude user_guide directory from composer installation by GitAttributes.

indrakaw:
Exclude `user_guide` directory if package is installed with composer with the --prefer-dist flag set.

Reff: https://github.com/bcit-ci/CodeIgniter/issues/5843

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,3 +20,4 @@ tests/travis/ export-ignore
 
 # User Guide Source Files
 user_guide_src
+user_guide
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,4 +20,6 @@ tests/travis/ export-ignore
 
 # User Guide Source Files
 user_guide_src
-user_guide
+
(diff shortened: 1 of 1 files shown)


narfbg:
OK ... is that all? If so, why is the PR marked as a Draft/WIP?

indrakaw:
Hello. This is my first attempt and I had no idea what do I do.

Should I pull it or are there certain conditions?

I'm helpless.

narfbg:
Well, apparently "Ready for review" is what you needed to do indeed.

I don't recall anyone else, including first-time contributors, having to do that extra step, so you must've somehow marked your PR as WIP manually to begin with. If you submit another PR in the future, just use whatever the defaults are and we're able to manage everything else easily after that.

That being said, I tested how this would work with our normal release procedure and it works fine so :+1:
Thanks for your contribution.

indrakaw:
Thanks.

PR #5866, opened 2019-10-26
Update welcome_message.php

therinji:
I think it will no matter if you don't use this lang="en"


jim-parry:
The lang attribute doesn't detract, and your other change is trivial - change for the sake of change.

PR #5874, opened 2019-11-21
Samesite attribute for session cookie.

campio97:
Added the option to set the samesite attribute for the session cookie. Default is "Strict"

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -366,6 +366,17 @@ $config['encryption_key'] = '';
 | Other session cookie settings are shared with the rest of the application,
 | except for 'cookie_prefix' and 'cookie_httponly', which are ignored here.
 |
+| 'sess_cookie_samesite'
+|
+| SameSite prevents the browser from sending this cookie along with cross-site requests.
+| There are three possible values for the same-site attribute:
+| 'Lax'
+|			-	Some cross-site usage is allowed. 
+| 'Strict' 
+|			-	The cookie is withheld with any cross-site usage. 
+| 'None'
+|			-	Clearly communicate you intentionally want the cookie sent in a third-party context.
+|
 */
 $config['sess_driver'] = 'files';
 $config['sess_cookie_name'] = 'ci_session';
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -374,6 +374,7 @@ $config['sess_save_path'] = NULL;
 $config['sess_match_ip'] = FALSE;
 $config['sess_time_to_update'] = 300;
 $config['sess_regenerate_destroy'] = FALSE;
+$config['sess_cookie_samesite'] = 'Strict'; // Work only on php 7.3 or later
 
 /*
 |--------------------------------------------------------------------------
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -144,15 +144,27 @@ class CI_Session {
 		// unless it is being currently created or regenerated
 		elseif (isset($_COOKIE[$this->_config['cookie_name']]) && $_COOKIE[$this->_config['cookie_name']] === session_id())
 		{
-			setcookie(
-				$this->_config['cookie_name'],
-				session_id(),
-				(empty($this->_config['cookie_lifetime']) ? 0 : time() + $this->_config['cookie_lifetime']),
-				$this->_config['cookie_path'],
-				$this->_config['cookie_domain'],
-				$this->_config['cookie_secure'],
-				TRUE
-			);
+			if(version_compare(PHP_VERSION, '7.3.0') >= 0){
+				setcookie($this->_config['cookie_name'], session_id(), 
+					  [
+					  	'samesite' 	=> $this->_config['sess_cookie_samesite'],
+					  	'secure' 	=> $this->_config['cookie_secure'],
+					  	'domain' 	=> $this->_config['cookie_domain'],
+					  	'path' 		=> $this->_config['cookie_path'],
+					  	'expires' 	=> (empty($this->_config['cookie_lifetime']) ? 0 : time() + $this->_config['cookie_lifetime']),
+					  	'httponly'	=> TRUE
+					  ]);
+			}else{
+				setcookie(
+					$this->_config['cookie_name'],
+					session_id(),
+					(empty($this->_config['cookie_lifetime']) ? 0 : time() + $this->_config['cookie_lifetime']),
+					$this->_config['cookie_path'],
+					$this->_config['cookie_domain'],
+					$this->_config['cookie_secure'],
+					TRUE
+				);
+			}
 		}
 
 		$this->_ci_init_vars();


application/config/config.php
```diff
@@ -374,6 +374,7 @@
 $config['sess_match_ip'] = FALSE;
 $config['sess_time_to_update'] = 300;
 $config['sess_regenerate_destroy'] = FALSE;
+$config['sess_cookie_samesite'] = 'Strict'; // Work only on php 7.3 or later
```
mmihalev: Maybe put some info about the available values?

application/config/config.php
```diff
@@ -366,6 +366,17 @@
...
 |
+| 'sess_cookie_samesite'
+|
+| SameSite prevents the browser from sending this cookie along with cross-site requests.
+| There are three possible values for the same-site attribute:
+| 'Lax'
+|			-	Some cross-site usage is allowed. 
+| 'Strict' 
+|			-	The cookie is withheld with any cross-site usage. 
+| 'None'
```
mmihalev: I'm a little bit confused about the "None" option. I can't find it documented in the official php documentation - https://www.php.net/manual/en/function.setcookie.php
They seems to document only support for "Strict" and "Lax".

campio97: A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None`. For more details[https://www.chromestatus.com/feature/5088147346030592]( https://www.chromestatus.com/feature/5088147346030592)
image

mmihalev: > A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None`. For more detailshttps://www.chromestatus.com/feature/5088147346030592
> image

Yes , I'm familiar with this. However, I can't find anything about the "None" value. Nothing in the PHP RFC - https://wiki.php.net/rfc/same-site-cookie and nothing in the PHP Official Docs - https://www.php.net/manual/en/function.setcookie.php

That's why I'm confused :)

sonnlv: I tried this but my code not working. 
here is my config
image
image
image
help me pls.

campio97: > > A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None`. For more detailshttps://www.chromestatus.com/feature/5088147346030592
> > image
> 
> Yes , I'm familiar with this. However, I can't find anything about the "None" value. Nothing in the PHP RFC - https://wiki.php.net/rfc/same-site-cookie and nothing in the PHP Official Docs - https://www.php.net/manual/en/function.setcookie.php
> 
> That's why I'm confused :)

Yes i know, i looked for this info too, but without result. Google said that will be required in April. So i think they will insert this info into the documentation

campio97: > I tried this but my code not working.
> here is my config
> image
> image
> image
> help me pls.

Can you post your Session and Cookie config please?

sonnlv: > Can you post your Session and Cookie config please?
you mean
image
 right?
please let me know how to fix this solution.

campio97: Google chrome say that if you want set the cookie SameSite = "None" you have to set the config "cookie_secure" = TRUE. 

> A future release of Chrome will only deliver cookies with cross-site requests if they are set with SameSite=None **and Secure**.

sonnlv: I did, but it not work even i upgraded php version to 7.3 to use new your commit.
image
image

narfbg:
OK, a few things ...

1. The patch needs to follow our [styleguide](https://codeigniter.com/user_guide/general/styleguide.html).
2. Why should 'Strict' be the default? I too want maximum security by default, but this is bound to trigger a wave of issue reports by people who haven't paid attention to this. And it would be especially confusing if (as it would be with the current patch) a PHP version change fundamentally alters this behavior, which leads me to the next point ...
3. There's a way to send cookies without `setcookie()` - use `header()`, making it possible to make the feature non-dependant on the PHP version, which we strive to do. This will also solve the problem with `SameSite=None`, which just won't work via `setcookie()` PHP's doesn't support it.

campio97:
> OK, a few things ...
> 
> 1. The patch needs to follow our [styleguide](https://codeigniter.com/user_guide/general/styleguide.html).
> 2. Why should 'Strict' be the default? I too want maximum security by default, but this is bound to trigger a wave of issue reports by people who haven't paid attention to this. And it would be especially confusing if (as it would be with the current patch) a PHP version change fundamentally alters this behavior, which leads me to the next point ...
> 3. There's a way to send cookies without `setcookie()` - use `header()`, making it possible to make the feature non-dependant on the PHP version, which we strive to do. This will also solve the problem with `SameSite=None`, which just won't work via `setcookie()` PHP's doesn't support it.

Thank you for your feedback.
1. Ok for the styleguide, i'll be more accurate.
2. So probably the best thing is to use a null value as default.
3. I know we can use the `header()` function, but this brings many complications. By default in the Session.php file there is this config `ini_set('session.use_cookies', 1);`, so the cookie is always created by the function `session_start()`. In php 7.3 has been added the option for **samesite**. Here the changelog of the `set_cookie_params()` function: `"7.3.0 | An alternative signature supporting an options array has been added. This signature supports also setting of the SameSite cookie attribute."`
I think the best solution is make the patch compatible only with 7.3 version. If you agree I will develop it accordingly. if you prefer to make this feature compatible with previous versions, I am open to suggestions.

system/libraries/Session/Session.php
```diff
@@ -144,15 +144,27 @@ public function __construct(array $params = array())
...
+					  	'path' 		=> $this->_config['cookie_path'],
+					  	'expires' 	=> (empty($this->_config['cookie_lifetime']) ? 0 : time() + $this->_config['cookie_lifetime']),
+					  	'httponly'	=> TRUE
+					  ]);
+			}else{
+				setcookie(
+					$this->_config['cookie_name'],
+					session_id(),
+					(empty($this->_config['cookie_lifetime']) ? 0 : time() + $this->_config['cookie_lifetime']),
+					$this->_config['cookie_path'],
```
ytetsuro: @narfbg @campio97

> There's a way to send cookies without setcookie() - use header(), making it possible to make the feature non-dependant on the PHP version, which we strive to do. This will also solve the problem with SameSite=None, which just won't work via setcookie() PHP's doesn't support it.

How about making this line the following code?
But that may not be a good solution.

```php
sprintf('%s"; samesite="%s', $this->_config['cookie_path'], $this->_config['sess_cookie_samesite']);
```

see: https://github.com/nbs-system/snuffleupagus/issues/184

dont-panic-42:
This code does not work for me - I get:

    Message: Undefined index: sess_cookie_samesite
    Filename: Session/Session.php
    Line Number: 166

Line 166 is the new line in question:

    'samesite' 	=> $this->_config['sess_cookie_samesite'],

Tracing the code, the `$this->_config` array is set up in the [`_configure()` method](https://github.com/bcit-ci/CodeIgniter/blob/d77fb0d55ac51b9655f0f14c20130fecfc965927/system/libraries/Session/Session.php#L254) beginning on line 254.  

That is set up to populate `$this->_config` with various defaults if they were not specified in the `$params` array passed to the Session constructor.  In my case `$params` is empty, and as `_configure()` has not been updated for this, then of course the new config item `config_item('sess_cookie_samesite')` never makes it into `$this->_config`.

Am I missing something?  Shouldn't `_configure()` also be updated to set up that element of the `_config` array?

PhoenixFnX:
> This code does not work for me - I get:
> 
> ```
> Message: Undefined index: sess_cookie_samesite
> Filename: Session/Session.php
> Line Number: 166
> ```

Same issue for me. I also patched the `__construct()` method with this `$param` set just before the `_configure($params)` call : 
```
		// Configuration ...
		isset($params['sess_cookie_samesite']) OR $params['sess_cookie_samesite'] = config_item('sess_cookie_samesite');
		$this->_configure($params);
```

OskarsOzols:
This patch doesn't address issue of Safari iOS version 12.x bug https://bugs.webkit.org/show_bug.cgi?id=198181  when setting SameSite to 'None' actually is interpreted as 'Strict'. So for this specific iOS version SameSite attribute should not be applied.

gxgpet:
Ok, so I think this should be addressed as soon as we can. Here go my thoughts on it...

1. Indeed, the patch needs to follow the [style guide](https://codeigniter.com/userguide3/general/styleguide.html).
2. I would say that the issue should be, though, fixed in separated PRs per library/feature. Those should be: `CI_Session`, `CI_Input`, and `CI_Security` (yes, we have to address it for the CSRF feature as well).
3.  I would advocate for the default value to be `Lax`, and not `Strict`. Exactly how the cookies worked "before", when the attribute value was missing. 
4. While we could use the "PHP bug hack", a `header()` call would ensure that the same code will work for all PHP versions, thus reducing complexity. @campio97, why are you saying that `ini_set('session.use_cookies', 1);` is preventing us from using `header()`? I made a super-fast test on `CI_Session` by using the `header()` instead of `setcookie()` and it worked as expected; no duplicated `Set-Cookie` headers were sent. Am I missing something here?
5. The config value should be under cookie's settings section, and named something like `cookie_samesite`, and not `sess_cookie_samesite`. While this should specify a global cookie `samesite` policy for the entire CI app, `CI_Input::set_cookie()` should allow this value to be changed when required, by adding a method parameter there.

@OskarsOzols we can't implement a browser version-specific fix into CI's core... Upgrading the browser should fix it.

narfbg:
1a2651040ef701e750b1c13cd69cc70814b079d0

Issue #5868, opened 2019-11-07
Query Builder is missing FULL OUTER join type


TalonTR:
The `join()` method in `/system/database/DB_query_builder.php` limits the usable join types with the following array:
```
if (! in_array($type, ['LEFT', 'RIGHT', 'OUTER', 'INNER', 'LEFT OUTER', 'RIGHT OUTER'], true)) {
```
This is true for Codeigniter 4 as well. `FULL OUTER` is missing from this list. Is there any reason why?

Source: https://en.wikipedia.org/wiki/Join_(SQL)

narfbg:
My knowledge about this is a bit rusty, but isn't `OUTER` the same as `FULL OUTER`?

Zigi84:
`OUTER` is optional. It's mentioned in wikipedia article.
Possibly wants A\B U B\A, or A without B union B without A?

TalonTR:
> My knowledge about this is a bit rusty, but isn't OUTER the same as FULL OUTER?

No. There is no mention of an outer-only join type in the form of `SELECT * FROM a OUTER JOIN b` (aside from an IBM Informix speciality).

> OUTER is optional. It's mentioned in wikipedia article.

Exactly. `OUTER` is optional. `FULL` is not! 

Going through the valid options within the QueryBuilder Array:
* `LEFT` ist the same as `LEFT OUTER` and both is valid SQL. 
* `RIGHT` ist the same as `RIGHT OUTER` and both is valid SQL. 
* `INNER`, too, is valid SQL.
* **Its counterpart should be `FULL` (or `FULL OUTER`) - not `OUTER`.**

`OUTER` alone is no valid SQL. The `FULL` option is missing completely in QueryBuilder. 

PS:
* A full-join can be reproduced by a union of a left- and a right-join, buts thats not the point.
* The `OUTER` entry within the aforementioned array seems to be wrong, but this is also of no importance to me.

TalonTR:
I proposed a pull request for this issue: https://github.com/bcit-ci/CodeIgniter/issues/5868

TalonTR:
Is here anything missing?

narfbg:
Possibly adjustments to the docs, though I'm not sure ... Sorry about that and about the huge delay on my part, but I'm very short on real free time right now and restrictive policies at my job don't make it easy either. :/
I'll try to get back to this and other stuff in the backlog in a couple of days.

TalonTR:
[...]

PR #5881, opened 2020-01-02
Added FULL OUTER JOIN to QueryBuilder

TalonTR:
fixes https://github.com/bcit-ci/CodeIgniter/issues/5868


jamieburchell:
Does it matter that this regex now matches `FULL INNER JOIN` which is invalid?

PR #5883, opened 2020-01-08
image/webp for Image lib 

adevesa:
This changes able to use images that has image webp encoding.


narfbg:
Cool, but 'image/webp' should not be associated with 'jpeg', 'jpg' and 'jpe' filename extensions in config/mimes.php - please revert that part; the rest looks fine. 👍

adevesa:
Fixed MIMEs

PR #5886, opened 2020-01-13
The problem of input class case sensitivity is solved in Unix systems…

hhuseyincihangir:
…. - Form_validation, Cookie problems that occur in stacks like Lampp etc. installed in Unix systems.

Signed-off-by: Hasan Hüseyin CİHANGİR <hhuseyincihangir@gmail.com>

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -214,7 +214,7 @@ if ( ! function_exists('is_loaded'))
 
 		if ($class !== '')
 		{
-			$_is_loaded[strtolower($class)] = $class;
+			$_is_loaded[mb_strtolower($class, 'UTF-8')] = $class;
 		}
 
 		return $_is_loaded;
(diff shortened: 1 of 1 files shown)


narfbg:
This has nothing to do with UNIX systems and certainly doesn't solve any problems related to the built-in Form Validation library or cookies.

The only potential problem this patch may solve is issues with custom class/library names written with non-Latin characters, and ... while someone may be able to present a potential case for that, it's a bad practice.

hhuseyincihangir:
If it's not a problem, why am I getting this error even though I haven't changed any standard libraries? Bitnami Lamp Stack and Xampp also have the normal httpd I get the same error. The Input class properties invoked in the form_validation library and in the cookie_helper are invoked as lowercase. It corrects when I initialize "I" in the called fields. I get the same error when I try this on a linux mint in English. Also, using "mb_strtolower" instead of "strtolower" does not mean support in more languages? Is it another problem?

narfbg:
I have absolutely no idea what you're talking about mate ... You're talking of some error, which I have never seen and makes no sense at all. Don't know what this has to do with backwards compatibility either. The only thing I can respond to is this:

> Also, using "mb_strtolower" instead of "strtolower" does not mean support in more languages?

Stock CI library names are all written in ASCII, which is what `strtolower()` handles natively. The only reason to use `mb_strolower()` instead of `strtolower()` is if the string being manipulated is not an ASCII one. The entire point of the multibyte extension (`mb_*()` function family) is to provide support for different charsets and therefore different languages too. That's all they do.

PR #5887, opened 2020-01-21
Add PDO support on Backup Utility

kornarakis:


narfbg:
Sorry, but no - this is obviously MySQL centric, which beats the entire goal of PDO; i.e. an abstraction interface that works for all databases.
On another note, you also don't need PDO with CI or any other framework that builds more abstractions on top of it. MySQLi is way better.

Issue #5879, opened 2019-12-18
Array and string offset access syntax with curly braces is deprecated


slps970093:
Array and string offset access syntax with curly braces is deprecated

`        $this->output->enable_profiler(TRUE); `

PHP Version 7.4

![圖片](https://user-images.githubusercontent.com/8385770/71050311-d6692680-217f-11ea-848f-38fc00a82dae.png)

GLedoussal:
Can reproduce with PHP 7.4. I don't understand why this is happening, they deprecated string offset access and this is just a dynamic property name. Is this a PHP bug ?

As a workaround you can replace that line with `$this->{'_compile_' . $section} = TRUE;` and change other lines using curly braces (138 and 557)

TalonTR:
I was pretty confused for some time by this, too - took my a while to figure it out. This notation actually creates a public array property `$this->_compile_` which is used as an associative array - keys from `$_available_sections`, `true` or `false` values. 
The correct notation would simply be `$this->_compile_[$section]`. The deprecated curly braces are used three times throughout the file.

TalonTR:
I created a PR: https://github.com/bcit-ci/CodeIgniter/pull/5888

TalonTR:
@narfbg wrote in https://github.com/bcit-ci/CodeIgniter/pull/5888
> Sorry, but your patch does the opposite of what it should be doing ... see 4fc777a for the correct fix.

Your fix is ok, but i´m pretty sure that you are mistaken on that. I totally understand how you arrived at this misconception, thats one of the reasons PHP deprecated this notation. The previous code `$this->_compile_{$section}` implicitly creates an array `$this->_compile` and then sets values by using $section as key.  Your fixed code now created multiple properties with different names, each concatenations of `_compile_` and `$section`. 

Nevertheless thanks for fixing this issue by removing the deprecated notation 👍

narfbg:
I'm not mistaken.

For one, we'd never name an array `_compile_`; then there's the matching `_compile_*()` method names and last but not least, [it's practicly documented](https://github.com/bcit-ci/CodeIgniter/blob/4fc777a92942eee78f72cfc86453942bc21da020/system/libraries/Profiler.php#L121).
As a side note, if I was wrong, then while the code would technically still work, simply making a notice go away is not a fix if the code doesn't do what it was intended to. :)

Props for being vigilant though.

TalonTR:
> For one, we'd never name an array _compile_

[...]

PR #5888, opened 2020-01-22
updated deprecated array access in Profiler.php

TalonTR:
https://github.com/bcit-ci/CodeIgniter/issues/5879


narfbg:
Sorry, but your patch does the opposite of what it should be doing ... see 4fc777a92942eee78f72cfc86453942bc21da020 for the correct fix.

PR #5890, opened 2020-01-24
Adds a public trans_depth function to determine transaction state

jerkob:
Currently there is no mechanism to determine if a transaction is currently active. This function, by allowing read-only access to the protected `_trans_depth` property, can be used to determine if a transaction is currently active, if the transaction is nested, and how many levels of nesting are in effect.

For my personal use case I only need to know if a transaction is active or not, however rather than create a `trans_active()` function or similar I felt it was better to just return the underlying `_trans_depth` as there may be cases where determining if a transaction is nested is useful. It also provides for an obvious function name and I was happy to not have to choose the right name for a `trans_active`-type function.

### DIFF
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -829,13 +829,13 @@ abstract class CI_DB_driver {
 	// --------------------------------------------------------------------
 
 	/**
-	 * Lets you retrieve the transaction depth
+	 * Returns whether a transaction is currently active
 	 *
-	 * @return	int
+	 * @return	bool
 	 */
-	public function trans_depth()
+	public function trans_active()
 	{
-		return $this->_trans_depth;
+		return (bool) $this->_trans_depth;
 	}
 
 	// --------------------------------------------------------------------
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -825,6 +825,18 @@ abstract class CI_DB_driver {
 	{
 		return $this->_trans_status;
 	}
+	
+	// --------------------------------------------------------------------
+
+	/**
+	 * Lets you retrieve the transaction depth
+	 *
+	 * @return	int
+	 */
+	public function trans_depth()
+	{
+		return $this->_trans_depth;
+	}
 
 	// --------------------------------------------------------------------
 
--- a/system/database/DB_driver.php
+++ b/system/database/DB_driver.php
@@ -829,7 +829,7 @@ abstract class CI_DB_driver {
 	// --------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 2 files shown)


narfbg:
Would you explain your use case?

jerkob:
@narfbg Sure. We have a system that tracks all changes to data, and all data changes within a transaction are grouped together with the same GUID. To accomplish this, every time we are about to do a query that changes data we check whether we are in a transaction and if so we maintain the same GUID. If we are not in a transaction, we generate a new one.

We've been doing it this way for a long time but with CodeIgniter 2 where the `_trans_depth` property wasn't protected so we could read it directly. We are now upgrading to CI 3 and want to be able to do this while keeping the variable protected.

narfbg:
OK, I guess that makes sense. Patch needs some modifications though, I'll leave inline comments.

system/database/DB_driver.php
```diff
@@ -825,6 +825,18 @@ public function trans_status()
 	{
 		return $this->_trans_status;
 	}
+	
+	// --------------------------------------------------------------------
+
+	/**
+	 * Lets you retrieve the transaction depth
+	 *
+	 * @return	int
```
narfbg: Nobody should be concerned with the nesting level, that's a leaky abstraction. Please make it a bool (just casting on return is fine).

narfbg: Of course, that also means changing the method name (and description), but not sure what to ... `trans_active()`?

user_guide_src/source/database/db_driver_reference.rst
```diff
@@ -161,6 +161,14 @@ This article is intended to be a reference for them.
 
 		Lets you retrieve the transaction status flag to
 		determine if it has failed.
+		
+	.. php:method:: trans_depth()
+
+                :returns:	The current transaction depth
```
narfbg: There's obviously an extra space here. Please just use tabs instead (everywhere).

system/database/DB_driver.php
jerkob: Makes sense, I made both these changes.

user_guide_src/source/database/db_driver_reference.rst
jerkob: Should be correct now. The function just above, `trans_status`, has the extra spaces and I copied that one. I decided not to correct that as part of this pull request just to keep it focused, bu would you want me to do a separate pull request to fix that?

narfbg: Good call and thanks for noting it.
I'll correct it myself as I'll be editing the docs to add a changelog entry myself anyway.

PR #5892, opened 2020-01-27
Create Session_memcache_driver.php

andreimosman:
Memcache driver using php-memcache module instead of php-memcached


narfbg:
Support for ext/memcache was intentionally not included, sorry.

PR #5895, opened 2020-01-28
Fix database connection leak when closing PDO connection

joshkel:
We discovered a resource leak (database connection leak) in our application's test suite. We investigated and discovered that, although the test suite was calling `CI_DB_driver::close()` after each test, the `result_id` member variable still maintained an internal reference to the database connection, which kept the connection from being closed.

This is using pdo_sqlsrv; I have not tested other PDO drivers.

This PR addresses the problem by clearing `result_id` within CodeIgniter's PDO driver's `_close` method, but if another approach would be better, please let me know.

### DIFF
--- a/system/database/drivers/pdo/pdo_driver.php
+++ b/system/database/drivers/pdo/pdo_driver.php
@@ -326,4 +326,17 @@ class CI_DB_pdo_driver extends CI_DB {
 		return 'TRUNCATE TABLE '.$table;
 	}
 
+	// --------------------------------------------------------------------
+
+	/**
+	 * Close DB Connection
+	 *
+	 * @return	void
+	 */
+	protected function _close()
+	{
+		$this->result_id = FALSE;
+		$this->conn_id = FALSE;
+	}
+
 }


narfbg:
Sorry, but this is honestly the first time I hear of a "connection leak" and am having a hard time understanding how it might be a problem. Care to elaborate?

On a side note, PHP itself is supposed to close any remaining connections at the end of script execution, so this might actually be a PHP bug.

Either way though, sounds more like an issue with the `result_id` handling and handling it in `close()` seems like an overkill solution.

joshkel:
Sure. Sorry for the lack of details.

We ran into this problem in our test suite, rather than when running the application normally. We're using PHPUnit plus [ci-phpunit-test](https://github.com/kenjis/ci-phpunit-test). As I understand it, PHPUnit creates one instance of the appropriate `TestCase` subclass for each test that's run, and ci-phpunit-test in particular apparently stores a [CodeIgniter instance](https://github.com/kenjis/ci-phpunit-test/blob/6a823bbc3a3bb0a7a7dc1cd0f5ada0b1d305c74e/application/tests/_ci_phpunit_test/CIPHPUnitTestCase.php#L44) within each subclass of its `CIPHPUnitTestCase` class. For each test, ci-phpunit-test [resets the CodeIgniter instance and closes any previous database connections](https://github.com/kenjis/ci-phpunit-test/blob/6d813b32482822b7dc9c97be02c742a2dde3a1df/application/tests/_ci_phpunit_test/functions.php#L25) - so previous tests may keep their own local references to previous CodeIgniter instances, but the database connections in particular are closed, so it ought to be okay. And it means that the test suite can't wait until PHP cleans up at the end of script execution, but as long as `CI_DB_driver::close()` works, it's met our needs very well.

We recently switched from CodeIgniter's sqlsrv driver to its pdo_sqlsrv driver (so that we could interoperate with a third-party library that expected a PDO object), and our test suite started failing with "Too many open files" errors. We found that the problem was a large number of TCP sockets open to SQL Server. Further investigation pinpointed the `result_id` member in particular as preventing `CI_DB_driver::close()` from actually closing the connection.

I looked at the `result_id` handling to see if it could be addressed there, but I don't see how. As long as `affected_rows()` can be called at any point after a query, it seems that the driver needs to keep `result_id`, and that leaves `close()` as the best place to make sure that all resources are, in fact, closed. If I'm missing something, please let me know.

Thanks for the quick reply. I can put together some sample code to demonstrate the problem if it would help.

narfbg:
OK, at least I get where you're coming from now.

Clearing `CI_DB_driver::$result_id` in `close()` is undoubtedly the easiest solution, but ... it's a fault caused by a third-party unit testing library, combined with a possible PHP/PDO bug, in turn combined with your own decision to switch to PDO because of an entirely different third-party library that you use.
Just in principle, if something is completely outside of CI's control it's also not CI's responsibility to handle it, but even if we ignore that, this "fix" is very likely just masking an undiagnosed problem somewhere else.

joshkel:
Thanks, but I really don't agree with that analysis. Even though the problem is triggered by a third-party testing library, the problem is within the `close()` method itself:

* `close()` is supposed to close the database connection.
* We've seen that, for one of CI's supported database drivers, when that driver is used in a normal, documented and supported fashion, it doesn't close that connection.
* PDO doesn't offer an explicit close operation, so the only way to close a connection is by unsetting any references to the PDO object.
* It's not a bug within PHP / PDO: a PDOStatement (the data type of `$result_id`) needs a server connection (for example, so that it can fetch the next row of a result set if needed), so as long as a PDOStatement exists, it may keep the connection open. This is [documented behavior](https://www.php.net/manual/en/pdo.connections.php):
    > The connection remains active for the lifetime of that PDO object. To close the connection, you need to destroy the object by ensuring that all remaining references to it are deleted—you do this by assigning NULL to the variable that holds the object... If there are still other references to this PDO instance (such as from a PDOStatement instance, or from other variables referencing the same PDO instance), these have to be removed also (for instance, by assigning NULL to the variable that references the PDOStatement).
* As a result of PDO's design, there may be circumstances in which `close()` can't ensure that the database connection is closed. (For example, if a piece of code is storing a `CI_DB_result` somewhere, there's nothing that the `close()` can do about that.) But it should at least do what it can to actually free the database resources that it is aware of, including `$result_id`.
* Invalidating `$result_id` within `close()` is consistent with the observed behavior of other CI drivers. For example, for the sqlsrv driver, if I `close()` the connection and then try to call `affected_rows()` (which uses `$result_id` internally), I get back a PHP error: "Warning: sqlsrv_rows_affected(): supplied resource is not a valid ss_sqlsrv_stmt  resource." So the specific behavior might be different, but the general result ("using the result ID after closing the database is undefined behavior") would be the same.

Based on the that, it seems to me that this PR is the proper design: `close()` should do what it can to properly close the database connection by properly freeing the database connection resources that it has. If it doesn't, that's a bug within `close()`, and it's under CI's control and part of CI's responsibility, rather than a misdiagnosed problem somewhere else. And, since other CI drivers already demonstrate undefined behavior from trying to use `$result_id` after calling `close()`, this PR should be low-risk and consistent with other drivers.

Thank you.

kenjis:
I agree with @joshkel .

I expect the method `close()` closes the database connection, because the name tells so.

And one driver closes the connection, but another driver does not close actually.
That is not consistent.

narfbg:
Well, that is much more convincing than references to ci-phpunit-test. If you tell me that you're only experiencing a problem with a third-party package, I'm 99% sure to conclude that it is the third-party package causing the problem. :)

Still, have you noticed that half of the points you listed are basically issues with PDO's design? That's why I keep referring to this as a possible PHP/PDO bug, and am still not convinced that `close()` needs to do anything to `$result_id` ... it's just not its job to do garbage collection and more in particular I disagree with this statement of yours:

> But it should at least do what it can to actually free the database resources that it is aware of, including `$result_id`.

As a general rule of thumb, I don't attempt to write code that is by definition imperfect and cannot ever be guaranteed to do its job. And the "undefined behavior" after `close()` triggered by other drivers are pretty consistent with that too - it's not actually undefined behavior; we know the operation will fail, because the resource they refer to is dead - problem solved. This has never even been a consideration for CI.

For whatever reason though, PDO will bend over backwards to maintain a connection and completely shifts the burden of tracking every reference onto you. There's 2 ways to go about that - pass on that responsibility to the user, as it is evidently intended to; or try to abstract it away so you don't have to deal with it and ... I have to repeat myself like that, but trying to solve an unsolvable problem is futile.

joshkel:
Thanks for the reply. I really appreciate your taking the time to discuss.

So...

Like you, I wish that PDO had an explicit close operation. I consider that a shortcoming of its design. Although it seems odd to equate "design shortcoming" with "a possible bug." After all, there are aspects of CodeIgniter that I wish were designed differently, and there are aspects of my own code that I wish were designed differently. :slightly_smiling_face: Technical debt happens, tradeoffs that made sense at one time may no longer make sense, etc. And, of course, having to work with a flawed API or less-than-ideal design doesn't excuse me from doing my best to make the code that I'm responsible for work.

You said that there are only two approaches CI could take: pass on that responsibility to the user, or try to abstract it away. But CI already made the decision to try to abstract it: that's what its database layer is. :slightly_smiling_face: Sure, it's true that preserving that abstraction in all circumstances is an unsolvable problem, but that's nothing new: [all abstractions leak](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/). An abstraction layer like CI's database layer (or any of the examples from Joel Spolsky's article) can still have enormous value by preserving the abstraction where it easily can.

I can understand saying that users are responsible for tracking database references, but `$result_id` is created by CI without any knowledge or action on behalf of the user, and it isn't documented in the user guide, and the fact that it's a PDOStatement instance appears to be an implementation detail. So I don't understand saying that CI provides an abstraction layer, except for this one scenario where it shifts full responsibility of understanding the abstracted internals to the user. And I don't understand saying that users are responsible for not only maintaining their own database references but also understanding references that are otherwise created and tracked internally by CI, when it would be very straightforward for CI to clean up those references itself.

I can understand not wanting to write code that can't be guaranteed to work. But, right now, the CI PDO driver's `close()` method is guaranteed to _not_ work. The PDO API documentation says how to manage connections (to close a connection, clear all references to connection-related objects), and the CI PDO driver does the opposite (by creating and keeping a reference to a connection-related object even when trying to close the connection). I can understand saying that the PDO API isn't designed the way you and I wish it were, but I don't understand having CI code that doesn't comply with what the design of the PDO API is. I can understand saying that we can't solve every problem users might create (depending on what they do with PDO references), but I don't understand choosing not to solve a problem that we create ourselves (with the CI PDO driver's own `$result_id` reference).

I would understand it if the proposed changes added complexity or added unreliability or required a lot of extra code, because then there are non-trivial tradeoffs in terms of future maintenance or reliability. But this is a straightforward change (1-4 lines of code, depending on how you count the code for overriding the parent function), and as long as users aren't storing `CI_DB_result` instances anywhere and aren't mucking around with `$conn_id` or `$result_id` themselves, it's 100% reliable.

That's the point I was trying to make about undefined behavior: I could understand not wanting to include this change if it made the CI PDO driver act differently than other CI drivers or if it risked breaking backward compatibility. But, as far as I can understand, it makes it behave closer to other drivers, and I'd argue that any changes relating to backward-compatible behavior should be safe, because they fall under the category of undefined behavior. (And I think it's expected and good that you'd get unspecified bad results - undefined behavior - if you try to use a result after `close()`.)

Honestly, even setting aside the discussions of API design and resource management and abstractions, I'm having trouble understanding the opposition to this PR: it reliably addresses a known use case, it makes the code better comply with the API interfaces that it's using, and it does so without noticeably hurting the code's maintainability. That seems like a win.

narfbg:
> Although it seems odd to equate "design shortcoming" with "a possible bug."

Bad design naturally often leads to bugs and I said "possible" for a reason. Whatever, this is a moot point. :)

> You said that there are only two approaches CI could take: pass on that responsibility to the user, or try to abstract it away. But CI already made the decision to try to abstract it: that's what its database layer is. 🙂 Sure, it's true that preserving that abstraction in all circumstances is an unsolvable problem, but that's nothing new: [all abstractions leak](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/). An abstraction layer like CI's database layer (or any of the examples from Joel Spolsky's article) can still have enormous value by preserving the abstraction where it easily can.

Sorry, but those are false equivalences in my view.

A framework or library will by definition try to abstract, but that doesn't mean it will abstract absolutely everything and you can't default to such a conclusion.
Also, the referenced article doesn't say "all abstractions", but "all *non-trivial* abstractions" and even if we take that as undisputable truth, I don't see how that justifies any decision on our part.

> I can understand saying that users are responsible for tracking database references, but `$result_id` is created by CI without any knowledge or action on behalf of the user, and it isn't documented in the user guide, and the fact that it's a PDOStatement instance appears to be an implementation detail. So I don't understand saying that CI provides an abstraction layer, except for this one scenario where it shifts full responsibility of understanding the abstracted internals to the user.

It doesn't have to need full understanding of internals (more on that later) and there's plenty of instances where we don't attempt to do something that PHP itself doesn't provide easy access to.
Some things are just not worth it, others are practically impossible and in general the DB layer is only a proxy to underlying PHP APIs.

> And I don't understand saying that users are responsible for not only maintaining their own database references but also understanding references that are otherwise created and tracked internally by CI, when it would be very straightforward for CI to clean up those references itself.

It should be straightforward to make it so that you only need to take care of (in this case) `CI_DB_result` instances. Let's go back to something you said earlier:

>> I looked at the `result_id` handling to see if it could be addressed there, but I don't see how. As long as `affected_rows()` can be called at any point after a query, it seems that the driver needs to keep `result_id`, and that leaves `close()` as the best place to make sure that all resources are, in fact, closed. If I'm missing something, please let me know.

I've only looked at `CI_DB_pdo_driver`, but it does look like `affected_rows()` is the only place where `result_id` is needed outside of `CI_DB_result` and for that, we can simply cache the result of `PDOStatement::rowCount()` and just like that we no longer need to keep `result_id`.

> I can understand not wanting to write code that can't be guaranteed to work. But, right now, the CI PDO driver's `close()` method is guaranteed to *not* work. The PDO API documentation says how to manage connections (to close a connection, clear all references to connection-related objects), and the CI PDO driver does the opposite (by creating and keeping a reference to a connection-related object even when trying to close the connection).

That's a convincing argument, but not necessarily one for handling `result_id` *inside* `close()`.

> I can understand saying that the PDO API isn't designed the way you and I wish it were, but I don't understand having CI code that doesn't comply with what the design of the PDO API is.

Um, what? Isn't the entire problem that PDO is clearly designed to shift reference management responsibility onto the user? How is doing exactly that not complying with its design? Maybe I don't understand what you're trying to say here, but it's obviously not helping to support your argument.

> I can understand saying that we can't solve every problem users might create (depending on what they do with PDO references), but I don't understand choosing not to solve a problem that we create ourselves (with the CI PDO driver's own `$result_id` reference).

I feel like you're repeating yourself here and I already responded to this argument.

> I would understand it if the proposed changes added complexity or added unreliability or required a lot of extra code, because then there are non-trivial tradeoffs in terms of future maintenance or reliability. But this is a straightforward change (1-4 lines of code, depending on how you count the code for overriding the parent function), and as long as users aren't storing `CI_DB_result` instances anywhere and aren't mucking around with `$conn_id` or `$result_id` themselves, it's 100% reliable.

It has nothing to do with how straightforward the change is. It's about whether a problem needs to be addressed to begin with *and* whether that's the correct way to address it. And by extension whether the solution is effective, which as you correctly pointed out is entirely dependant on what the user does - my biggest issue with it.
But on another note, it's also about setting a precedent where we would now have to consider everybody wanting us to manage and cleanup object references for them, because we'd already be doing it. Even when limited to the DB classes, `close()` will fundamentally change from a simple proxy for `<insert_driver_name_here>_close()` into the one place to cleanup everything.
If this aspect wasn't important, there would be no talk of consistency, ever.

> That's the point I was trying to make about undefined behavior: I could understand not wanting to include this change if it made the CI PDO driver act differently than other CI drivers or if it risked breaking backward compatibility. But, as far as I can understand, it makes it behave closer to other drivers, and I'd argue that any changes relating to backward-compatible behavior should be safe, because they fall under the category of undefined behavior. (And I think it's expected and good that you'd get unspecified bad results - undefined behavior - if you try to use a result after `close()`.)

OK ...

> Honestly, even setting aside the discussions of API design and resource management and abstractions, I'm having trouble understanding the opposition to this PR: it reliably addresses a known use case, it makes the code better comply with the API interfaces that it's using, and it does so without noticeably hurting the code's maintainability. That seems like a win.

Well, for starters it's not necessarily opposition, but rather resistance - the difference is subtle, but important (the PR is still open, isn't it?) :)
Everything goes under scrutny. That might not be obvious sometimes, like when I just merge a pull request without having any discussion, but that doesn't mean it's not happening. I get that it is unpleasant at times, but I firmly believe that if CI would be an unmaintainable mess by now if I approached it any differently.

But now that you mention it, there's your use case too ... I didn't feel like it mattered much, but let's consider it for a second.
Firstly, you say that it reliably addresses it, while we already agreed that it's not a 100% reliable solution - please don't do that; though unintentional, it's mildly dishonest.
But more importantly, a known use case isn't necessarily a good use case. You're so far the only person with such a use case, and it's based entirely on a third-party library that wants to access a PDO object, yet you yourself said that `conn_id` and `result_id` are implementation details. And I don't know if you realize how correct you were in saying that - they are indeed intended for internal use only; the only reason they're public is because of legacy design decisions that preceed visibility declarations in PHP. To put it in other words - your particular use case shouldn't exist at all.

With all of that being said, I'm simply not convinced yet. On `conn_id` - sure, if nullifying it is the equivalent of another driver's `<name>_close()` function call, that's good enough. But I'm not convinced that `result_id` should be handled *there* as well, if at all.

joshkel:
I apologize if I repeated myself; I realized after I wrote that that I was probably getting too long-winded. I apologize. I'll admit I'm getting frustrated here.

> Well, for starters it's not necessarily opposition, but rather resistance - the difference is subtle, but important (the PR is still open, isn't it?) :)

That's a good point, and thank you. I really do appreciate your taking the time to engage.

And, to be clear, I'm good with design scrutiny - I'm just having trouble seeing how we're seeing this so differently.

> Firstly, you say that it reliably addresses it, while we already agreed that it's not a 100% reliable solution - please don't do that; though unintentional, it's mildly dishonest.

It reliably addresses this particular use case. (See below.) I'm not too interested in arguing there, but I do object to being told I'm being mildly dishonest. :slightly_smiling_face:

I wonder if you're focusing too much on my particular scenario of a third-party library wanting to access a PDO object? That's what motivated me to switch from the sqlsrv driver to the PDO driver, which is how I encountered this problem, but this problem exists within CodeIgniter itself. It would affect anyone who uses CI's PDO driver (without third-party libraries, and without directly accessing `$conn_id` or `$result_id`) and expects `close()` to free all database resources. _That's_ the use case that I'm trying to address. (I'll also note that I'm not the only one to encounter that sort of scenario - #5896 addresses the same issue for the Oracle driver.) As you said, it's not a common scenario, because database connections normally exist for the lifetime of the HTTP request and PHP automatically frees them once the request is done, but given that CI supports a `close()` method and supports multiple database instances, it seems like something that CI intends to support.

I realize I may be repeating myself, but to try to summarize my argument and reply to your latest points: (And I'll be a little verbose and list things in bullet points to hopefully make it easier to clarify where we might agree or disagree.)

* I understand you to be saying that CI shouldn't be in the business of managing and cleaning up users' object references. I agree.
* CI should manage and clean up _its own_ object references. That's usually good software design in general, and for PDO in particular, it's necessary to comply with PDO's design, because PDO links cleaning up resources with closing the connection.
* `$result_id` in particular is one of CI's own object references. CI creates it, users never have to interact with it, and it's not documented (besides a couple of lines of PHPDoc). So users shouldn't have to know about it, and CI should clean it up. In other words, it should be part of the abstraction that CI maintains.
* Similar to what you said, clearing `$conn_id` is somewhat equivalent to another driver's <name>_close() function call, and so the PDO `close()` method should do it.
* But the PDO docs explain that clearing `$conn_id` is only _somewhat_ equivalent to another driver's <name>_close() function. It's necessary but not sufficient: the PDO docs explain that closing a connection involves clearing any references to the PDO instance itself and any references to any PDOStatement objects for that PDO instance. That's the documented design of PDO, and so it's reasonable to treat clearing PDOStatement references as part of closing the database connection.
* It's not practical to guarantee that all PDO-related references are cleared. Maybe the user messed around with the public `$conn_id` or `$result_id` members; maybe the user is storing a `CI_DB_result` somewhere. But those are less common scenarios and difficult or impossible to solve. The CI PDO driver can and should clear PDOStatement instances that it creates and uses internally, because that fits PDO's design, preserves the abstraction (so users of CI don't have to worry about the CI PDO driver's implementation), and is easy to do (so no noticeable costs on code quality or maintainability).

I initially understood you to be saying that you thought CI should shift the responsibility of cleaning up `$result_id` onto whoever's using CI, and so that's part of what I was trying to address in my earlier reply. After reading your latest reply, it sounds like you agree with CI cleaning it up but are unconvinced of the mechanism. If I've misunderstood, please let me know.

> It's also about setting a precedent where we would now have to consider everybody wanting us to manage and cleanup object references for them, because we'd already be doing it. Even when limited to the DB classes, close() will fundamentally change from a simple proxy for <insert_driver_name_here>_close() into the one place to cleanup everything.

Of course, part of my argument is that, for PDO, acting as a proxy for <insert_driver_name_here>_close() is defined by the docs as involving clearing references to PDOStatements.

The precedent concern is, to some extent, a slippery slope argument: Just because there are good reasons for cleaning this particular object reference (because it's the only one that the CI PDO driver created and managed itself, because it's trivial to implement, because it's necessary for `close()` to behave correctly, because that's the behavior mandated by the PDO docs) doesn't mean that the rest of CI needs to do it.

> I've only looked at CI_DB_pdo_driver, but it does look like affected_rows() is the only place where result_id is needed outside of CI_DB_result and for that, we can simply cache the result of PDOStatement::rowCount() and just like that we no longer need to keep result_id.

If I understand correctly, you're proposing addressing the issue by removing the `$result_id` member variable completely and changing `CI_DB_driver::query` to retrieve and cache the number of affected rows, so that `CI_DB_driver::affected_rows()` can return it later?

I'd be concerned about that approach. Some database driver documentation (e.g., the "Note" [here](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlrowcount-function?view=sql-server-ver15), or the mysqli [affected_rows](https://www.php.net/manual/en/mysqli.affected-rows.php) / [num_rows](https://www.php.net/manual/en/mysqli-result.num-rows.php) documentation) explains that drivers may give different values depending on whether query results have been fetched yet. Guaranteeing that a design change like this would work correctly across all configurations and use cases of all database drivers seems hard and risky. And there are backward compatibility questions around removing the public `CI_DB_driver::$result_id` member (even if, as you said, it shouldn't be public) - sometimes breaking BC is best, but clearing `$result_id` within `close()` seems much simpler and safer (especially when clear-as-part-of-close is consistent with the PDO's documentation).

Thanks again.

narfbg:
>> Firstly, you say that it reliably addresses it, while we already agreed that it's not a 100% reliable solution - please don't do that; though unintentional, it's mildly dishonest.
>
> It reliably addresses this particular use case. (See below.) I'm not too interested in arguing there, but I do object to being told I'm being mildly dishonest. 🙂

Sorry, tried to word that as friendly as possible, it's not a dig at you. But if there are conditions outside of our control (i.e. the user not keeping other objects) outside of our control, then it is by definition not reliable.

>  wonder if you're focusing too much on my particular scenario of a third-party library wanting to access a PDO object? That's what motivated me to switch from the sqlsrv driver to the PDO driver, which is how I encountered this problem, but this problem exists within CodeIgniter itself. It would affect anyone who uses CI's PDO driver (without third-party libraries, and without directly accessing `$conn_id` or `$result_id`) and expects `close()` to actually close the database connection. *That*'s the use case that I'm trying to address.

As I mentioned, I didn't think your particular use case was that important; exactly because I know you're addressing a more general problem. I promise I'm not focusing on that at all, but it was worth mentioning given the connection as you're the first one to experience it.

> (I'll also note that I'm not the only one to encounter that sort of scenario - #5896 addresses the same issue for the Oracle driver.)

The OP in that issue is a regular contributor and having some observations I'm 99% sure that the only reason he opened that pull request is that he saw yours, as opposed to actually experiencing a problem.
Not to say that makes the whole thing less important, but have it in mind that he's also practically copying your solution without independently having identified it as the best possible one. Plus, the OCI driver is a different beast and it does close its connections; the cursors left behind and possibly not being closed fast enough by the database itself is a slightly different issue.

> - CI should manage and clean up *its own* object references. That's usually good software design in general, and for PDO in particular, it's necessary to comply with PDO's design, because PDO links cleaning up resources with closing the connection.
> - `$result_id` in particular is one of CI's own object references. CI creates it, users never have to interact with it, and it's not documented (besides a couple of lines of PHPDoc). So users shouldn't have to know about it, and CI should clean it up. In other words, it should be part of the abstraction that CI maintains.

I agree in general, but again, not sold on it happening in `close()` in particular. Especially if we can isolate `$result_id` to `CI_DB_result` instances, then that should solve the problem by itself.
Attempting to solve it in `close()` is also making a promise that we know we can't keep.

> - Similar to what you said, clearing `$conn_id` is somewhat equivalent to another driver's `_close()` function call, and so the PDO `close()` method should do it.

Already agreed on that.

> - But the PDO docs explain that clearing `$conn_id` is only *somewhat* equivalent to another driver's `_close()` function. It's necessary but not sufficient: the PDO docs explain that closing a connection involves clearing any references to the PDO instance itself and any references to any PDOStatement objects for that PDO instance. That's the documented design of PDO, and so it's reasonable to treat clearing PDOStatement references as part of closing the database connection.

Reasonable, but arguable.
And I'm always keen on ironing out every detail first. :)

> - It's not practical to guarantee that all PDO-related references are cleared. Maybe the user messed around with the public `$conn_id` or `$result_id` members; maybe the user is storing a CI_DB_result somewhere. But those are less common scenarios and difficult or impossible to solve. The CI PDO driver can and should clear PDOStatement instances that it creates and uses internally, because that fits PDO's design, preserves the abstraction (so users of CI don't have to worry about the CI PDO driver's implementation), and is easy to do (so no noticeable costs on code quality or maintainability).

Yes, it's not practical, but users expect it and so we try to avoid unsolvable problems.

> I initially understood you to be saying that you thought CI should shift the responsibility of cleaning up `$result_id` onto whoever's using CI, and so that's part of what I was trying to address in my earlier reply. After reading your latest reply, it sounds like you agree with CI cleaning it up but are unconvinced of the mechanism. If I've misunderstood, please let me know.

I wasn't necessarily thinking that it should, but it was a possible option. At this point, I am indeed rather only concerned with the mechanism.

> The precedent concern is, to some extent, a slippery slope argument: Just because there are good reasons for cleaning this particular object reference (because it's the only one that the CI PDO driver created and managed itself, because it's trivial to implement, because it's necessary for `close()` to behave correctly, because that's the behavior mandated by the PDO docs) doesn't mean that the rest of CI needs to do it.

True, but if the opportunity for someone to argue that arises, it will not only be used as an argument, but quite likely with a lot of zeal too. I'm yet to see a lengthy discussion that doesn't involve someone fighting a battle on the consistency hill.

> If I understand correctly, you're proposing addressing the issue by removing the `$result_id` member variable completely and changing `CI_DB_driver::query` to retrieve and cache the number of affected rows, so that `CI_DB_driver::affected_rows()` can return it later?

Not completely removing it, as this is a PDO problem and the PDO driver can easily override `query()` to apply this caching mechanic just for itself. But yes, that's the gist.

... which I feel also addresses the rest of your comment, that I'm too lazy to properly quote with formatting. However, it should be noted that `affected_rows()` is only used for "write" type queries such as `INSERT` and `UPDATE` statements and is not affected by buffering, while `num_rows()` is a `CI_DB_result` method.

joshkel:
> The OP in that issue is a regular contributor and having some observations I'm 99% sure that the only reason he opened that pull request is that he saw yours, as opposed to actually experiencing a problem.

I'm not sure why you're saying that - he explained that he was getting an ORA-01000 error from too many open cursors and that his PR fixed his problem.

So, back to the main discussion... Here's my understanding.

We agree that, one way or another, the CI PDO driver should try to make `close()` close the database connection and free database resources. (We're still discussing the best way to make this happen.)

We agree that it's impossible to guarantee that `close()` will close the database connection. Regardless of what we do, the user may directly access `$result_id` or `$conn_id`, or they may store a reference to a `CI_DB_result`. For PDO, doing any of that will keep the connection from being closed. As I understand it, this is your biggest objection to the approach in this PR.

Am I understanding correctly? If so, then that's _not an argument against this PR_ - because, regardless of what approach we take for this issue, the user could do something that keeps a reference to a PDO object and keeps the connection open.

(That's not _entirely_ true. I think the CI PDO driver could, in theory, guarantee closing - make `$conn_id` and any other PDO-related references protected or private so that users can't access them, and keep references or weak references to every `CI_DB_result` object it hands out, so that when `close()` is called, it tracks down every reference to every result it ever created and clears their PDO-related object references too. There would be lots of disadvantages to this approach in terms of backward compatibility and implementation complexity and memory requirements or dependencies. I don't think it's a good idea.)

So, as I see it, regardless of what we do, the user could do something to make a problem for themselves. We're just trying to find the cleanest approach to ensuring that CodeIgniter doesn't contribute to that problem by how it manages its own resources. There are two design approaches to handling the resource issue: the approach in this PR (where the current `$result_id` remains unchanged, and `close()` frees it), and the approach you suggested (where the PDO no longer keeps a `$result_id` and, for the one `affected_rows()` method where it used `$result_id`, it instead retrieve the number of rows within `query` and caches it). Thinking about how I'd evaluate the tradeoffs of the two alternatives if I were a maintainer of a framework:

The approach in this PR:

* It solves the current problem.
* It's not a guaranteed fix (if users directly access `$conn_id` or `$result_id` or keep references to `CI_DB_result`).
* It's a clean design: It's logical and consistent for a CI driver's `close()` to free database resources that are managed by the CI driver, and the PDO docs say that closing a database connection involves freeing references to PDO objects.
* It makes minimal changes to the code (so there's less risk of bugs).
* It raises the possibility that users will use it to argue that CI should take responsibility for managing more resources. (As I understand it, this is one of your big concerns, so if I'm misstating it, please let me know.) If so, there's a straightforward counterargument: `close()` should only attempt to take responsibility for resources that the CI driver is managing, and the only resources that the CI driver is managing are `$conn_id` and `$result_id`.

The approach of removing `$result_id` within the CI PDO driver and changing `affected_rows`:

* It solves the current problem.
* It's not a guaranteed fix (if users directly access `$conn_id` or keep references to `CI_DB_result`).
* It's a clean design within the CI PDO driver (although it may hurt the broader design - see below). In fact, it's probably a cleaner design within the CI PDO driver than the PR approach, because it means the driver is tracking less state and keeping fewer resources.
* It would take more time to implement and test.
* It still has the possibility that users will argue that CI should take responsibility for managing more resources (because the CI PDO driver still can't guarantee that `close()` works unless it does so). If so, there's an extremely straightforward counterargument: `close()` only tries to close the direct `$conn_id` reference.
* It means that the CI PDO driver would be implemented differently than the other CI drivers; it would no longer use `$result_id` and would eagerly retrieve and cache the affected row count, while other drivers would still use `$result_id` and would determine the affected row count on demand. If I were maintaining CI, I'd feel like that complicates the design, and I'd be concerned that this inconsistency would make the CI PDO driver harder to maintain.
* It's not straightforward to implement. You mentioned overriding `query()`, but `query()` is a fairly complex method (handling caching, creating result instances, error handling, logging, etc.); I suspect it's not really intended to be overridden. (This may be an example of the open-closed principle?) From looking at the source code, overriding `_execute` seems like the most appropriate place, although the [docs](https://codeigniter.com/user_guide/database/queries.html#simplified-queries) make it sound like `simple_query()` (which calls `_execute`) has no side effects, and its current implementation has no side effects, and storing the affected row count would be a side effect, so that doesn't seem ideal. Or we could create a new overridable function as a new extension point that's only implemented by the CI PDO driver, but again, that complicates the design.
* It affects backward compatibility: the public `$result_id` member would no longer behave as it used to. We've talked about how this is public for legacy reasons and is largely implementation-specific, but since it's public, I'd guess there's a decent chance that someone's using it. (Cf. [Hyrum's law](https://www.hyrumslaw.com/).)
* It likely affects the behavior of `affected_rows()`: based on the database documentation I mentioned, `affected_rows()` could start to return different results depending on what type of query is executed, whether the CI PDO driver or another CI driver is used, and whether queries are buffered or unbuffered. (I'll admit I haven't chased down all of the potential considerations here.) I know that you said that `affected_rows()` is only intended for write type queries, so I don't think its use as intended would be affected. However, I've seen plenty of code that's written in a way that happens to work even if it's not the way that the API / framework authors intend, so I'd be concerned about backward compatibility there, too. (Hyrum's law applies here, too.)
* The current `CI_DB_result` constructor takes as its only parameter the `CI_DB_driver` instance and assumes it can get the `$result_id` it needs from there; this would have to change for the PDO driver. It's unlikely that any users are relying on this, but it is part of the publicly available API, so it's a (very?) minor backward compatibility concern.

kenjis:
> > The OP in that issue is a regular contributor and having some observations I'm 99% sure that the only reason he opened that pull request is that he saw yours, as opposed to actually experiencing a problem.
> 
> I'm not sure why you're saying that - he explained that he was getting an ORA-01000 error from too many open cursors and that his PR fixed his problem.

@narfbg I'm  also not sure why you think so.
He got the error in his life and tried to fix the problem.
He actually counted the number of open cursors:
https://github.com/kenjis/ci-phpunit-test/pull/322#issuecomment-579762908

ytetsuro:
@narfbg

> The OP in that issue is a regular contributor and having some observations I'm 99% sure that the only reason he opened that pull request is that he saw yours, as opposed to actually experiencing a problem.

Your delusion is wrong.

I not read this PR, when I made PR.

I attach the screenshot I encountered error.

スクリーンショット 2020-02-01 20 07 17

narfbg:
> Your delusion is wrong.

My "delusion"?! I hope this is a language barrier problem.

OK, fine. Excuse me for being convinced that 2 PRs within 24 hours for a previously unseen problem aren't independent.
I should still note that this screenshot shows an ORA-00604 error and not ORA-01000.

> We agree that, one way or another, the CI PDO driver should try to make `close()` close the database connection and free database resources. (We're still discussing the best way to make this happen.)

We agree that it should close the connection.
We do not agree that it is responsible for resource management.
That's 2 different things that you see as part of a single process and I don't.

> We agree that it's impossible to guarantee that `close()` will close the database connection. Regardless of what we do, the user may directly access `$result_id` or `$conn_id`, or they may store a reference to a `CI_DB_result`. For PDO, doing any of that will keep the connection from being closed. As I understand it, this is your biggest objection to the approach in this PR.

To be more precise, `CI_DB_result` objects are my biggest concern as they are our way of exposing result sets to the user, but yes.

> Am I understanding correctly? If so, then that's *not an argument against this PR* - because, regardless of what approach we take for this issue, the user could do something that keeps a reference to a PDO object and keeps the connection open.

`close()` doing more than it is supposed to is.
`close()` attempting to do more than it is capable of is.

> The approach of removing `$result_id` within the CI PDO driver and changing `affected_rows`

It's not removing or even changing the visibility of `$result_id`, this is all that needs to happen in `CI_DB_pdo_driver`:

    protected $affected_rows = 0;

    public function query($sql, $binds = FALSE, $return_object = NULL)
    {
        $result = parent::query($sql, $binds, $return_object);
        $this->affected_rows = $this->is_write_type($sql)
            ? $this->affected_rows = $this->result_id->rowCount()
            : 0;
        $this->result_id = FALSE;
    }

    public function affected_rows()
    {
        return $this->affected_rows;
    }

This leaves us with `$result_id` immediately freed from `CI_DB_driver` and encapsulated entirely into `CI_DB_result`.

Yes, it's more code and slightly more complex - that's still one point in favor of your solution - but it's not nearly as complicated as you made it out to be.

joshkel:
> We agree that it should close the connection.
> We do not agree that it is responsible for resource management.
> That's 2 different things that you see as part of a single process and I don't.
> `close()` [is] doing more than it is supposed to.

A database connection is a resource. Closing connections is part of resource management. That's a general software development principle.

PDO in particular sees them as part of a single process. So, when writing code that uses PDO in particular, they have to be linked.

I really don't understand why you're trying to draw such a sharp distinction here. This seems to be the crux of our disagreement, so I'd really like to understand where you're coming from.

I thought, "An object generally needs a place where it frees its internal resources - e.g., closes any database connections - and, for CI drivers, that's the `close()` method." To me, that's a very clean and straightforward and uncontroversial approach. Your approach seems to instead be, "The CI driver's `close()` method is responsible for closing any database connections, but it would be wrong to try to handle any other internal resources there." Am I understanding you correctly?

In principle (not specifically for this PR), if `close()` should not be responsible for managing a CI driver's internal resources, then where should they be managed? Your preferred approach seems to be, "Try as hard as possible to make sure that there _are_ no other internal resources", but what if there are circumstances (perhaps #5896) where that isn't possible?

And what makes the "Try as hard as possible to make sure there are no other internal resources" approach better, given the costs in design effort  and implementation complexity and backward compatibility?

Again, I really would like to understand where you're coming from.

> `close()` attempting to do more than it is capable of is.

But, regardless of what approach is taken here, `close()` is attempting to do more than it's capable of. `close()` is trying to close the database connection, and we can't guarantee that it does so. `CI_DB_result` keeps references to `$conn_id` as well as `$result_id`, so `close()` is as incapable of actually closing `$conn_id` as it is `$result_id`. That's the point I was trying to make when I said that this isn't an argument against this PR.

> It's not removing or even changing the visibility of $result_id, this is all that needs to happen in CI_DB_pdo_driver:

I thought I'd considered this approach of overriding `query` and thought it wouldn't work. I must have been wrong. Thank you.

But it does effectively remove `$result_id`. A property with that name still exists, but it's never used (always false) outside of `query`, so it might as well have been removed.

(I'll also admit I'm not a huge fan of the approach. It means that the driver's correct behavior now depends on fairly specific interactions and side effects between `CI_DB_driver::query`, `CI_DB_pdo_driver::query`, `CI_DB_result`, and `$result_id`. That seems to hurt maintainability - the new flow is harder to reason about, and the tighter coupling can make future changes harder - even though (as you said) the actual code is only slightly more complex. I'm not too interested in debating this, though - this discussion's long enough already.  🙂)

And the other concerns I'd listed still apply: an internally inconsistent design, `query()` being a complicated function that doesn't lend itself to overriding (open-closed principle), backward-incompatible changes to the observable behavior of `$result_id` and `affected_rows()`.

With all of that said - this approach works. (Or, at least, it will work, once it properly handles the error case.) You're the maintainer, so it's ultimately your call; you don't have to convince me. 🙂(Unfortunately, I've already spent a lot more time on this issue than I could have reasonably expected, so I'm not sure how much more I'll be able to engage.)

Thank you.

narfbg:
> A database connection is a resource. Closing connections is part of resource management. That's a general software development principle.

Oh, come on. You know what I meant.

> PDO in particular sees them as part of a single process. So, when writing code that uses PDO in particular, they have to be linked.

Sorry to have to put it so bluntly, but PDO sees them as crap that it refuses to manage itself. Anything else is speculation on our part.

> I really don't understand why you're trying to draw such a sharp distinction here. This seems to be the crux of our disagreement, so I'd really like to understand where you're coming from.

Well, I'm doing my best. :)

> My approach is, "An object generally needs a place where it frees its internal resources - e.g., closes any database connections - and, for CI drivers, that's the `close()` method." To me, that's a very clean and straightforward and uncontroversial approach. Your approach seems to instead be, "The CI driver's `close()` method is responsible for closing any database connections, but it would be wrong to try to handle any other internal resources there." Why?

Because, as I've said earlier, `close()` has never had that job. That place in PHP is `__destruct()` and as a bonus PHP itself usually does it by itself. Just so we're clear, I'm not suggesting to add a destructor now - I don't see it working - only providing some context.

> In principle (not specifically for this PR), if `close()` should not be responsible for managing a CI driver's internal resources, then where should they be managed? 

As weird as this may sound - nowhere. In an ideal world anyway.

Keep in mind that we're talking only about a single property that only exists because of legacy design decisions prior to visibility declarations being a thing in PHP. `CI_DB_driver::$result_id` wouldn't have existed if the entire thing was written today.
There are no other resources to manage; that's just not a thing for us, has never been and I'd like to keep it that way.

It used to be the case that a `$RES->result_id = $this->result_id` assignment was happening inside `query()`, and probably other similar operations elsewhere ... don't remember anymore, but the point is code relied on `$result_id` being public so it can be seen between multiple classes, which is no longer the case.
It could probably be just a local variable at this point, and that should practically eliminate the problem from our POV as local vars are naturally erased by PHP as soon as possible. Users don't have access to mess with them either.

> Your preferred approach seems to be, "Try as hard as possible to make sure that there are no other internal resources", but what if there are circumstances (perhaps #5896) where that isn't possible?

Tell you what - we've both invested much more time in this already than it's really worth, and it might indeed be the case that #5896 isn't resolvable by different means, so let's review that and if that's the case - I'm merging this as is.
The oci8 driver has it's own legacy baggage, including methods and properties that are exclusive to it, so I'm not keen on it being a deciding factor, but let's say at least that would be consistent.

> And what makes the "Try as hard as possible to make sure there are no other internal resources" approach better, given the costs in design effort and implementation complexity and backward compatibility?

Design effort is hardly a problem in my view, as good design pays for itself in advance. And I don't really see BC issues here - sure, you referenced Hyrum's law, but in my book it basically amounts to "people are idiots"; we can't reasonally be expected to concern ourselves with anything but APIs documented and designed for public use.
You have a much more academic approach to this than me, so don't expect me to link citations, but I believe you would agree that the shorter the lifespan of a variable/process/resource/whatever - the better. Therefore, why wouldn't it be better to get rid of `$result_id` as early as possible as opposed to keeping it around for no obvious reason and then having to figure out where do we free up resources?

>> close() attempting to do more than it is capable of is.
> But, regardless of what approach is taken here, `close()` is attempting to do more than it's capable of. `close()` is trying to close the database connection, and we can't guarantee that it does so. `CI_DB_result` keeps references to `$conn_id` as well as `$result_id`, so `close()` is as incapable of actually closing `$conn_id` as it is freeing `$result_id`. That's the point I was trying to make when I said that this isn't an argument against this PR.

Actually ...

    $ grep -ri 'conn_id' system/database/drivers/ | grep -i 'result\.php'
    $

Apparently `CI_DB_result::$conn_id` can be removed entirely without any side effects, so there's that. Glad you brought it up, hadn't noticed it myself. :)

Yes, with absolutely every detail considered `close()` will never be able to guarantee to have closed a PDO connection. The closest thing it can guarantee is that no `PDO` instance exists anymore outside of someone intentionally copying the internal property that is `$conn_id`. In fact, we probably could make `$conn_id` non-public to guarantee non-interference.
However, the mere existence of `CI_DB_result` (and of course, that it keeps a `$result_id` reference itself) means that `close()` attempting to deal with `$result_id` is by definition trying to do something outside of its control.
If we could at least *detect* that other `$result_id` references still exist, then we could throw an exception and that'd be an acceptable way of saying "Sorry, something out of our control is happening", but I don't see how we'd be able to do that in a reasonable way.

> But it does effectively remove `$result_id`. A property with that name still exists, but it's never used (always false) outside of `query`, so it might as well have been removed.

True, but my point was that it's not such a huge change and it doesn't have to affect other drivers - it's limited to a very small amount of code isolated to the PDO driver. PDO driver creates the problem - PDO driver solves the problem.
We probably can remove it entirely if we wanted to - haven't really looked into it, it might actually simplify things, who knows - but we don't necessarily need to.

> And the other concerns I'd listed still apply: an internally inconsistent design, `query()` being a complicated function that doesn't lend itself to overriding (open-closed principle), backward-incompatible changes to the observable behavior of `$result_id` and `affected_rows()`.

I honestly don't see how these things are of any concern.

> With all of that said - this approach works. (Or, at least, it will work, once it properly handles the error case.) You're the maintainer, so it's ultimately your call; you don't have to convince me. :slightly_smiling_face: (Unfortunately, I've already spent a lot more time on this issue than I could have reasonably expected, so I'm not sure how much more I'll be able to engage.)

What can I say, I'm a stickler for minor details. :)
I have to admit it's refreshing to have someone else just as motivated on the other end, without it getting ugly anyway - thanks for that, I do appreciate your patience - but I can assure you I take no pleasure in dragging this for such a long time.

joshkel:
> Oh, come on. You know what I meant.

I honestly wasn't trying to be difficult - just trying to understand where you're coming from.

> Because, as I've said earlier, close() has never had that job. That place in PHP is __destruct() and as a bonus PHP itself usually does it by itself. Just so we're clear, I'm not suggesting to add a destructor now - I don't see it working - only providing some context.

For what it's worth, it was my understanding that code often needs a place besides the destructor to clean up non-memory resources (since the destructor runs while freeing memory / destroying objects / running the garbage collector, but that may not happen until a while). That's why Java has try-with-resources, C# has `IDisposable` / `using`, Python has `with` and context managers (even though CPython, like PHP, uses reference counting), etc.

I'll admit that I'm coming at that mostly from how other languages handle it, though; I haven't been doing serious PHP work for as long.

Regarding the rest of your reply, I'd definitely agree on shorter lifespan being better (and no need to link citations). And I agree on how `$result_id` and `$conn_id` would ideally be handled. I'm not sure I agree on the best way to get there, but, again, that's your call, so I'm happy to go along with whatever you decide.

> Tell you what - we've both invested much more time in this already than it's really worth, and it might indeed be the case that #5896 isn't resolvable by different means, so let's review that and if that's the case - I'm merging this as is.

Sounds good. Just let me know if and how you'd like for me to proceed.

> I have to admit it's refreshing to have someone else just as motivated on the other end, without it getting ugly anyway - thanks for that, I do appreciate your patience - but I can assure you I take no pleasure in dragging this for such a long time.

Thank you. And thanks again for your time and patience.

narfbg:
OK, well ... I guess it's up to you whether you want to subscribe as a watcher and/or participate in the #5896 discussion. If you're not involved, I'll just get back here with the results after it's done.

joshkel:
Thanks. I'd already subscribed to #5896, so I'll keep an eye there.

narfbg:
Further info in #5896.

PR #5896, opened 2020-01-29
fix too many connection for oracle

ytetsuro:
I ran the following error when I ran about 3000 unit tests.

ORA-01000

The cause is that the cursor became huge because `oci_free_statement` was not performed.

This code reduced the oracle open cursor by a factor of ten.

I can't submit the actual test code, but I don't think this implementation will cause a bug.

I measured the number of cursors with the following SQL.

```
select   sid,   statistic#,   value from v$sesstat where statistic# = 5   and sid in (select sid from v$session where username ='test');
```

### DIFF
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -682,6 +682,14 @@ class CI_DB_oci8_driver extends CI_DB {
 	 */
 	protected function _close()
 	{
+		if (is_resource($this->curs_id))
+		{
+			oci_free_statement($this->curs_id);
+		}
+		if (is_resource($this->stmt_id))
+		{
+			oci_free_statement($this->stmt_id);
+		}
 		oci_close($this->conn_id);
 	}
 
--- a/system/database/drivers/oci8/oci8_driver.php
+++ b/system/database/drivers/oci8/oci8_driver.php
@@ -686,10 +686,12 @@ class CI_DB_oci8_driver extends CI_DB {
 		{
 			oci_free_statement($this->curs_id);
 		}
+
 		if (is_resource($this->stmt_id))
 		{
 			oci_free_statement($this->stmt_id);
 		}
+
 		oci_close($this->conn_id);
 	}
 


ytetsuro:
The following is almost the same problem.

#5896

ytetsuro:
@narfbg 

This is a comment to another PR, but I will reply here because it is relevant here.

> My "delusion"?! I hope this is a language barrier problem.

I'm sorry.
I am very poor at English.
It may have inadvertently hurt you.
I'm very very sorry.

> I should still note that this screenshot shows an ORA-00604 error and not ORA-01000.

I think `ORA-00604` in my attached screenshot is caused by` ORA-01000`.
See inside the red frame.

スクリーンショット 2020-02-01 20 07 17

And this error was resolved in this PR.

narfbg:
OK, fair enough.

So ... following the long discussion in #5895, we decided to first see how this one should ideally be handled.
Ideal for me would be a way to eliminate `curs_id` and `stmt_id` altogether, and after some digging I found out that:

- `CI_DB_oci8_driver` only needs `stmt_id` (practically the equivalent of `result_id` in other drivers, dunno why it is named differently) for `affected_rows()` and possible error reporting. We can possibly cache the results for both.
- The oci8 driver has a custom `stored_procedure()` method that kind of relies on `stmt_id`, but:
  - Well, it's a custom method; it's not compatible with other drivers and I've long been looking for an excuse to either rewrite (to make compatibility possible) or remove it altogether.
  - Can likely be altered so it can open and close `stmt_id` by itself on the fly.
- `curs_id` is only created (by `get_cursor()`, another custom method) and never actually used by the driver, at all. `CI_DB_oci8_result` will try to use it *instead* of `stmt_id`, if it was previously created, but it is unclear if and how anybody uses that feature at all.
- `CI_DB_oci8_result` [sets `CI_DB_oci8_driver::$stmt_id` to false in its constructor](https://github.com/bcit-ci/CodeIgniter/blob/432c22ba111a4b3c5c2f96edbb47c22b0e7dbf36/system/database/drivers/oci8/oci8_result.php#L95), in practice taking it away from the `*driver` class control for read-type queries.

So ... We're again left with a situation where the patch attempts to free resources in `close()` while that may not be possible at all, yet you've measured that it actually does release a good amount of them.
Can you provide some insight into what your tests generally do (seeing as that's where you saw the problem).
Does the percentage of now released cursors correlate with the amount of write-type queries that you have? Do you maybe use `stored_procedure()` or `get_cursor()`?

ytetsuro:
> Can you provide some insight into what your tests generally do (seeing as that's where you saw the problem).
> Does the percentage of now released cursors correlate with the amount of write-type queries that you have? 
> Do you maybe use stored_procedure() or get_cursor()?

Investigate other questions, but provide the range you know.

> Can you provide some insight into what your tests generally do

I mostly use `query builder` except for complicated merge statements and select.
But I don't know if it's generally.
At least the tests that have failed are write-type queries using the `query builder`.

> (seeing as that's where you saw the problem)
> Does the percentage of now released cursors correlate with the amount of write-type queries that you have? 

Prepare a log in tearDown, measure it, and respond.

> Do you maybe use stored_procedure() or get_cursor()?

Not using.

ytetsuro:
@narfbg

> Ideal for me would be a way to eliminate curs_id and stmt_id altogether,

What about the following solution?
1. Create a class that holds `stmt_id` and `curs_id` (tentatively `oci8_statement`).
1. Instantiate `oci8_statement` when generating stmt_id or curs_id.
1. Refer to the latest `oci8_statement` with the __get magic method of oci8_driver.
1. Implement so that the destructor of `oci8_statement` releases stmt_id and curs_id.

narfbg:
I mean ... that's an option, but why go there if we can eliminate them instead?

> At least the tests that have failed are write-type queries using the `query builder`.

That's an indicator that we can release `$stmt_id` as soon as write-type queries are executed.

ytetsuro:
> I mean ... that's an option, but why go there if we can eliminate them instead?

`there` means my solution?

If `there` means my solution,
This is because `stmt_id` and` curs_id` can be excluded from `oci8_driver`.

ytetsuro:
@narfbg 

> Does the percentage of now released cursors correlate with the amount of write-type queries that you have?

For the percentage of cursors, there were many write-types.
However, there were a certain number of read-types.

My verification method is as follows.
The following shell script is executed for tearDownAfterClass.
Attach sum.log generated from shell script.

logger.sh
```shell
#!/bin/bash

sql+ -s / as sysdba @/vagrant/select_open_cursor.sql >> /vagrant/total_connection_count.log
sql+ -s / as sysdba @/vagrant/select_open_cursor.sql > /vagrant/connection_count.log
_read_sum=`cut -f 3 -d "," /vagrant/connection_count.log | grep -i -E '^SELECT' | wc -l`
_write_sum=`cut -f 3 -d "," /vagrant/connection_count.log | grep -i -E '^(INSERT|UPDATE|MERGE|DELETE|TRUNCATE)' | wc -l`
echo "$_read_sum,$_write_sum" >> /vagrant/sum.log
```

select_open_cursor.sql
```sql
set head off
set pagesize 0
set colsep ','
set linesize 9999

SELECT SID, USER_NAME, SQL_TEXT FROM V$OPEN_CURSOR WHERE USER_NAME <> 'SYS';
```

[sum.log](https://github.com/bcit-ci/CodeIgniter/files/4190240/sum.log)

Left column is read-type query.
Right column is write-type query.

narfbg:
Sure, it would be helpful to be able to compare the two logs.

ytetsuro:
> Sure, it would be helpful to be able to compare the two logs.

Attach the log with PR applied.

[pr_apply_sum.log](https://github.com/bcit-ci/CodeIgniter/files/4195962/pr_apply_sum.log)

narfbg:
Right, as I suspected it closes a majority of cursors for write-type queries. If you used `free_result()` on `CI_DB_oci8_result` objects after you're done with them, I'm sure the read-type ones will be dramatically reduced too.

I have a couple of ideas for how to automate both "branches", but quite a bit of tinkering will be required and it's probably best that I do that later myself, while just merging this and #5895 as a temporary measure for now.
Will get back to you after I give it some more thought.

ytetsuro:
OK.
Thanks response.

narfbg:
OK, I'll merge this and #5895 for now and I'll work on a better patch myself later.
But you'll need to make it compliant with the [styleguide](https://codeigniter.com/user_guide/general/styleguide.html) - after closing a brace, you must follow it with a single empty line before writing a new statement.

ytetsuro:
> But you'll need to make it compliant with the styleguide - after closing a brace, you must follow it with a single empty line before writing a new statement.

I'm sorry for the slow my response.

fixed.

kenjis:
@narfbg Could you merge this when you have time?

narfbg:
So, it's been quite awhile and I had to basically reinvent my ideas on this, so no guarantee that it's better than what I had originally thought, but still - as long as destructors work correctly, the last commit I linked here should drastically reduce garbage statement IDs, if not eliminate them completely. :)

If you can, please drop me a line when you have a chance to test this for yourselves.

ytetsuro:
@narfbg 

Thanks merged.
Welcome home.
I'm glad to see you back in the CodeIgniter repository.

I've see your commit.
I will test it and if there are any problems, I will create a separate PR.
Thank you for your correction.

PR #5903, opened 2020-02-10
Add server protocol value variation

achraf52:
In some Apache servers running the `HTTP/2` protocol, calling `$_SERVER['SERVER_PROTOCOL']` may return `HTTP/2.0` instead, thus causing CodeIgniter to (incorrectly) determine the output header protocol to use. This commit adds the different variation value to fix that.

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -569,7 +569,7 @@ if ( ! function_exists('set_status_header'))
 			return;
 		}
 
-		$server_protocol = (isset($_SERVER['SERVER_PROTOCOL']) && in_array($_SERVER['SERVER_PROTOCOL'], array('HTTP/1.0', 'HTTP/1.1', 'HTTP/2'), TRUE))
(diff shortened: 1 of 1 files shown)


narfbg:
Sounds reasonable, but it might be a violation of the protocol ... Can you provide references to such references?

achraf52:
It seems like `HTTP/2.0` the original name before it was renamed into `HTTP/2` for simplicity's purposes.

In the [RFC 7540](https://tools.ietf.org/html/rfc7540) standard there is a single mention of `HTTP/2.0` as part of the Connection Preface string, which seems to have nothing to do with the reason Apache would return `HTTP/2.0` for `$_SERVER['SERVER_PROTOCOL']` but it's there.

I should note that I'm experiencing the issue in Windows 10 using XAMPP v3.2.4 under Apache v2.4.41 so it might be just a bug in this particular Apache build. However, I suggest that CodeIgniter accounts for this to prevent unintentionally downgrading the connection to `HTTP/1.1` while `HTTP/2` is available and being used.

PR #5907, opened 2020-02-15
Fixed double spacing in mimes.php

pkarunyu:
Found two errant double spaces in mimes.php, fixed herein.


pkarunyu:
The error reported by travis-ci seems to be coming from the job with PHP 5.4 and using pgsql DB. The actual error is coming from composer:

```
sh -c "composer install --dev --no-progress"
You are using the deprecated option "dev". Dev packages are installed by default now.
Loading composer repositories with package information

[Composer\Downloader\TransportException] 
Peer fingerprint did not match 
```

PR #5922, opened 2020-03-03
Add UptimeRobot user agent

robsonpiere:
Add UptimeRobot user agent, a free website monitor.
 
https://uptimerobot.com/


narfbg:
Why wouldn't both be reported as e.g. "UptimeRobot"?

robsonpiere:
it is because they are different companies, uptimebot is a paid service, while uptimerobot has a free plan.

https://uptimerobot.com/

https://uptime.com/pricing

PR #5938, opened 2020-04-02
Fixed issue with creating a resource from png using Image_lib

taitomagatsu:
Due to this pr https://github.com/bcit-ci/CodeIgniter/pull/5883 while changing the function `image_create_gd` to add support for webp, the option to create a resource from png is no longer possible.


narfbg:
Good catch :+1:

PR #5940, opened 2020-04-03
Add optional footer to table with custom content and aggregation functions support ( Sum, Average,Min,Max,Count )

birddevelper:
**set_footer_columns** function is added to class. Now you can have footer at the bottom of the table where it can show Sum, Average,Min,Max, Count of column or even a custom text. Moreover it can be stylized by footer templates.

Example : 
Here we have table with WLID , DEBIT, CREDIT , DATETIME, DESCRIPTION
`$query = $this->db->query("SELECT  WLID, DEBIT, CREDIT, DATETIME ,DESCRIPTION FROM wallet  ");`

We want to have sum of CREDIT and DEBIT columns, count of non empty cells in DESCRIPTION column, and custom text "Total" in WLID column : 

`$this->table->set_footer_columns(["CREDIT"=>"SUM","DEBIT"=>"SUM","WLID"=>"Total","DESCRIPTION"=>"CNT"]);` 

Result :

ci

### DIFF
--- a/system/libraries/Table.php
+++ b/system/libraries/Table.php
@@ -175,17 +175,16 @@ class CI_Table {
 	
 	
 	/**
-	 * Set the table footer sum columns
+	 * Set the table footer columns with their function or custom content 
 	 *
-	 * Can be passed as an array or discreet params
+	 * Must be passed as an array
 	 *
 	 * @param	mixed
 	 * @return	CI_Table
 	 */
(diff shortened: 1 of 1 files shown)


narfbg:
Sorry, but no.

I've many reasons to not want this, but primarily I'd rather just remove this library entirely than add more features to it.

PR #5946, opened 2020-04-13
Pass the travis php7.4 test

sapics:
By coping the file ('application/config/foreign_chars.php') directly, travis test in php7.4 are passed.
I guess that there are some encoding errors in `ci_vfs_clone` for cloning the UTF-8 file in php7.4.

The error was just a coping the file for creating test-environment, thus,
we could consider that CodeIgniter itself could pass php7.4 travis test, I guess.

EDIT: I add composer cache in Travis, because composer downloading causes error sometimes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ services:
   - postgresql
 
 before_script:
-  - sh -c "composer install --dev --no-progress"
+  - sh -c "composer install --no-progress"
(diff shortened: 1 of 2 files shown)


narfbg:
Sorry for the late reply, this is good work that you've done ... Do you want this or #5945 though? One would have to be closed.

sapics:
Thanks for response!
I would like to merge this PR, thus I close the other one.

PR #5947, opened 2020-04-13
Add the ability to get the first Input item value from many

achraf52:
This pull request adds a feature to the Input core library that allows the developer to use a comma-separated string of items to any of the `get(), post(), cookie(), server(), input_stream()` methods and get back the value of the first item that's not NULL if one is found, otherwise NULL is returned as expected. This is an example of the expected usage:

`$requested_items = $this->input->get->('item_id, item_selection');`

This is useful for example when you're designing an API and you want to allow the user to either supply a String called `item_id `or an Array called `items_selection` for the requested data, but want to avoid having to manually check which one (of many) has been used by the user.

I hope this example showcases how this feature could be useful.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -138,7 +138,7 @@ class CI_Input {
 			return $output;
 		}
 		
-		if(is_string($index) === TRUE && strpos($index, ',') !== FALSE)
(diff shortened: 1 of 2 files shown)


narfbg:
Sorry, but no ... I'll always favor explicitness.

Issue #5949, opened 2020-04-18
Wrong vendor/autoload.php being loaded by default.


technoknol:
### Issue
Enabling `$config['composer_autoload'] = TRUE;` will load file from `application/vendor/autoload.php`.

But default CodeIgniter has composer.json file on project root. So by default `vendor` folder will be generated on root.

### Possible Fix
To fix this I need to set this variable like below, which is not good for all users. 

```
$config['composer_autoload'] = FCPATH. 'vendor'. DIRECTORY_SEPARATOR . 'autoload.php';
```

### Correct way
CodeIgniter should load file by default from `FCPATH/vendor/autoload.php`. This can be changed in `system/core/CodeIgniter.php` file.

gxgpet:
The `composer.json` from the CodeIgniter's root it's for the framework's own purpose (as listing on Packagist), and shouldn't be really changed by your project.

 Also, keep in mind that the same `system` directory can be used for multiple `application` directories. And changing it the way it works now it would trigger BC breaks (if you're saying about replacing the current default path with the new one).

technoknol:
Why changing `composer.json` for my own project is not recommended? Once I've required my project composer.json file is not longer CodeIgniter's own file. I can't see any problem doing so. Like other frameworks like (Laravel, laminas-mvc-skeleton) there are composer.json and that's for their packagist purpose too. But still people `require` it and use it the way they want. 

Regarding having multiple `application` directories, I believe majority of people are be using single directory, so setting that flag `TRUE` makes very easy for them. Also who are using multiple directories are considered advanced users and they can find their way around.

gxgpet:
I'm not sure how Laravel works, it may create a `composer.json` for you via some of its tools, yes, but for sure you are not changing the framework's own `composer.json`, which stays untouched in the package source directory.

> Once I've required my project composer.json file is not longer CodeIgniter's own file.

`my project composer.json` != `CodeIgniter's own file`. It's either yours or the project's one. If you change it, it's not the CodeIgniter's anymore.

> Also who are using multiple directories are considered advanced users and they can find their way around.

You can't simply throw that in people's faces... Their project is working because it's made based on the CI's documentation. There is nothing broken and thus, they shouldn't change the way it already works. Here you are taking into consideration a personal preference, which, of course, you can do it already by setting the `$config['composer_autoload']` to a path, and not to `TRUE`.

[...]

PR #5950, opened 2020-04-18
#5949 - By default load composer autoloader from vendor/autoload.php.

technoknol:
Load vendor/autoload.php file from `FCPATH` instead of `APPPATH`. This will load correct file when `$config['composer_autoload']` is set to `TRUE`.


narfbg:
Don't create issues (#5949) just to link them from a pull request ... Either create an issue and wait for a conclusion to it, or go directly with a PR - it can act as an issue thread too.

PR #5955, opened 2020-04-29
Fixed function "list_fields" using oci8 database.

thadus:
Without this fix, this function ignores database configuration "dbprefix" in the query.


narfbg:
This can have side effects - `protect_identifiers()` will add quotes and the output is used in a string comparison.

Issue #5791, opened 2019-06-30
[Suggestion] same-site attribute for cookie


yehgdotnet:
Let me suggest to support same-site attribute through $config['cookie_samesite'] 

Ref: https://www.owasp.org/index.php/SameSite

narfbg:
That's cool, but there's some issues around it ... For one, right now we just proxy to `setcookie()` and SameSite isn't supported on it until PHP 7.3.0; it's even trickier with sessions.

Do you have ideas about how it should ideally work?

yehgdotnet:
Hi @narfbg Andrey, If I understand your question, Samesite is a very cheap way of preventing CSRF attack.  But it will break if a web site has multiple subdomain to communicate for its authenticated content.

narfbg:
No ... I'm very well aware of what the SameSite attribute is and what it does.

What I'm asking is how would you implement it in CI?

campio97:
It would be a very useful thing. I have applications that make **CORS** calls to each other.
Google in April 2020 will need SameSite cookies = none. https://www.chromestatus.com/feature/5633521622188032
![image](https://user-images.githubusercontent.com/49672939/67033690-930a2180-f116-11e9-8083-0e7fb21d31cb.png)
`A cookie associated with a cross-site resource at <URL> was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at <URL> and <URL>`

mmihalev:
I see that session is already doing some php version checks like `is_php(XXX)`, so why not using the same technique to identify the php version which supports this natively (e.g. 7.3) and implement it into the `setcookie()` call.

narfbg:
I guess that makes sense for sessions, though I'm not sure if I'd want that to be configurable at all there.

Tell you what - it's up for grabs and anyone who wants this should submit a pull request with an implementation.

campio97:
[here my pull request.](https://github.com/bcit-ci/CodeIgniter/pull/5874)

narfbg:
Referencing it properly, so the threads get linked: #5874

harmandeep-singh:
if you have same issue like am having in shopify app written with CI 
upgrade the php version 7.3.x  and in your config file add these two lines will fix the issue in your config.php : 

```
<?php
/*config.php*/
[...]

PR #5961, opened 2020-05-29
Additional config for session cookie samesite attribute 

zawaruddin:
I write some additional config and edit core system to handle samesite attribute that will be rejected in most of browser nowadays.

`_The "csrf_cookie_name" cookie will be rejected immediately because it has the "sameSite" attribute that has a value of "none" or an invalid value, without the "secure" attribute. To find out more about the "sameSite" attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies_`

Based on issues [5791](https://github.com/bcit-ci/CodeIgniter/issues/5791) and [5874](https://github.com/bcit-ci/CodeIgniter/pull/5874) 

Using function `setcookie `  with `array option` to add '_samesite_' attribute that didn't handle previously. The value of _samesite_ depends on _cookie_samesite_ on configuration file.

### DIFF
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -302,7 +302,7 @@ class CI_Input {
 	 * @param	bool		$httponly	Whether to only makes the cookie accessible via HTTP (no javascript)
 	 * @return	void
 	 */
-	public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path = '/', $prefix = '', $secure = NULL, $httponly = NULL)
+	public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path = '/', $prefix = '', $secure = NULL, $httponly = NULL, $samesite = NULL)
 	{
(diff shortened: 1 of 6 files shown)


system/core/Input.php
```diff
@@ -347,8 +352,20 @@ public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path
 		{
 			$expire = ($expire > 0) ? time() + $expire : 0;
 		}
-
-		setcookie($prefix.$name, $value, $expire, $path, $domain, $secure, $httponly);
+		
+		// using setcookie with array option to add cookie 'samesite' attribute
+		setcookie(
```
sapics, line 357: I think that this looks work only for php7.3+.

narfbg:
1a2651040ef701e750b1c13cd69cc70814b079d0

PR #5969, opened 2020-06-23
Pass the travis php7.4 test for 3.1-stable

sapics:
This is for 3.1-stable branch version of #5946.


narfbg:
Sorry, I already handled this here 20ddc4692ed4d91cfbeae3354c0e5ef518b882a3

It is generally not a good idea to submit the same changes to two branches ... merges are the way to handle that.

Issue #5982, opened 2020-08-30
Redis session driver does not support TLS connections


Ddog800:
TLS connections for Redis are often required for compliance reasons (as I'm currently dealing with myself). While the underlying phpredis extension does support TLS connections by including `tls://` in the host string, it appears that the CI Redis session driver does not properly parse TLS connection strings when input into the application session config, causing the app to bomb.

After a bit of testing, this appears to be resolvable with a slight tweak to the Regex matching on the value of `$this->_config['save_path']` in the Redis session driver library constructor to ensure that `tls://` is matched and included in the first capture group. That will then allow a full host string of, i.e., `tls://myrediscache.amazonaws.com` to be passed along to the underlying phpredis functions.

Ddog800:
Opened PR #5983 to implement this change.

PR #5983, opened 2020-08-30
Add support for TLS connection strings

Ddog800:
Signed-off-by: Daniel Guidry <ddog800@gmail.com>

This will additionally match `tls://` and include it in the first capture group (containing the host string), if found. 

I have tested this against live Redis servers using both `tcp://` and `tls://` connection strings and have tested the pattern itself against various strings.

Issue #5982 has details.


narfbg:
Sorry for the long delay and thank you for your contribution. It's been quite awhile and I wanted to alter your patch a bit, so I just went ahead with it myself - see the linked commit.

On another note, if you want to contribute code again, you don't need to create an issue just to link a PR to it - a pull request can serve as an issue by itself.

PR #5991, opened 2020-09-15
Updated pdo_mysql_driver.php

hhuseyincihangir:
Table missing table metadata added.
- I needed the null and extra data from the table's metadata, but these were not added internally.


hhuseyincihangir:
Understood. Not added for compatibility with php 5.

PR #5995, opened 2020-10-01
Fixed issue with Redis Expire setting

mkilinskidev:
Signed-off-by: Mateusz Kiliński <mateusz.kilinski@gmail.com>

I've added some code to set the key's expire value.


narfbg:
What is the issue exactly? Why is this change necessary?

mkilinskidev:
Because the value has been returned even it was expired.

narfbg:
Sorry, but that didn't clear up anything to me. Can you elaborate? Maybe provide an example?

PR #5999, opened 2020-10-26
Custom captcha img tag  class for responsive captcha helper.

hhuseyincihangir:
Updated captcha_helper for customizable image tag class.

- It is useful to give class in order for the captcha visual to be responsive (we can send img-fluid class with bootstrap).

### DIFF
--- a/system/helpers/captcha_helper.php
+++ b/system/helpers/captcha_helper.php
@@ -57,11 +57,11 @@ if ( ! function_exists('create_captcha'))
 	 * @param	array	$data		Data for the CAPTCHA
 	 * @param	string	$img_path	Path to create the image in (deprecated)
 	 * @param	string	$img_url	URL to the CAPTCHA image folder (deprecated)
-	 * @param	string	$img_tag_class	Image tag for class (deprecated)
 	 * @param	string	$font_path	Server path to font (deprecated)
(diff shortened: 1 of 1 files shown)


system/helpers/captcha_helper.php
```diff
@@ -57,10 +57,11 @@
 	 * @param	array	$data		Data for the CAPTCHA
 	 * @param	string	$img_path	Path to create the image in (deprecated)
 	 * @param	string	$img_url	URL to the CAPTCHA image folder (deprecated)
+	 * @param	string	$img_tag_class	Image tag for class (deprecated)
 	 * @param	string	$font_path	Server path to font (deprecated)
 	 * @return	string
 	 */
-	function create_captcha($data = '', $img_path = '', $img_url = '', $font_path = '')
+	function create_captcha($data = '', $img_path = '', $img_url = '', $img_tag_class = '', $font_path = '')
```
orionstar, line 64: I think you should change the order of $img_tag_class and $font_path paramter to maintain compatibility with the previous signature of the function.

orionstar (changes requested):

orionstar (approved):

lazarospsa (commented):
👍🏼

PR #429, opened 2011-09-12
result_array(), result_object() to accept a field name to use as key of returned array

wyred:
http://codeigniter.com/forums/viewthread/193209/

http://codeigniter.uservoice.com/forums/40508-codeigniter-reactor/suggestions/1561893-pass-column-name-into-active-record-result-to-be

Example use:
In my users database table, I have:

<pre>
id  - name - email
1 - Preston - preston@libria.com
3 - Seamus - seamus@libria.com
4 - Dupont - dupont@libria.com
</pre>

In my app, I do:

<pre>
$rs = $this->db->select('id, name')->get('users');
$users = $rs->result_array('id');
</pre>

$users will be:

<pre>
Array
(
    [1] => Array
        (
            [id] => 1
            [name] => Preston
        )

    [3] => Array
        (
            [id] => 3
            [name] => Seamus
        )

    [4] => Array
        (
            [id] => 4
            [name] => Dupont
        )
)
</pre>

PR #6002, opened 2020-12-03
Changes in CI_DB_query_builder and CI_DB_result

lubo1998:
Hi, I have done some changes, which I hope will be useful for the project.

- Added custom key option in result_array - I know that I have duplicated the loops, but that is from optimization purposes. Better first do the check and then iterate, than do the check on each iteration.
- Added possibility to use where and or_where with a single field name and array of values which will produce "$field = $value" or "$field IN ($value1, $value2 [, ...$value2])".

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -607,13 +607,22 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 	 * Separates multiple calls with 'AND'.
 	 *
 	 * @param	mixed
-	 * @param	mixed
+	 * @param	mixed|array
 	 * @param	bool
 	 * @return	CI_DB_query_builder
 	 */
 	public function where($key, $value = NULL, $escape = NULL)
 	{
-		return $this->_wh('qb_where', $key, $value, 'AND ', $escape);
+		return !is_array($value)
+			? $this->_wh('qb_where', $key, $value, 'AND ', $escape)
+			: $this->_wh_in(
+				'qb_where',
+				$key,
+				$value,
+				false,
+				'AND ',
+				$escape
+			);
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


gxgpet:
For the custom key option:

1. There are very few use cases for this in the wild... I don't see how you take it as a general case feature
2. It won't work if the `$key_column_name` column will have duplicate values
3. This effect can be easily achieved by the following one-liner: `array_column($array_result, NULL, $key_column_name)`

For the second proposal:

I'm not sure if I get it right, but you have `$this->db->or_where_in()` and `$this->db->where_in()`... Isn't this what are you looking for?

lubo1998:
> For the custom key option:
> 
>     1. There are very few use cases for this in the wild... I don't see how you take it as a general case feature
> 
>     2. It won't work if the `$key_column_name` column will have duplicate values
> 
>     3. This effect can be easily achieved by the following one-liner: `array_column($array_result, NULL, $key_column_name)`
> 
> 
> For the second proposal:
> 
> I'm not sure if I get it right, but you have `$this->db->or_where_in()` and `$this->db->where_in()`... Isn't this what are you looking for?

For the custom key option:
1. I thought this will be useful since there is a PDO fetching style PDO::FETCH_KEY_PAIR where a column can be chosen for a key.
2. Yes, it is really possible to occur collision with the keys, but I trigger an error if this happen. The feature should be used responsibly.
3. I know there are a lot of ways to achieve this affect, but all of them include addition array loops. array_column iterate over the array too.

For the second proposal:
My idea was to use $this->db->where() and $this->db->or_where() regardless of the parameter type I pass as a value/s (array or int/string).

gxgpet:
Well, why trigger an error? Why not override with the last duplicate value? Why not keep only the first value? All of these assumptions of how it could work will lead to numerous "this feature doesn't work as I want/expect" bug reports and it won't make everyone happy.

```
The feature should be used responsibly.
```

Never expect this from a user.

For the second improvement, it looks like this was addressed previously in #4910.

narfbg:
In addition to @gxgpet's comments, see also #429.

lubo1998:
> In addition to @gxgpet's comments, see also #429.

Actually [**array_column** also loops through the array](https://github.com/php/php-src/blob/b3c5b8cf2485ea99eede9bfa245d26fdb4eeb6c3/ext/standard/array.c#L4067), which are extra iterations. So, I don't think that actually **array_column** solves the problem, but if you have taken your decision, ok.

PR #6007, opened 2020-12-17
PHP 8 compatibility issues

blasto333:


system/libraries/Migration.php
```diff
@@ -288,7 +288,7 @@ public function version($target_version)
 				$this->_error_string = sprintf($this->lang->line('migration_class_doesnt_exist'), $class);
 				return FALSE;
 			}
-			elseif ( ! is_callable(array($class, $method)))
+			elseif (!(class_exists($class) && in_array($method, get_class_methods($class))))
```
gxgpet, line 291: The style guide is not followed; see how was the `!` operator written before (the left and right space). Also, I see that there's an unnecessary pair of parentheses.

gxgpet (changes requested):

blasto333:
@gxgpet I have made the changes. I also removed the check for class_exists is that is already taken care of before this elseif is called

blasto333:
@gxgpet Also found another place of is_callable that won't work. So far these are all the PHP 8 issues I have found

system/libraries/Session/Session.php
```diff
@@ -396,9 +396,8 @@ protected function _ci_init_vars()
 				{
 					$_SESSION['__ci_vars'][$key] = 'old';
 				}
-				// Hacky, but 'old' will (implicitly) always be less than time() ;)
-				// DO NOT move this above the 'new' check!
-				elseif ($value < $current_time)
+				//This is what we had to change for php 8
```
sapics, line 399: This comment said about your fixing, not about the code itself ;)
It looks better to write the fixing comment in the comment log.

system/libraries/Session/Session.php
```diff
@@ -396,9 +396,8 @@ protected function _ci_init_vars()
...
+				elseif ($value === 'old')
```
sapics, line 400: At a glance, when $value is number, the result would be changed, I guess.
`$value === 'old' || $value < $current_time`
or
`$value === 'old' || (is_numeric($value) && $value < $current_time)`
looks better.

kitsites: thank you so much

PR #6009, opened 2020-12-23
Replace phpunit/phpunit with sminnee/phpunit in php8+

sapics:
This PR replaces phpunit/phpunit with sminnee/phpunit in php8+ travis test.
The errors occurring in phpunit itself would be resolved by this replacement.

This commit also removes cache directory.
I set this setting to avoid composer download errors, but this setting didn't work well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,8 @@ services:
   - mysql
   - postgresql
 
-cache:
-  directories:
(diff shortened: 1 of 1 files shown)


sapics:
I would like to switch travis-ci to github actions, thus I close this pr.

PR #6011, opened 2020-12-28
Update Log.php

pjsde:
Fix bug on validation of $level if that does not exist.
If !isset($this->_levels[$level] then the validation of  $this->_threshold_array[$this->_levels[$level] throw an exception.


narfbg:
I don't understand what this is fixing. Please elaborate.

pjsde:
For example, if we execute the following:

```php
 log_message('test', 'This is a test message');
```

As the 'test' level does not exist when validating at line 180:

```php
! isset($this->_threshold_array[$this->_levels[$level]])
```

will throw a Notice exception because ``` $this->_levels['test'] ``` does not exist.

With this correction, the exception will no longer be thrown.

narfbg:
And that would mean you'd never receive a notification about an incorrect log level being coded into your application, potentially missing log messages forever. It might not look pretty right now, but you getting an error is more helpful than not.

PR #6013, opened 2021-01-15
Fix Flashdata time comparison for PHP 8

1stwebdesigns:
In PHP 8, string to number comparison works differently (see https://www.php.net/manual/en/migration80.incompatible.php)

This means that Flashdata now no longer clears after each request, and instead persists indefinitely.  The author of the Session library helpfully left a comment stating that the string `'old' < time()` which under PHP 8 is no longer the case.

### DIFF
--- a/system/libraries/Session/Session.php
+++ b/system/libraries/Session/Session.php
@@ -398,7 +398,7 @@ class CI_Session {
 				}
 				// Hacky, but 'old' will (implicitly) always be less than time() ;)
 				// DO NOT move this above the 'new' check!
-				elseif ($value < $current_time)
+				elseif ($value === 'old' || $value < $current_time)
 				{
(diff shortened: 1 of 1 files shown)


sapics:
It looks same as #6007.

1stwebdesigns:
Sorry, didn't notice that.  However that PR seems to address multiple things, and the one for the Flash data check isn't quite correct

amoktar:
So, how to fix it ?

rickyromansyah2045:
Yes, why you fix it?

shyju93:
This is an issue in newer version of php . You can fix it by replacing line 420 in system/libraries/Session/Session.php:
elseif ($value < $current_time)
with
elseif ($value === 'old' || $value < $current_time)

PR #6015, opened 2021-01-24
Introduce the public folder

elidrissidev:
After this pull request is merged:
- `index.php` will be located at `public/index.php`
- The web server will need to be configured to point to the new web root in the public folder
- `index.html` files that serve as a "Forbidden" error page will be deleted
- The line on top of the php files that prevents direct access to them will be removed
- An `.htaccess` file with a URL rewrite rule will be added in the public folder

With contributions from Jane.


elidrissidev:
I'm very sorry I meant to open this PR in my own fork.

PR #6025, opened 2021-02-28
PoC: SameSite attribute implementation for CI_Input::set_cookie

gxgpet:
Hello!

Here's a proof of concept for how the samesite cookie attribute could be implemented for `CI_Input::set_cookie`. While the lines for computing the `header()` call are no the nicest, I think they should do the job. We can switch, however, to the original `setcookie()` call, by taking into the account the PHP version and using the "path hack".

Any suggestions and ideas are welcomed.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -385,6 +385,7 @@ $config['sess_regenerate_destroy'] = FALSE;
 | 'cookie_path'     = Typically will be a forward slash
 | 'cookie_secure'   = Cookie will only be set if a secure HTTPS connection exists.
 | 'cookie_httponly' = Cookie will only be accessible via HTTP(S) (no javascript)
+| 'cookie_samesite' = Cookie's samesite attribute (Lax, Strict or None)
 |
 | Note: These settings (with the exception of 'cookie_prefix' and
 |       'cookie_httponly') will also affect sessions.
@@ -395,6 +396,7 @@ $config['cookie_domain']	= '';
 $config['cookie_path']		= '/';
 $config['cookie_secure']	= FALSE;
 $config['cookie_httponly'] 	= FALSE;
+$config['cookie_samesite'] 	= 'lax';
 
 /*
 |--------------------------------------------------------------------------
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -396,7 +396,7 @@ $config['cookie_domain']	= '';
 $config['cookie_path']		= '/';
 $config['cookie_secure']	= FALSE;
 $config['cookie_httponly'] 	= FALSE;
-$config['cookie_samesite'] 	= 'lax';
+$config['cookie_samesite'] 	= 'Lax';
 
 /*
 |--------------------------------------------------------------------------
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -364,7 +364,7 @@ class CI_Input {
 		}
 
 		$cookie_header = 'Set-Cookie: '.$prefix.$name.'='.rawurlencode($value);
-		$cookie_header .= ($expire === 0 ? '' : '; expires='.gmdate('D, d-M-Y H:i:s T', 0));
+		$cookie_header .= ($expire === 0 ? '' : '; expires='.gmdate('D, d-M-Y H:i:s T', $expire));
 		$cookie_header .= '; path='.$path.($domain !== '' ? '; domain='.$domain : '');
 		$cookie_header .= ($secure ? '; secure' : '').($httponly ? '; HttpOnly' : '').($samesite !== NULL ? '; SameSite='.$samesite : '');
 		header($cookie_header);
--- a/system/core/Input.php
+++ b/system/core/Input.php
@@ -365,11 +365,26 @@ class CI_Input {
 			log_message('error', $name.' cookie sent with SameSite=None, but without Secure attribute.');
 		}
 
-		$cookie_header = 'Set-Cookie: '.$prefix.$name.'='.rawurlencode($value);
-		$cookie_header .= ($expire === 0 ? '' : '; expires='.gmdate('D, d-M-Y H:i:s T', $expire));
-		$cookie_header .= '; path='.$path.($domain !== '' ? '; domain='.$domain : '');
-		$cookie_header .= ($secure ? '; secure' : '').($httponly ? '; HttpOnly' : '').'; SameSite='.$samesite;
-		header($cookie_header);
+		if (is_php('7.3'))
+		{
+			$setcookie_options = array(
+				'expires' => $expire,
+				'path' => $path,
+				'domain' => $domain,
+				'secure' => $secure,
+				'httponly' => $httponly,
+				'samesite' => $samesite,
+			);
+			setcookie($prefix.$name, $value, $setcookie_options);
+		}
+		else
+		{
+			$cookie_header = 'Set-Cookie: '.$prefix.$name.'='.rawurlencode($value);
+			$cookie_header .= ($expire === 0 ? '' : '; expires='.gmdate('D, d-M-Y H:i:s T', $expire));
+			$cookie_header .= '; path='.$path.($domain !== '' ? '; domain='.$domain : '');
+			$cookie_header .= ($secure ? '; secure' : '').($httponly ? '; HttpOnly' : '').'; SameSite='.$samesite;
+			header($cookie_header);
+		}
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


application/config/config.php
```diff
@@ -395,6 +396,7 @@
 $config['cookie_path']		= '/';
 $config['cookie_secure']	= FALSE;
 $config['cookie_httponly'] 	= FALSE;
+$config['cookie_samesite'] 	= 'lax';
```
narfbg, line 399: I think we should default to None.

system/core/Input.php
```diff
@@ -348,7 +349,25 @@ public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path
...
+		if ($samesite === NULL && config_item('cookie_samesite') !== NULL)
+		{
+			$samesite = strtolower(config_item('cookie_samesite'));
+		}
+		elseif ($samesite !== NULL)
+		{
+			$samesite = strtolower($samesite);
+		}
+
+		if ( ! in_array($samesite, array('lax', 'strict', 'none', NULL), TRUE))
```
narfbg, line 361: `NULL` is redundant here if you'll set to it anyway, but also it will never match because `strtolower(NULL)` above will create an empty string.

Either way, I'd do the whole thing like this:

    isset($samesite) OR $samesite = config_item('cookie_samesite');
    if (isset($samesite)) 
    {
            $samesite = ucfirst(strtolower('samesite'));
            in_array($samesite, array('Lax', 'Strict', 'None'), TRUE) OR $samesite = 'None';
    }
    else
    { 
            $samesite = 'None';
    }

... I don't see a reason why we should ever default to `NULL` and therefore not send SameSite as opposed to sending None.

narfbg:
If the `setcookie()` hack works, it's a better solution. Not only is it simpler, but also when the time comes, it is easier to just remove the version-specific code later.

system/core/Input.php
gxgpet: `strtolower(NULL)` can never occur. Both `if` and `elseif` have `!== NULL` checks.

The idea with the `NULL` case was to have the option to just not send the attribute. The way I see it is that if we force sending the samesite attribute is a BC change from CI's point-of-view. According to MDN, 

> The cookie-sending behavior if `SameSite` is not specified is `SameSite=Lax`.

However, I have some worries about this working the same across all browsers and devices.

While I don't mind that we set a default value for it to `None`, in my opinion, I would have preferred that the behavior of users' applications to be the same upon calling `CI_Input::set_cookie` when upgrading the framework and _only if_, based on a conscious choice, to send the samesite attribute with a specific value (i.e. a conscious choice can be adding the new `cookie_samesite` config value in config.php). Also, a conscious choice is developers changing the method calls to `CI_Input::set_cookie` by setting the new parameter value.

gxgpet:
Also, I would like to address a question from UX point of view. MDN also states that 

> Cookies with `SameSite=None` must now also specify the `Secure` attribute (they require a secure context/HTTPS)

Currently, we have no constraint/validation/logging when `SameSite=None` occurs and when the cookie is non-secure. Should we keep the code this way? Should we log something about it?

narfbg:
Hmm, my intent was to preserve BC via `SameSite=None`, but that's going to be a problem with `Secure` required - good point. I guess `Lax` would be the safest default.
We could log a warning if `None` is configured without `Secure`, sure.

gxgpet:
> but that's going to be a problem with `Secure` required

Also, indeed this scenario. 

You said that we should log a warning. In this case, if I would use the internal `log_message()` function, that would be an error-type log (the most appropriate one). Or did you mean `E_WARNING` via `trigger_error()`? :P

And the last - wanted to double-check that we are on the same page about the `NULL` value thingy. My BC-break concern was that not defaulting the parameter to the `NULL` value would make the lib actually send a samesite value. This is the BC-break I wanted to say about: actually sending a value, doesn't matter what value. By letting the default value to `NULL` it means that no samesite should be sent, upon upgrading the framework - which is the current behavior of the latest version.

narfbg:
I'm not sure about error vs debug, either could be appropriate ... your call I guess.

As for the default, does it really make a diffence if browsers would default to the `Lax` behavior in case no `SameSite` is sent? As far as I can tell, it's only a BC break on paper in that case.
On a related note, since `config_item()` would return `NULL` for non-existing config values anyway, there's no way to tell if this was an explicit choice or not. It's probably more appropriate to have `FALSE` as the "don't send" value, while `NULL` is whatever we determine is the safe default.

gxgpet:
After weighing more what has been discussed, I dropped the possibility to "don't send" the samesite attribute. A concern was outdated browsers and how would they react, but that should be no concern, as probably they should ignore it.

Not sending the samesite should equal to `SameSite=Lax`. So we should be fine the way it is now. What do you think of this?

Also, used the `error` logging level for the `SecureSite=None` scenario, because it would cause the application to break, in the end.

system/core/Input.php
```diff
@@ -348,7 +349,27 @@ public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path
...
+		{
+			$samesite = ucfirst(strtolower($samesite));
+			in_array($samesite, array('Lax', 'Strict', 'None'), TRUE) OR $samesite = 'Lax';
+		}
+		else
+		{
+			$samesite = 'Lax';
+		}
+
+		if ($samesite === 'None' && !$secure)
```
narfbg, line 363: Always have a space after `!`.

system/core/Input.php
```diff
@@ -348,7 +349,27 @@ public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path
...
+		{
+			log_message('error', $name.' is a non-secure cookie sent with SameSite=None. It can be discarded by the browser.');
```
narfbg, line 365: I think the following is clearer:

   "$cookiename" sent with SameSite=None, but without Secure attribute.

system/core/Input.php
```diff
@@ -348,7 +349,27 @@ public function set_cookie($name, $value = '', $expire = 0, $domain = '', $path
...
+		}
+
+		$cookie_header = 'Set-Cookie: '.$prefix.$name.'='.rawurlencode($value);
+		$cookie_header .= ($expire === 0 ? '' : '; expires='.gmdate('D, d-M-Y H:i:s T', $expire));
+		$cookie_header .= '; path='.$path.($domain !== '' ? '; domain='.$domain : '');
+		$cookie_header .= ($secure ? '; secure' : '').($httponly ? '; HttpOnly' : '').'; SameSite='.$samesite;
+		header($cookie_header);
```
narfbg, line 372: Again, why not go with the `setcookie()` hack?

user_guide_src/source/libraries/input.rst
```diff
@@ -252,6 +252,7 @@ Class Reference
 		:param	string	$prefix: Cookie name prefix
 		:param	bool	$secure: Whether to only transfer the cookie through HTTPS
 		:param	bool	$httponly: Whether to only make the cookie accessible for HTTP requests (no JavaScript)
+		:param	string	$samesite: Cookie's SameSite attribute ('Lax', 'Strict', 'None')
```
narfbg, line 255: "Cookie's" is a bit superfluous, less is more :)

user_guide_src/source/libraries/input.rst
```diff
@@ -265,13 +266,14 @@ Class Reference
...
-				'prefix' => 'myprefix_',
-				'secure' => TRUE
+				'name'     => 'The Cookie Name',
+				'value'    => 'The Value',
+				'expire'   => 86500,
+				'domain'   => '.some-domain.com',
+				'path'     => '/',
+				'prefix'   => 'myprefix_',
+				'secure'   => TRUE,
+				'samesite' => 'strict'
```
narfbg, line 276: Please ucfirst this value.

But on another note, the diff implies that usage of tabs/spaces is changed in all the other lines shown above ... Keep it consistent.

narfbg:
> Not sending the samesite should equal to SameSite=Lax. So we should be fine the way it is now. What do you think of this?

:+1:

system/core/Input.php
narfbg: Actually, apparently that hack triggers notices, so nevermind that, but it would be better to make this conditional on the version, so we can just remove this hack easily in the future.

gxgpet:
Pushed the requested changes. Sorry for being late.

> Again, why not go with the setcookie() hack?

Sorry about this, it looks like I missed pushing the commit, I guess. But after reading the conversation from here, I changed to the _mixed_ solution.

rwalkerands:
I have a "legacy" CI project running on PHP 5.6 that I need to patch so that I can set the `SameSite` attribute in cookies.

I just incorporated the upstream changes to `set_cookie()`. But it seems there's a problem, well, at least for me: calling `header()` with just one argument (in `system/core/Input.php`, line 381). This means the `$replace` parameter uses the default value of `true`. This has really bad consequences: you can only set one cookie per response!

I think you have to say `header(..., false)` everywhere to enable setting multiple cookies per response.

PR #6026, opened 2021-03-04
Update Input.php

MeteorGX:
Determine whether the request parameter is `payload request`


gxgpet:
This should be an application code, which suits your particular case and HTTP request structure. For instance, other projects may use different content types than `application/json`.

Issue #6030, opened 2021-03-18
Bug: potential redis session error


kenjis:
**Describe the bug**
Multiple Ajax communications at the same time sometimes result in session errors.
In rare cases, a key is created by another process between the time the `$ttl` is retrieved and the time the key is set.

Reported: (Japanese) <https://groups.google.com/g/codeigniter_jp/c/h65Q7KZtDN4/m/Ef_Vl8DmAAAJ>.

**Affected module(s)**
CI_Session_redis_driver
https://github.com/bcit-ci/CodeIgniter/blob/0648ae2acb3fae84f2cc49e15edb345ccef6cad5/system/libraries/Session/drivers/Session_redis_driver.php#L420-L443

**Expected behavior**
Use [setnx](https://redis.io/commands/setnx) to acquire the lock.
If it fails, retry after some time.

Improved code from the original report:
```php
protected function _get_lock($session_id)
{
   //log_message('error', 'lock' . getmypid() . date(' i:s'));
   // PHP 7 reuses the SessionHandler object on regeneration,
   // so we need to check here if the lock key is for the
   // correct session ID.
   if ($this->_lock_key === $this->_key_prefix.$session_id.':lock')
   {
      return $this->_redis->setTimeout($this->_lock_key, 300);
   }

   // 300 attempts(about 30 second) to obtain a lock, in case another request already has it
   $lock_key = $this->_key_prefix.$session_id.':lock';
   $attempt = 0;
   do
   {
      // Use setnx to perform exclusions with other processes using the lock key. Loop if the lock key cannot be set.
      $result = $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300));
      if ( ! $result)
      {
         usleep(100000);
         continue;
      }
      $result = $this->_redis->setex($lock_key, 300, time());
 
      $this->_lock_key = $lock_key;
      break;
   }
   while (++$attempt < 300);
   if ($attempt === 300)
   {
      log_message('error', 'Session: Unable to obtain lock for '.$this->_key_prefix.$session_id.' after 300 attempts, aborting.');
      return FALSE;
   }

   $this->_lock = TRUE;
   return TRUE;
}
```

narfbg:
I can't read Japanese, so I can't understand what the issue really is.
Also, it's hard to compare code like that ... can you submit a PR?

kenjis:
The issue is a race condition.

- In rare cases, a key is created by another process between the time the $ttl is retrieved and the time the key is set.
[...]

PR #6031, opened 2021-03-25
Fix potential redis session error

kenjis:
In rare cases, a key is created by another process between the time the $ttl is retrieved and the time the key is set.

See #6030

### DIFF
--- a/system/libraries/Session/drivers/Session_redis_driver.php
+++ b/system/libraries/Session/drivers/Session_redis_driver.php
@@ -419,23 +419,14 @@ class CI_Session_redis_driver extends CI_Session_driver implements SessionHandle
 		$attempt = 0;
 		do
 		{
-			if (($ttl = $this->_redis->ttl($lock_key)) > 0)
+			// Use setnx to perform exclusions with other processes using the lock key. Loop if the lock key cannot be set.
+			$result = $this->_redis->set($lock_key, time(), array('nx', 'ex' => 300));
+			if ( ! $result)
 			{
 				sleep(1);
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


gxgpet:
Hello! The first thing I see here is that you are not checking the result of the `$this->_redis->setex($lock_key, 300, time())` operation.

Secondly, The only non-locking call we have in the current implementation is [the 434th line](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Session/drivers/Session_redis_driver.php#L434). The way I see it, is that an issue could occur if you reach that line twice, in two different requests. And that means `$ttl == -1`, which is the case when the key exists, but no TTL assigned to it. Correct me if I'm wrong.

While I don't disprove the possibility of such a scenario, I don't see how it could really happen. Meaning that the lock key exists, but without TTL. 

I'm still thinking if this is really a potential issue, but... assuming that we should do something about it, why not replace it that `setex()` call with a blocking one, from that line?

narfbg:
The more I look at this, the more I believe that it's just sweeping the potential error under the rug.

People have complained from seeing this error in their logs in the past, and unfortunately they just don't understand that it's better to see an error message than remain in the dark. Frankly, the only thing that could be done here, without sacrificing safety, is to make the loop infinite - instead of failing after 30 attempts, attempt forever; instead of giving up on obviously failing lock attempt - reattempt forever. Requests will hang infinitely until the httpd itself gives up, but I suppose some people will simply be happier because they didn't see an error message generated by CI.
That's a half joke, but at this point it's the only alternative I'd seriously consider.

PR #6042, opened 2021-05-21
Auto detect protocol and calculate base_url

doan281:
Auto detect protocol and calculate base_url in config.php file


gxgpet:
Hi! CodeIgniter has already `base_url` auto-detection, and it happens in `CI_Config`'s constructor. This config item exists so that you don't rely on the auto-detection mechanism on a production environment, but rather explicitly set it (see the explanation of the `base_url` config item).

PR #6043, opened 2021-05-27
remove unnecessary extra space in query builder _compile_select()

pjsde:
When query builder is compiling the select make the implode with `', '`. The extra space can lead to error.

For example if we have the following:
```
$this->db->select("GROUP_CONCAT(DISTINCT id ORDER BY id SEPARATOR ',') as group_name", false)
         ->group_by('name')
         ->get('users);
```

This result will be:
```
SELECT GROUP_CONCAT(DISTINCT id ORDER BY id SEPARATOR ', ') as group_name
FROM users
GROUP BY name
```

When the expected should be:
```
SELECT GROUP_CONCAT(DISTINCT id ORDER BY id SEPARATOR ',') as group_name
FROM users
GROUP BY name
```

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -2415,7 +2415,7 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 					$this->qb_select[$key] = $this->protect_identifiers($val, FALSE, $no_escape);
 				}
 
-				$sql .= implode(', ', $this->qb_select);
+				$sql .= implode(',', $this->qb_select);
 			}
 		}
 


gxgpet:
Hello! 
This fixes only this particular case. What if you really want to have `SEPARATOR ', '`? Or even `SEPARATOR ' ,'` (and so on)...

I'm not sure how easy is this to fix, but I would seek a different approach on this issue. Until then, a solution for this would be to stick with the `->query()` method.

pjsde:
When the explode is done in the select() method on line 288, it is assumed that the string will only have field names and that the comma is used to separate the fields, which may not be true as seen in this case. The string can contain formulas that use the comma to separate arguments which causes the formula to be broken by the comma.

In my opinion, adding a space when the implode is done does not add anything useful to the query, making it merely visual, so there is no need for that same space and it can also generate errors in the construction of the query as shown

gxgpet:
> which may not be true as seen in this case 

But this is not the only case, as I was previously saying. What if you have, for instance, `SEPARATOR ', '`, `SEPARATOR ' |'` or anything similar? The fix you are proposing won't act on those scenarios, but rather keep the issue on a different form.

> it can also generate errors in the construction of the query as shown

I understand what you are saying, but let's not call them "errors". :P The generated query is 100% working. But for such scenarios, this is why I recommended using `->query()` instead.

> Another solution is that when the select () method is called with the $ escape = false argument, the string does not explode.

After checking the code a bit, I could agree that this would a fix that should encapsulate the cases I previously mentioned, so I would be fine with it.

I would also like to see (and wait for) Narf's input on this issue.

narfbg:
This issue has been reported multiple times in the past, I think, and problem is that it may mask this side effect for this one particular use case, but it's not really fixing anything. The underlying QB logic remains the same and it's just not meant to deal with complex function calls; in fact, it's not meant to deal with functions at all, though some attempts have been made to not be over-intrusive because people misuse APIs all the time.

This isn't the first time somebody, me included, has thought of just removing that space too, but we just don't know what side effects that would have ... There's a lot of string interpolation under the hood and I'm 99% confident that this space will turn out to be significant somewhere - unless and until someone goes through *all* of the logic to verify and correct the potential side effects, we just can't make this change.
Trust me, this may seem super simple on the surface, but having fixed hundred of QB-related issues myself, I know the tiniest change has massive cascading side-effects.

PR #6045, opened 2021-05-31
Fixes Travis pipeline. Adding PHPUnit 8.

gxgpet:
Lately, the Travis builds were not so green. :P This PR addresses that.

* Added PHPUnit 8 as a potential PHPUnit version. This also sets up the ground for adding in a future PR PHP 8 in the Travis pipeline
* Upgraded to latest vfsstream. The `if` added in #5946 was removed, as it seems there's no issue with PHP 7.4 in latest version.
* `-d mbstring.func_overload=7` is added conditionally now, only if PHP < 7.3
* PHPUnit's `TestCase`'s `void` return type hints were removed to maintain compatibility. This approach was inspired by one of symfony's phpunit bridge features
* other fixes

As already being said, this is a ground for adding PHP 8 in the Travis pipeline, in another separate PR, with test fixes that will most probably come as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_script:
   - sh -c "if [ '$DB' = 'pgsql' ] || [ '$DB' = 'pdo/pgsql' ]; then psql -c 'create database ci_test;' -U postgres; fi"
   - sh -c "if [ '$DB' = 'mysql' ] || [ '$DB' = 'mysqli' ] || [ '$DB' = 'pdo/mysql' ]; then mysql -e 'create database IF NOT EXISTS ci_test;'; fi"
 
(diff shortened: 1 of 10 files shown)


1stwebdesigns:
Great news :-)

narfbg:
Looks solid. But because I tend to immediately merge applicable changes between `develop` and `3.1-stable`, would this work in the latter or do we need to keep it only in `develop`?

gxgpet:
Tried locally with this branch `gxgpet:develop_fixtravis` merged into `3.1-stable` with a PHP 5.6 and it worked. So I guess it's safe to assume that it works on both.

narfbg:
Well, of course it will work on 5.6, but older versions are unfortunately still in the matrix for 3.1-stable ... I guess if it breaks, it breaks, we shouldn't care about those anymore.

gxgpet:
Oh, I see what you meant. I'm kinda confident it shouldn't break anything :D

PR #6047, opened 2021-06-08
Add: ip in Log message

rmdhfz:
Hi,

Just add ip address in error message

### DIFF
--- a/system/core/Log.php
+++ b/system/core/Log.php
@@ -240,6 +240,26 @@ class CI_Log {
 
 	// --------------------------------------------------------------------
 
+	/*get ip address*/
+	private function get_client_ip() {
+	    $ipaddress = '';
+	    if (getenv('HTTP_CLIENT_IP'))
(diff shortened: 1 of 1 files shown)


gxgpet:
I think this feature would be nice, but there are some things to be taken into consideration:

* This should be enabled by a toggle in the config file, and by default should be disabled.
* The CLI executions must not add the IP at all, no matter the config setting, because there's no IP in such cases.
* There's no need for the `Log::get_client_ip()` method. We have the Input class, which handles this already.
* The code must follow our [stylguide](https://codeigniter.com/userguide3/general/styleguide.html).

narfbg:
No, this isn't suitable for us. It may seem like it is at first glance, but it really isn't.
I really don't want to go into lengthy explanations, so I'll just say what we log is very different from e.g. an Apache access.log and that's quite obvious. If you really need some extra info in there, it's easy to extend our core libraries and do that for your own apps.

PR #6066, opened 2021-09-26
Adds PHP 8 in Travis

gxgpet:
This PR adds PHP 8.0 in the Travis matrix.


gxgpet:
The build doesn't start because lately, you have to _explicitly_ select Free plan on the Travis settings (at least this is what I had to do). I do not have access to the bcit-ci organization on Travis, but I requested access. Until then, if someone who has access there could select that plan and retrigger the build for this PR, it would be much appreciated.

narfbg:
Travis is setup, but builds won't start for old PRs. I'm not sure if you have access to modify settings if something comes up, but you can ping me.
As for this one - are the changes to `CI_Output` necessary?

kenjis:
> are the changes to CI_Output necessary?

Yes. If not, there is a warning:
```
Severity:    Warning
Message:     array_shift(): Argument #1 ($array) must be passed by reference, value given
```

gxgpet:
> but you can ping me.

Okay, so somehow, I was able to access the bcit-ci organization settings for Travis. I picked the Free plan. Let me know if it looks ok or if we should change something else there. Also, retriggered the build here! :D

> Yes. If not, there is a warning

Yep, that's exactly why a test failed. :P 

Sorry for the delayed reply.

gxgpet:
Pushed other changes:

* It seems that the queries from `\Mock_Database_Schema_Skeleton::create_tables` were failing for PostgreSQL-PDO always, only that it was silently because of a bug described in the next item; checks were added here anyway.
* Fixed `pdo_pgsql_forge.php`'s `create_table()` to work just as it does for `postgre_forge.php`.

gxgpet:
Hi @narfbg,

I'm going to prepare a PR for adding support for PHP 8.1 as well. If there is anything else which we should change on this PR, please let me know.

PR #6067, opened 2021-09-26
Update Lang.php

paulpivot94:
Adding lines() method to Fetch all lines of text from the language array

### DIFF
--- a/system/core/Lang.php
+++ b/system/core/Lang.php
@@ -199,5 +199,31 @@ class CI_Lang {
 
 		return $value;
 	}
+	
+	
+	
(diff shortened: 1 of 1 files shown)


paulpivot94:
i'm so happy to join CI community to maintain our Likely Framework

gxgpet:
Hi and welcome to the community!

I am unable to see a broad-use case for such a method. I think it's very specific to your project/what are you trying to do, and [extending the framework](https://codeigniter.com/userguide3/general/core_classes.html) by adding your own logic should be the way for this.

PR #6069, opened 2021-10-16
we should reset select after get sql

robotsniper:
if one select sql is error, then will trigger exception handler, and _reset_select won't run . and in exception handler we trigger an update sql ,then update qb_where array will merge with select qb_where array because we do not _reset_select after we get compile sql, then update sql trigger an error again

so we need _reset_select after we get complied select sql

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -1451,9 +1451,9 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 		{
 			$this->limit($limit, $offset);
 		}
-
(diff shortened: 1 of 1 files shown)


gxgpet:
What do you mean by "select sql is error"? 

Can you provide an example of the problem you are encountering?

narfbg:
You're not supposed to treat those exceptions as simple if/else conditions. The fact that you can recover from them is more of a debugging feature than anything else and I don't want to potentially break other stuff for this super niche use-case.

Issue #4361, opened 2016-01-06
If base_url is empty,When using VHOST, there will be some inconvenience. for 3.03


zoujingli:
If base_url is empty, Config will automatically generate this value, using SERVER_ADDR to replace the HTTP_HOST, resulting in the system link and access to the domain name of the link is not consistent. When using VHOST, there will be some inconvenience.

For example: 
access to http://localhost/,
base_url=http://127.0.0.1/ in the system

kenjis:
It is specification for security, not a bug.
Set `http://localhost/` to `base_url`.

Please read http://www.codeigniter.com/user_guide/installation/upgrade_303.html#step-2-make-sure-your-base-url-config-value-is-not-empty
> Auto-detection is never reliable and also has security implications, which is why you should always have it manually configured!

zoujingli:
Since the proposed manual configuration of base_url, the default configuration is naturally not important, but I still want to be able to retain access to the domain name, rather than the IP address.

zoujingli:
I think this is not a serious security problem, if you really want to consider the domain name security, you can add a security domain name configuration items, if the configuration of the security domain name just to allow access to these domain names.

narfbg:
Duplicate of #4209, #4226, #4231.
Won't be changed even if a thousand more people complain about it - security has higher priority.

zoujingli:
I don't think using HTTP host will have security problems, if you insist to use IP to generate a URL, users can only to modify the framework to realize the automatic mechanism of the base URL, every project need to manually configure the domain name, seemed to be somewhat tasteless, for a server to say, in most cases not only run a project.

aloMalbarez:
any hint on what are those "security implications"? 
If I develop my own autodetect for easy code sharing and deploy on multiple domins/paths it would be nice to know.

narfbg:
It introduces a host header injection.

aloMalbarez:
Thanks for the clarification, so I'm going to limit autodetection to dev/QA environments and enforcing config on production.

PR #6076, opened 2021-12-02
Automatically setup base_url

SyedMuradAliShah:
It will set automatically the base_url even if somebody used a folder in localhost or online server.

### DIFF
--- a/system/core/Config.php
+++ b/system/core/Config.php
@@ -88,24 +88,8 @@ class CI_Config {
 		// Set the base_url automatically if none was provided
 		if (empty($this->config['base_url']))
 		{
-			if (isset($_SERVER['SERVER_ADDR']))
-			{
-				if (strpos($_SERVER['SERVER_ADDR'], ':') !== FALSE)
-				{
-					$server_addr = '['.$_SERVER['SERVER_ADDR'].']';
-				}
-				else
-				{
-					$server_addr = $_SERVER['SERVER_ADDR'];
-				}
-
-				$base_url = (is_https() ? 'https' : 'http').'://'.$server_addr
-					.substr($_SERVER['SCRIPT_NAME'], 0, strpos($_SERVER['SCRIPT_NAME'], basename($_SERVER['SCRIPT_FILENAME'])));
-			}
-			else
-			{
-				$base_url = 'http://localhost/';
-			}
+			
+			$base_url = "{$_SERVER['REQUEST_SCHEME']}://{$_SERVER['HTTP_HOST']}" . dirname($_SERVER['SCRIPT_NAME']) . (substr(dirname($_SERVER['SCRIPT_NAME']), -1) == '/' ? '' : '/');
 
 			$this->set_item('base_url', $base_url);
 		}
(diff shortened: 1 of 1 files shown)


gxgpet:
Hi! Could you please explain the issue?

As a side note, CodeIgniter has already `base_url` auto-detection, which should work for local environments as well. The config item `base_url` exists so that you don't rely on this auto-detection mechanism, but rather explicitly set it - even for a `localhost` environment.

kenjis:
This PR introduces Host header injections again.
See https://github.com/bcit-ci/CodeIgniter/commit/0a6b0661305f20ac1fbd219d43f59193bea90d1d

SyedMuradAliShah:
@gxgpet actually without this when you use a Codeigniter inside a directory, the auto-detection mechanism does not work. It detects the HOST_NAME rather than detecting the hostname with directory!
Example: `c:/www/codeigniter-blog`
If `localhost` point to `c:/www` so when we try to access `localhost/codeigniter-blog` codeigniter will detect its `base_url` as `localhost` not with directory which is `codeigniter-blog`

SyedMuradAliShah:
@gxgpet actually without this when you use a Codeigniter inside a directory, the auto-detection mechanism does not work. It detects the HOST_NAME rather than detecting the hostname with directory!
Example: `c:/www/codeigniter-blog`
If `localhost` point to `c:/www` so when we try to access `localhost/codeigniter-blog` codeigniter will detect its `base_url` as `localhost` not with directory which is `codeigniter-blog`

SyedMuradAliShah:
@gxgpet actually without this when you use a Codeigniter inside a directory, the auto-detection mechanism does not work. It detects the HOST_NAME rather than detecting the hostname with directory!
Example: `c:/www/codeigniter-blog`
If `localhost` point to `c:/www` so when we try to access `localhost/codeigniter-blog` codeigniter will detect its `base_url` as `localhost` not with directory which is `codeigniter-blog`

SyedMuradAliShah:
@gxgpet actually without this when you use a Codeigniter inside a directory, the auto-detection mechanism does not work. It detects the HOST_NAME rather than detecting the hostname with directory!
Example: `c:/www/codeigniter-blog`
If `localhost` point to `c:/www` so when we try to access `localhost/codeigniter-blog` codeigniter will detect its `base_url` as `localhost` not with directory which is `codeigniter-blog`

SyedMuradAliShah:
@gxgpet actually without this when you use a Codeigniter inside a directory, the auto-detection mechanism does not work. It detects the HOST_NAME rather than detecting the hostname with directory!
Example: `c:/www/codeigniter-blog`
If `localhost` point to `c:/www` so when we try to access `localhost/codeigniter-blog` codeigniter will detect its `base_url` as `localhost` not with directory which is `codeigniter-blog`

narfbg:
Duplicate of #4361, among many others. You have to configure this by hand - if there was a good way around this, we wouldn't have the config option at all. But there isn't, it's a security sensitive thing as @kenjis pointed out.

PR #6078, opened 2021-12-13
PHP 8.1 Session Driver Errors

mackieee:
Using any Session driver on 8.1 will trigger: 

`CI_Session_database_driver must be compatible with SessionHandlerInterface.`

Errors upon page load. To address these: 

- Added matching `SessionHandlerInterface` typehints 
- Added [#[\ReturnTypeWillChange]](https://wiki.php.net/rfc/internal_method_return_types#proposal) for those methods require a union return type. The comment is a suppresses the deprecation notice to maintain backwards compatibility.

### DIFF
--- a/system/libraries/Session/drivers/Session_database_driver.php
+++ b/system/libraries/Session/drivers/Session_database_driver.php
@@ -126,7 +126,7 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 	 * @param	string	$name		Session cookie name, unused
 	 * @return	bool
 	 */
-	public function open($save_path, $name)
+	public function open(string $save_path, string $name): bool
 	{
 		if (empty($this->_db->conn_id) && ! $this->_db->db_connect())
 		{
@@ -148,7 +148,8 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 	 * @param	string	$session_id	Session ID
 	 * @return	string	Serialized session data
 	 */
-	public function read($session_id)
+	#[\ReturnTypeWillChange]
+	public function read(string $session_id)
 	{
 		if ($this->_get_lock($session_id) === FALSE)
 		{
@@ -204,7 +205,7 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 	 * @param	string	$session_data	Serialized session data
 	 * @return	bool
 	 */
-	public function write($session_id, $session_data)
+	public function write(string $session_id, string $session_data): bool
 	{
 		// Prevent previous QB calls from messing with our queries
 		$this->_db->reset_query();
@@ -276,7 +277,7 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 	 *
 	 * @return	bool
 	 */
-	public function close()
+	public function close(): bool
 	{
 		return ($this->_lock && ! $this->_release_lock())
 			? $this->_failure
@@ -293,7 +294,7 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 	 * @param	string	$session_id	Session ID
 	 * @return	bool
 	 */
-	public function destroy($session_id)
+	public function destroy(string $session_id): bool
 	{
 		if ($this->_lock)
 		{
@@ -331,7 +332,8 @@ class CI_Session_database_driver extends CI_Session_driver implements SessionHan
 	 * @param	int 	$maxlifetime	Maximum lifetime of sessions
 	 * @return	bool
 	 */
-	public function gc($maxlifetime)
+	#[\ReturnTypeWillChange]
+	public function gc(int $maxlifetime)
 	{
 		// Prevent previous QB calls from messing with our queries
 		$this->_db->reset_query();
--- a/system/libraries/Session/drivers/Session_files_driver.php
+++ b/system/libraries/Session/drivers/Session_files_driver.php
@@ -129,7 +129,7 @@ class CI_Session_files_driver extends CI_Session_driver implements SessionHandle
 	 * @param	string	$name		Session cookie name
 	 * @return	bool
 	 */
-	public function open($save_path, $name)
+	public function open(string $save_path, string $name): bool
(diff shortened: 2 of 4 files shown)


gxgpet:
Hi @mackieee,

Thanks for this!
We can't use type-hints since we aim to support PHP 5 as well. The solution here would be to change how we call `session_set_save_handler`, something like:

```
if (PHP_VERSION_ID >= 80100)
{
	session_set_save_handler(
		array($class, 'open'),
		array($class, 'close'),
		array($class, 'read'),
		array($class, 'write'),
		array($class, 'destroy'),
		array($class, 'gc')
	);
} 
else 
{
	session_set_save_handler($class, TRUE);
}
```

Let me know if you are up for changing your PR like this. Otherwise, I'm going to submit a PR very soon to add compatibility for PHP 8.1 as well, adding it to the Travis matrix, fixing broken things, and so on - and I could include this change as well. Just let me know your thoughts on this.

narfbg:
Session handling broken without notice yet again ... I guess that'll never change. Anyway, the individual function declaring format for the save handler is an old one that I was looking forward to get rid of, not get back to. I'm thinking the solution is rather a wrapper class that we include depending on the PHP version.

gxgpet:
> I'm thinking the solution is rather a wrapper class that we include depending on the PHP version.

I would love this solution, tho I wasn't sure how comfortable we would be with it :D
But yeah, let's go for having a wrapping class(es) for our drivers.

narfbg:
Should be done with the commit I just pushed, but as I can only verify so much in a single environment, I'll wait for one of you guys to confirm that it works before I close this.

gxgpet:
Hello,

Sorry for being late for the party :P
It seems that you managed to submit already some changes for PHP 8.1 and switched to GH. It's really cool that it works, but I find it funny that adding the `-d mbstring.func_overload=7` all the time still makes the build green, as this was triggering a warning in PHP, and on Travis was failing.

But anyway, back to Session changes. I was working on a very similar implementation, and looking at the pushed code, I would like to suggest some changes: 
* shouldn't `CI_SessionWrapper` be named `CI_sesssion_wrapper`, but also the file names `Old_session_wrapper.php` and `PHP8_session_wrapper.php`? (or even more simple for the latter, `Session_wrapper.php`)
* it seems that the `$wrapper` variable isn't used
```diff
                $wrapper = new CI_SessionWrapper($class);
                if (is_php('5.4'))
                {
-                       session_set_save_handler($class, TRUE);
+                       session_set_save_handler($wrapper, TRUE);
                }
                else
                {
                        session_set_save_handler(
-                               array($class, 'open'),
-                               array($class, 'close'),
-                               array($class, 'read'),
-                               array($class, 'write'),
-                               array($class, 'destroy'),
-                               array($class, 'gc')
+                               array($wrapper, 'open'),
+                               array($wrapper, 'close'),
+                               array($wrapper, 'read'),
+                               array($wrapper, 'write'),
+                               array($wrapper, 'destroy'),
+                               array($wrapper, 'gc')
                        );

                        register_shutdown_function('session_write_close');
```
* It seems that `samesite` is not used for PHP 5. A workaround would be to tweak the `session_set_cookie_params()` call:
```diff
                {
                        session_set_cookie_params(
                                $params['cookie_lifetime'],
-                               $params['cookie_path'],
+                               $params['cookie_path'].'; samesite='.$params['cookie_samesite'],
                                $params['cookie_domain'],
                                $params['cookie_secure'],
                                TRUE // HttpOnly; Yes, this is intentional and not configurable for security reasons
```

And a happy new year, by the way! :D

narfbg:
Yea, the wrapper names could be different for a number of reasons, but I'd rather keep them like so as a reminder that it's a temporary hack and not something that the Loader would handle with our conventions.

Addressed the other 2 issues, spot on review.

Still waiting for people to actually run the code though. I can only do so much in my local environment.

Shmakov:
Thank you for working on this @narfbg!

I apologies if this is not the right place to ask for it. It looks like the current version of `dev-3.1-stable` on [packagist](https://packagist.org/packages/codeigniter/framework#dev-3.1-stable) is broken due to Session error (the error I get is below). The packagist version does not reference the latest commit, and instead uses [25ae2a3](https://github.com/bcit-ci/CodeIgniter/commit/25ae2a3c0a8f20098181e20d55a61c0c42559ee9).

Any chance the packagist could be updated to use the latest version of 3.1-stable branch?

Thank you!

Error message:

```
session_set_save_handler() expects parameter 1 to be SessionHandlerInterface, object given
```

narfbg:
I manually updated it just now. Please do tell what happens on your end.

Shmakov:
Thank you so much for the quick response @narfbg!

The error message is gone and the sessions are working as expected.

jamieburchell:
Are sessions using the database driver supposed to be working fine now from dev-3.1-stable on PHP 8.1? I'm still getting the return type deprecation notices and a new session cookie on each page load.

narfbg:
Seeing your edit, does that mean you don't actually get errors with the latest code?

jamieburchell:
> Seeing your edit, does that mean you don't actually get errors with the latest code?

That's correct, sorry for the noise. I had put "3.1-stable" in my composer.json, instead of "dev-3.1-stable".

narfbg:
It utilizes some internal PHP mechanics to refresh sessions in which data wasn't changed, but really, just making sure it doesn't obviously break anything is good enough.

jamieburchell:
> It utilizes some internal PHP mechanics to refresh sessions in which data wasn't changed, but really, just making sure it doesn't obviously break anything is good enough.

I checked out dev-3.1-stable after your merge and haven't noticed anything broken with regards to sessions.

narfbg:
3.1.12 tag is live, closing this.

jamieburchell:
@narfbg All of the tagged releases are showing "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." ?

narfbg:
I know, don't worry about it, there's nothing wrong going on.

PR #6088, opened 2022-01-27
normalize composer.json

totoprayogo1916:
Normalize `composer.json` using [`ergebnis/composer-normalize`](https://github.com/ergebnis/composer-normalize)


narfbg:
Why do this? What's the benefit?

totoprayogo1916:
sorted and neater so it's easy to read

narfbg:
Yeah, no ... this is just moving things around, a change for the sake of change.

Issue #6089, opened 2022-01-27
PHP 8.1 deprecation in Output class (elapsed time/memory usage)


jamieburchell:
Seeing this on dev-3.1-stable with PHP 8.1:

```
ERROR - 2022-01-27 23:03:20 --> Severity: 8192 --> str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated /home/jamie/projects/example/vendor/codeigniter/framework/system/core/Output.php 461
```

For reference, it's the `str_replace` here:

```
if ($this->parse_exec_vars === TRUE)
{
	$memory	= round(memory_get_usage() / 1024 / 1024, 2).'MB';
	$output = str_replace(array('{elapsed_time}', '{memory_usage}'), array($elapsed, $memory), $output);
}
```

`$output` is `NULL` because `$this->final_output` [defaults to `NULL`](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Output.php#L58) instead of a string and [`$output` can then be set to `NULL`](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/core/Output.php#L438)

```
// Set the output data
if ($output === '')
{
	$output =& $this->final_output;
}
```

PR #6090, opened 2022-01-28
Default $final_output to string instead of NULL

jamieburchell:
Fixes https://github.com/bcit-ci/CodeIgniter/issues/6089


narfbg:
In the future, please just describe the issue in the pull request instead of opening separate issue and PR.

PR #6094, opened 2022-02-04
User agent can't differentiate between Chrome for iOS and Safari

autribaghkhanian:
Google documentation: https://chromium.googlesource.com/chromium/src/+/HEAD/docs/ios/user_agent.md

Fix: check browsers that look like Safari for the special "crios" keyword and register as a Chrome browser instead if a match is found.


narfbg:
Sorry, but adding conditionals like that directly in the code is not sustainable. If you want this added, you'll have to find a suitable pattern to include in *config/user_agents.php*.

PR #6098, opened 2022-02-05
[SKIP] Indent & whitespaces

totoprayogo1916:


totoprayogo1916:
I have fixed the PR as your comment.

[SKIP] is what I usually use to skip the CI.

narfbg:
I see ... we use `[ci skip]`, but you shouldn't be worrying about that.

totoprayogo1916:
OK, thanks for accepting.

PR #6100, opened 2022-02-09
profiler.php PHP 8.1 deprecation notice 

svennd:
check for null value on $val

for this warning : 
`Message: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated`

I think there might be more cases, but I only fixed the case I ran into. (also not sure if this is a valid style of fixing)

### DIFF
--- a/system/libraries/Profiler.php
+++ b/system/libraries/Profiler.php
@@ -496,7 +496,7 @@ class CI_Profiler {
 			}
 
 			$output .= '<tr><td style="padding:5px;vertical-align:top;color:#900;background-color:#ddd;">'
-				.$config.'&nbsp;&nbsp;</td><td style="padding:5px;color:#000;background-color:#ddd;">'.$pre.htmlspecialchars( (string) $val, ENT_QUOTES, config_item('charset')).$pre_close."</td></tr>\n";
+				.$config.'&nbsp;&nbsp;</td><td style="padding:5px;color:#000;background-color:#ddd;">'.$pre.htmlspecialchars((string) $val, ENT_QUOTES, config_item('charset')).$pre_close."</td></tr>\n";
 		}
 
 		return $output."</table>\n</fieldset>";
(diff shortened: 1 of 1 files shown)


jamieburchell:
Technically, '0' and 0 would be converted to '' (although an unlikely scenario). But I suspect that @narfbg will prefer you to cast `$val` to a string here instead:

`htmlspecialchars((string) $val)`

svennd:
Thanks for the feedback Jamie, I changed the update to string cast instead.

system/libraries/Profiler.php
```diff
@@ -496,7 +496,7 @@ protected function _compile_config()
 			}
 
 			$output .= '<tr><td style="padding:5px;vertical-align:top;color:#900;background-color:#ddd;">'
-				.$config.'&nbsp;&nbsp;</td><td style="padding:5px;color:#000;background-color:#ddd;">'.$pre.htmlspecialchars($val, ENT_QUOTES, config_item('charset')).$pre_close."</td></tr>\n";
+				.$config.'&nbsp;&nbsp;</td><td style="padding:5px;color:#000;background-color:#ddd;">'.$pre.htmlspecialchars((string)$val, ENT_QUOTES, config_item('charset')).$pre_close."</td></tr>\n";
```
narfbg, line 499: Should have a space in between. Same for the other line that's being changed ...

narfbg (changes requested):

svennd:
Thanks for the feedback!

system/libraries/Profiler.php
```diff
@@ -496,7 +496,7 @@ protected function _compile_config()
 			}
 
 			$output .= '<tr><td style="padding:5px;vertical-align:top;color:#900;background-color:#ddd;">'
-				.$config.'&nbsp;&nbsp;</td><td style="padding:5px;color:#000;background-color:#ddd;">'.$pre.htmlspecialchars($val, ENT_QUOTES, config_item('charset')).$pre_close."</td></tr>\n";
+				.$config.'&nbsp;&nbsp;</td><td style="padding:5px;color:#000;background-color:#ddd;">'.$pre.htmlspecialchars( (string) $val, ENT_QUOTES, config_item('charset')).$pre_close."</td></tr>\n";
```
narfbg, line 499: *In between*, not around ... Just one space between the typecast and the variable.

narfbg (changes requested):

system/libraries/Profiler.php
svennd: Thanks for the patience :) fixed now !

PR #6113, opened 2022-03-12
Update Exceptions.php

SynacktivCerv:
In some cases, error messages could lead to Cross-Site Scripting vulnerability. This small fix sanitize potential user input from GET/POST parameters that could be returned into error messages.

For example with this kind of code:
`
		if (isset($_GET['page'])) { 
			$this->load->view($_GET['page']);
		 }
`

If the page is load with _/?page=<script>PAYLOAD</script>_ the JavaScript will be executed.

### DIFF
--- a/system/core/Exceptions.php
+++ b/system/core/Exceptions.php
@@ -171,6 +171,17 @@ class CI_Exceptions {
 		else
 		{
 			set_status_header($status_code);
+			if (is_array($message))
+			{
+				foreach ($message as &$value)
+				{
+				   $value = htmlspecialchars($value);
+				}
+			}
+			else
+			{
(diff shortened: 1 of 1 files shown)


gxgpet:
Why would you pass `$_GET['page']` raw variable? We have a [dedicated library](https://codeigniter.com/userguide3/libraries/input.html) for fetching GET parameters.

SynacktivCerv:
The example is not my code but something I saw during a security review.

Furthermore, by using the dedicated library, the injection on error page seems to still be possible. Indeed, the documentation from your link is saying that " XSS escaping should be performed on output, not input!".

SynacktivCerv:
Hi,

Any update on this issue?

Regards,

PR #6114, opened 2022-03-14
on documentation

sinanartun:
https://www.codeigniter.com/userguide3/helpers/download_helper.html
it says we can use force_download with second parameter is NULL
but online 67 there is an OR condition makes this impossible to use.
so I believe it must be an AND.

// Contents of photo.jpg will be automatically read
force_download('/path/to/photo.jpg', NULL);

### DIFF
--- a/system/helpers/download_helper.php
+++ b/system/helpers/download_helper.php
@@ -64,7 +64,7 @@ if ( ! function_exists('force_download'))
 	 */
 	function force_download($filename = '', $data = '', $set_mime = FALSE)
 	{
-		if ($filename === '' OR $data === '')
+		if ($filename === '' AND $data === '')
 		{
 			return;
 		}


gxgpet:
Hello,

That `if` helps to avoid calling the function without ~any parameters~ at least any of the parameters. Further checks of `$data` being `NULL` or not are done a few lines below.

You can actually try the helper itself and see it for yourself.

sinanartun:
on documentation, it says you can use force download like this. 

`force_download('/path/to/photo.jpg', NULL);`

but It is not possible with an OR condition. 

I tried it

gxgpet:
> but It is not possible with an OR condition.

So, by calling `force_download('/path/to/photo.jpg', NULL);` you have `$filename = '/path/to/photo.jpg'` and `$data = NULL`. 
So please explain to me how would the following `if` pass (or have any effect): `if ('/path/to/photo.jpg' === '' OR NULL === '')`.

> I tried it

Well... I doubt you did, or at least that you did with this very specific scenario you are pointing to. 
And by trying it, I didn't mean only to try to imagine what the code would do, but actually have some real code that is running, where you can see/debug how the flow goes, how the variables are changing, etc.

sinanartun:
discord synan#8563

narfbg:
`''` is an empty string, not the same thing as `NULL` ...

sinanartun:
thanks alot

PR #6117, opened 2022-03-17
fix: PHP 8.1 deprecated error in URI::ruri_string()

kenjis:
```
ltrim(): Passing null to parameter #1 ($string) of type string is deprecated
```

**Steps to Reproduce**
In a controller:
```php
echo $this->uri->ruri_string();
```

### DIFF
--- a/system/core/URI.php
+++ b/system/core/URI.php
@@ -656,7 +656,7 @@ class CI_URI {
 	 */
 	public function ruri_string()
 	{
-		return ltrim(load_class('Router', 'core')->directory, '/').implode('/', $this->rsegments);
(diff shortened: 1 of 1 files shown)


gxgpet:
I'm not sure if this should be the fix or it, or that we should default the property `CI_Router::$directory` to an empty string (`''`), as we expect anyway that to be a string.

Of course, this would imply for sure other 2 places to be changed as well, since we check for the `NULL`ness of the property.

slimdestro:
instead string(), it should be an isset() like this: 
`return ltrim(isset(load_class('Router', 'core')->directory), '/').implode('/', $this->rsegments);`

PR #6129, opened 2022-04-12
use null coalesce to suppress deprecation notice in php 8.1

tony-dunlop:
Signed-off-by: Tony Dunlop <tony.dunlop@fisss.org>

spotted a deprecation notice when running unit test with php 8.1

addresses the following:

`RuntimeException: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated on line 726 in file /vendor/codeigniter/framework/system/core/Common.php`

### DIFF
--- a/system/core/Common.php
+++ b/system/core/Common.php
@@ -727,7 +727,7 @@ if ( ! function_exists('remove_invisible_characters'))
 
 		do
 		{
-			$str = preg_replace($non_displayables, '', $str, -1, $count);
+			$str = preg_replace($non_displayables, '', $str??'', -1, $count);
 		}
 		while ($count);
 


tony-dunlop:
hmm, just noticed the tests failed for php 5.x, is that still supported? might need to change the null coalesce for a ternary if is?

gxgpet:
Hello! 
Yes, we still aim at maintaining compatibility with PHP 5. But no worries, if we find there's a bug that we should handle, we can ensure that `$str` is a string in other ways.

The only way that this would happen is that either you are passing `NULL` as the value of `$str`, or that over that iteration, `$str` gets nulled by `preg_replace()` call.

Could you please share an example of how we could reproduce this?

PR #6140, opened 2022-06-21
Bug Fix: [cookie_helper.php] set_cookie was missing sameSite argument

Mouerr:
Hi,

I was trying to set a cookie with **sameSite** using the cookie helper function **set_cookie**, but i found out that the function does not have samesite as **argument**, after reading the source code and PR i found that **input** class has already that parameter, so i think changes in cookie_helper was forgotten in(https://github.com/bcit-ci/CodeIgniter/pull/6025).

Thanks.

### DIFF
--- a/system/helpers/cookie_helper.php
+++ b/system/helpers/cookie_helper.php
@@ -68,10 +68,10 @@ if ( ! function_exists('set_cookie'))
 	 * @param	bool	true makes the cookie accessible via http(s) only (no javascript)
 	 * @return	void
 	 */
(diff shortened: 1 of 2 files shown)


jamieburchell:
@Mouerr @gxgpet the expire parameter in the input class defaults to an empty string not an int in 3.1.13. If this is merged that will be inconsistent.

Venkat-Internshala:
@Mouerr @gxgpet Is there any possibility of fixing this issue and releasing it as a stable update?

PR #6150, opened 2022-07-24
fixes #6149 + removing deprecated variable variables in general

### LINKED ISSUE #6149: PHP 8.2: ${var} string interpolation deprecated
Ref: https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated

This is used in `system/libraries/Xmlrpcs.php`

```php
	/**
	 * Multi-call Function: Error Handling
	 *
	 * @param	mixed
	 * @return	object
	 */
	public function multicall_error($err)
	{
		$str = is_string($err) ? $this->xmlrpcstr["multicall_${err}"] : $err->faultString();
		$code = is_string($err) ? $this->xmlrpcerr["multicall_${err}"] : $err->faultCode();

		$struct['faultCode'] = new XML_RPC_Values($code, 'int');
		$struct['faultString'] = new XML_RPC_Values($str, 'string');

		return new XML_RPC_Values($struct, 'struct');
	}

```

Please fix it to prepare support for php 8.2

orionstar:
remove deprecated variable variables / string interpolation in variable names

### DIFF
--- a/system/database/DB_query_builder.php
+++ b/system/database/DB_query_builder.php
@@ -692,11 +692,11 @@ abstract class CI_DB_query_builder extends CI_DB_driver {
 				$k = substr($k, 0, $match[0][1]).($match[1][0] === '=' ? ' IS NULL' : ' IS NOT NULL');
 			}
 
-			${$qb_key} = array('condition' => $prefix.$k, 'value' => $v, 'escape' => $escape);
-			$this->{$qb_key}[] = ${$qb_key};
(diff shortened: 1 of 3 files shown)


jamieburchell:
I'm not sure the change to `system/libraries/Xmlrpc.php` is necessary as this was previously a literal string, so no variable interpolation would have taken place anyway? :/ Unless that was a bug too and it should have been @narfbg ?

gxgpet:
Yes, for sure that was a string interpolation bug in `Xmlrpc.php`.

Thanks for this PR!

PR #6154, opened 2022-08-19
Fix composer_autoload

igophper:
Fix composer_autoload：I got an error when I set `$config['composer_autoload'] = TRUE;`  in `application/config/config.php` .
I find the path vender directory is wrong. It should be **FCPATH** instead of **APPPATH**.

### DIFF
--- a/system/core/CodeIgniter.php
+++ b/system/core/CodeIgniter.php
@@ -119,9 +119,9 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 	{
 		if ($composer_autoload === TRUE)
 		{
-			file_exists(APPPATH.'vendor/autoload.php')
-				? require_once(APPPATH.'vendor/autoload.php')
-				: log_message('error', '$config[\'composer_autoload\'] is set to TRUE but '.APPPATH.'vendor/autoload.php was not found.');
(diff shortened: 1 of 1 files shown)


madwings:
You have to read the docs more carefully. This one does not need fixing. Here they are and a fix for your problem.
```
/*
|--------------------------------------------------------------------------
| Composer auto-loading
|--------------------------------------------------------------------------
|
| Enabling this setting will tell CodeIgniter to look for a Composer
| package auto-loader script in application/vendor/autoload.php.
|
|	$config['composer_autoload'] = TRUE;
|
| Or if you have your vendor/ directory located somewhere else, you
| can opt to set a specific path as well:
|
|	$config['composer_autoload'] = '/path/to/vendor/autoload.php';
|
| For more information about Composer, please visit http://getcomposer.org/
|
| Note: This will NOT disable or override the CodeIgniter-specific
|	autoloading (application/config/autoload.php)
*/
$config['composer_autoload'] = FCPATH.'vendor/autoload.php';
```

igophper:
@madwings Thank you for your reply.I'm sorry I didn't see it clearly.Your reply is really grateful.This is actually my first one "pull request".It is obvious that this is a failure.But your reply really encouraged me.Anyway, thank you.

PR #6158, opened 2022-09-12
Update user guide link

ulaszek-james:
Also added cool emojis
Signed-off-by: James Ulaszek


gxgpet:
That isn't our URL for the user guide. This was fixed anyway in #6169.

PR #6160, opened 2022-09-14
Drop support for PHP <7.3

elidrissidev:
Only removed the conditions that execute different code for PHP <7.3.


elidrissidev:
Meant to create in my fork, sorry.

PR #6165, opened 2022-10-04
Fix PHP 8.2 deprecated use of ${} syntax in Query Builder

jamieburchell:
I'm not really sure why it's necessary to create a dynamic variable name here in the first place @narfbg, but this PR resolves the deprecation here.


jamieburchell:
Never mind. I've just spotted https://github.com/bcit-ci/CodeIgniter/pull/6150

PR #6166, opened 2022-10-07
fix incorrect variable name and missing TTL in Cache_redis

tomek-artneo:
There is a little bug, which causes warning. Entire method uses `$data` variable, but in conditional check, there is `$value`. Leftover from recent updates.


tomek-artneo:
I have also added setting expire time for cached items inside `save` method. TTL wasn't handled at all.

rickhewes:
Would be great to get this merged!

:+1:

PR #6174, opened 2022-11-08
Adapt unit tests so as not to fail when run as root

tenzap:
When the phpunit tests are run as root the test Log_test::test_configuration fails because it can write to the /root/ dir while the test expects that /root/ is not writeable. So fix the test so that it doesn't fail when it is run as root.


gxgpet:
No, that part of the test should check the situation where the log path is not writeable/not a dir. We shouldn't check for the _enabled_ property differently. 

We have `vfsStream` and we shall use it for this. Please see the below commit.

PR #6182, opened 2022-12-13
Fix auto_typography() if no chunks are found

dmmihaylov:


gxgpet:
Hello!

Can you please describe the problem and give an input example for reproducing the issue?

PR #6186, opened 2023-01-04
Fix PHP 8.x compatibility

manusfreedom:
`xml_parser_create` no longer returns a resource but a opaque class `XMLParser`.

Xmlrpcs.php must be corrected as well.

### DIFF
--- a/system/libraries/Xmlrpc.php
+++ b/system/libraries/Xmlrpc.php
@@ -1151,7 +1151,7 @@ class XML_RPC_Message extends CI_Xmlrpc
 		//-------------------------------------
 
 		$parser = xml_parser_create($this->xmlrpc_defencoding);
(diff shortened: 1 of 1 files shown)


gxgpet:
Sounds good, only that `spl_object_id()` isn't compatible with all the PHP versions we are supporting. We will have to use something else.

I will try to come up with a suggestion or an implementation in the following days.

marios88:
@gxgpet you can have a look at 

https://github.com/symfony/polyfill-php72/blob/70f4aebd92afca2f865444d30a4d2151c13c3179/Php72.php#L89

for an implementation

madwings:
@gxgpet anything new on this?

gxgpet:
Hi! Please check #6264 and let me know there if it doesn't work for you.

PR #6226, opened 2023-06-27
use up-to-date sphinx_rtd_theme extension

tenzap:
Currently sphinx_rtd_theme is a fixed asset in the repository.

This can lead to various problems:

- search doesn't work anymore when building the doc with recent sphinx
- generated html & js code not in line with current standard
- no fixes from upstream are integrated

BTW, as per the doc at:
  https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html#via-git-or-download

  Installing directly from the repository source is deprecated and is not recommended. Static assets won’t be included in the repository starting in release 3.0.0.

This removes the theme, and updates conf.py to use the now prefered method to get it as per:
https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html#how-to-install-and-use-the-theme

**You may want to tweak some options & update the configuration of the theme if you want it to look exactly like before.**


tenzap:
IIRC I recently saw a similar change in CI4

kenjis:
CI4 are using Sphinx 5.3.
https://github.com/codeigniter4/CodeIgniter4/pull/8428

tenzap:
That's not the one I was thinking of. There was one which removed the sphinx theme. The change of this pr is already applied in the Debian package, but yes Debian runs a recent version of sphinx I guess.
Btw I was able to make cilexer work even with python 3.12, so I'm not sure it was actually required to remove it in ci4.

I think it was commit https://github.com/codeigniter4/CodeIgniter4/commit/806bb40a1077dc7a98b1afb216d058c668c12d3e

PR #6227, opened 2023-07-04
Fix PHP 8.x compatibility

satulg01:


satulg01:
preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated

PR #6237, opened 2023-08-26
Ignoring specific error messages

MichalSkoula:
If you want to silence some specific errors from the log - e.g. bots that generate 404 etc.


gxgpet:
No... and for the same rationale explained in #6239.

PR #6239, opened 2023-08-28
Current URL to be appended at log messages

MichalSkoula:
It is often useful to know not only the file and line where the error occurred, but also the actual URL.


bunglegrind:
I've no idea if anyone from the core team will ever merge this, IMHO those extensions (this one and https://github.com/bcit-ci/CodeIgniter/pull/6237 ) shouldn't belong to the base class but to an extension. However, we could debate if the base class should implement the template method pattern to simplify class extension.

MichalSkoula:
If I can edit log_filename and log_date_format, why not this?

I did it primarily for myself and my fork, so I offer it here as well.

gxgpet:
Thanks for the PR, but I can't merge it.

While for you it might be the URL, for other people might be the controller name/method name, and so on. While for you it might be an appending, for others a prepending would be better, and so on.

Indeed, the solution to this is to extend the core class of the framework's base code and adapt it to your needs.

PR #6260, opened 2024-01-09
Fix compatibility for PHP 8☠

NathanMontois:


tenzap:
AFAIK ci3 must remain compatible with php 5.6 (maybe 5.4?)

pocketarc:
@tenzap Good call. Adding symfony/polyfill could be an option; it'd handle this and enable the use of more recent PHP functions in other parts of the system.

Edit: Alternatively, spl_object_hash() is PHP 5.2+, and would probably be the best option.

gxgpet:
Hi there. 

If you could check the fix provided in #6264, it would be much appreciated. If you find anything faulty, please leave a comment there.

PR #6264, opened 2024-02-10
Fixes the compatibility of XML-RPC library with PHP 8

gxgpet:
After long thoughts and checks on this, I came up with this fix for the XML-RPC library for the new PHP 8 versions.

While I pushed a unit test and made manual tests by myself, anyone who actively uses this library and can test it with a real-world scenario would be much appreciated.

I don't see the need for the `xh` property to be a multi-dimensional array. If a new request gets fired, then the previous values won't be used at all.

### DIFF
--- a/system/libraries/Xmlrpc.php
+++ b/system/libraries/Xmlrpc.php
@@ -1151,8 +1151,7 @@ class XML_RPC_Message extends CI_Xmlrpc
 		//-------------------------------------
 
 		$parser = xml_parser_create($this->xmlrpc_defencoding);
-		$pname = (string) $parser;
-		$this->xh[$pname] = array(
+		$this->xh = array(
 			'isf'		=> 0,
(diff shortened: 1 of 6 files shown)


tenzap:
Unfortunately the tests report deprecation warnings with php 8.4

```
Deprecated: xml_set_element_handler(): Passing non-callable strings is deprecated since 8.4 in system/libraries/Xmlrpc.php on line 1167

Deprecated: Function xml_set_object() is deprecated since 8.4, provide a proper method callable to xml_set_*_handler() functions in system/libraries/Xmlrpc.php on line 1165
```

I also had to change the line in `/tests/codeigniter/libraries/Xmlrpc_test.php` otherwise phpunit would fail
```
$this->input = new CI_Input($security);
```
to
```
$this->input = new Mock_Core_Input($security, new Mock_Core_Utf8());
```

PR #6173, opened 2022-11-06
Adding PHP 8.2 support

gxgpet:
This adds support for PHP 8.2.

Basically, fixing dynamic props assigning.

PR #6277, opened 2024-04-11
php8.2 core related changes - V1

Usha-kundar123:


jafjaf:
PHP 8.2 changes are already implemented in #6173.

Also see: https://github.com/bcit-ci/CodeIgniter/pull/6173#issuecomment-1967673736

If you instead would like an updated repo with PHP 8.2 support see;
- https://github.com/ib3ltd/CodeIgniter/tree/master
- https://github.com/pocketarc/codeigniter

PR #6279, opened 2024-04-22
Add WebP and AVIF support

manusfreedom:
- Add support for WebP and AVIF support
- Use PHP IMAGETYPE_x constant
- Remove WebP support from NetPBM


tenzap:
will this work for users running php < 7.1 (eg. php 5.6) ? ci3 still supports these PHP old versions.
`IMAGETYPE_AVIF` constant doesn't seem known by php 5 https://php-legacy-docs.zend.com/manual/php5/en/function.exif-imagetype

manusfreedom:
This is not supported in PHP 5.x, so for them you need to add a `defined('IMAGETYPE_AVIF') OR define('IMAGETYPE_AVIF', 19);` in `constant.php.`.

PR #6299, opened 2024-09-15
PHP 8.3: Deprecation Warning for Dynamic Property Creation in Class

Richin:
This pull request addresses the deprecation warning introduced in PHP 8.3 regarding dynamic property creation in core classes.

### DIFF
--- a/system/core/Controller.php
+++ b/system/core/Controller.php
@@ -67,6 +67,84 @@ class CI_Controller {
 	public $load;
 
 	/**
+	 * CI_Benchmark
+	 *
+	 * @var	CI_Benchmark
+	 */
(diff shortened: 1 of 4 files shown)


codestellar:
sounds good @Richin however, the current version of codeigniter is supporting 8.1 only. I think there would be more changes to support 8.2

pocketarc:
There's already a heavily-discussed PR for this change: https://github.com/bcit-ci/CodeIgniter/pull/6173

Once/if that gets merged in, this PR will no longer be necessary.

There have also been a number of forks to CodeIgniter 3 that add the necessary changes, including my own: https://github.com/pocketarc/codeigniter

PR #6330, opened 2026-03-19
Restore deprecated functions that were removed in 3.2.0-dev (for backward compatibility).

pocketarc:
There were several removals in 3.2.0-dev that forced users to change working code for no real benefit. Each is a trivial wrapper or alias with zero maintenance burden. Restoring them reduces friction for anyone migrating from CI 3.1.x to this fork.

We've reverted several of these in the past, but I wanted to go through the upgrade guide and reduce as many breaking changes as I could. This PR gets rid of a bunch of breaking changes.

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -15,10 +15,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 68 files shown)


pocketarc:
Closed, apologies - it was not my intention to raise it with upstream, just [inside the fork](https://github.com/pocketarc/codeigniter/pull/44).

